PR #2, opened 2018-06-09
Fix GitGitGadget on Linux and with different Git versions

dscho:
This patch series addresses the assumptions in the current `master` branch that apply only to my own Windows setup, but probably not to anybody else's. And certainly not to the VSTS build which I want to use as "release gate", i.e. I want to require that build to pass before merging any Pull Request.

### DIFF
--- a/lib/patch-series.ts
+++ b/lib/patch-series.ts
@@ -236,12 +236,13 @@ export class PatchSeries {
                 throw new Error("No From: line found in header:\n\n" + header);
             }
 
+            let replaceSender = thisAuthor;
             if (isGitGitGadget) {
                 const onBehalfOf = i === 0 && senderName ?
                     senderName : authorMatch[2].replace(/ <.*>$/, "");
                 // Special-case GitGitGadget to send from
                 // "<author> via GitGitGadget"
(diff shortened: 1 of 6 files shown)


dscho:
@derrickstolee any chance you can review this?

lib/patch-series.ts
```diff
@@ -68,7 +70,9 @@ export class PatchSeries {
                 }
             });
 
-            rangeDiff = await git(["range-diff", "--no-color", range]);
+            if (await gitCommandExists("range-diff", workDir)) {
```
derrickstolee: What happens if `rangeDiff` stays uninitialized?

derrickstolee: (I see higher up that it is initialized to "", so things probably aren't _broken_, but not great)

dscho: If it is empty, or false, or undefined, nothing gets inserted in `insertRangeDiff()`: https://github.com/dscho/gitgitgadget/blob/2fb305a285398b8ac957ef35f51e0be5e295c63a/lib/patch-series.ts#L356-L358

Issue #16, opened 2018-07-11
multiple "Cc:" headers


sunshineco:
Emails generated by `gitgitgadget` may contain multiple `Cc:` headers. For example:

```
Message-Id: <20cd1b914e74c8f45e5cc9ed82d83eee1463cd91.1531312689.git.gitgitgadget@gmail.com>
In-Reply-To: <pull.8.git.gitgitgadget@gmail.com>
References: <pull.8.git.gitgitgadget@gmail.com>
From: Johannes Schindelin via GitGitGadget <gitgitgadget@gmail.com>
Date: Fri, 9 Mar 2018 17:36:47 +0100
Subject: [PATCH 3/3] rebase --rebase-merges: adjust man page for octopus support
...
To: git@vger.kernel.org
Cc: Junio C Hamano <gitster@pobox.com>
Cc: Johannes Schindelin <johannes.schindelin@gmx.de>
Sender: git-owner@vger.kernel.org
List-ID: <git.vger.kernel.org>
X-Mailing-List: git@vger.kernel.org
```

According to [RFC 5322 § 3.6](https://tools.ietf.org/html/rfc5322#section-3.6), only a single `Cc:` header is allowed.

Moreover, multiple `Cc:` headers may cause delivery problems. For instance, anti-spam software may reject such email:

```
X-Amavis-Alert: BAD HEADER SECTION Duplicate header field: "Cc"
```

dscho:
> Emails generated by gitgitgadget may contain multiple Cc: headers

I guess the code responsible for this is https://github.com/gitgitgadget/gitgitgadget/blob/c4805370f59532aa438283431b8ea7d4484c530f/lib/patch-series.ts#L271

Care to suggest how to fix this properly? (IIRC there are some sort of length limitations on `Cc:` lines, too, and they are not necessarily comma separated, continuations are usually indicated by the next line starting with a space, right?)

dscho:
@sunshineco I could really use your help here, as I am rather swamped with other stuff...

sunshineco:
I think you're correct about line length limit, commas, and continuation lines, based upon my reading of that RFC.

[...]

PR #18, opened 2018-07-14
Avoid duplicate Cc: lines

dscho:
It was reported in https://github.com/gitgitgadget/gitgitgadget/issues/16 that GitGitGadget produces mails with duplicate Cc: lines, which could be interpreted by anti-malware software to indicate non-benign emails.

So let's fix this.

/cc @ sunshineco


derrickstolee:
I don't understand the implementation, but I understand the tests that achieve what we want.

PR #22, opened 2018-07-17
send-mail: use current date

derrickstolee:
We are trying to emulate 'git send-mail', and that command ignores the
date that was created by 'git format-patch' in favor of the current
time. This helps users keep their inbox more organized, especially when
they do not thread their messages.

Signed-off-by: Derrick Stolee <dstolee@microsoft.com>

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -16,7 +16,6 @@ into a few categories (listed by priority, most important tasks first).
 
 ## Tasks that would be really nice to have, too, time permitting
 
-- Imitate `git send-email` (which apparently overrides the `Date:` header and uses the current time instead)
 - The branches merged into `pu` are also pushed individually to
   https:/github.com/gitster/git. We will want to add a comment to the PR every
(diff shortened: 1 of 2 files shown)


dscho:
Oh, one more thing: I just figured out that we need to use `new Date().toUTCString()` (to generate an RFC 2822 compliant date) instead of `Date.now().toString()` (which output some epoch-like number).

dscho:
Superseded by #23. Let's 🤞 that it works as intended.

dscho:
I guess, from looking at the sent mails, that it does *not* quite work as intended: Gmail insists on overriding the Date: headers. I guess it really wants to see some `+0000` time zone information, and node.js seems to only produce that if there is a non-zero offset to UTC. Or Gmail does not like the "GMT" string. Or whatever.

The good news is that Gmail overrides the date with the current one, so we kinda still get what we wanted (because it is slow enough that each mail seems to take at least one second).

PR #24, opened 2018-07-19
Work harder to avoid duplicate Cc: header lines

### LINKED ISSUE #16: multiple "Cc:" headers

dscho:
We already have code that tries to reuse existing Cc: header lines
instead of adding a new one.

And we already fixed the situation where our code failed to anticipate the situation when the `Cc:` header is the last one in the header block.

But we still failed to handle the case correctly where the existing Cc:
header was already a multi-line one.

Let's also handle that case.

This hopefully fixes #16 for good.

Signed-off-by: Johannes Schindelin <johannes.schindelin@gmx.de>

### DIFF
--- a/lib/patch-series.ts
+++ b/lib/patch-series.ts
@@ -264,7 +264,8 @@ export class PatchSeries {
                 return;
             }
 
-            const ccMatch = header.match(/^([^]*\nCc: .*?)(|\n(?![ \t])[^]*)$/);
+            const ccMatch =
+                header.match(/^([^]*\nCc: [^]*?)(|\n(?![ \t])[^]*)$/);
             if (ccMatch) {
                 header = ccMatch[1] + ",\n    " + authorMatch[2] + ccMatch[2];
             } else {


dscho:
/cc @sunshineco for visibility.

FWIW this is a bit of an advanced regex (@derrickstolee your partner will be delighted to learn that I made this mistake despite considering myself not exactly a beginner on the topic). The `.` usually stands for "any character" in regexes, but in *multi-line* regexes, `.` frequently means "any character except newline". The latter is the case in Javascript's regexes. To match "any character, including newlines", one needs to use `[^]` instead.

And in this case, I mixed them up: the first group ends in a reluctant pattern so that the second pattern either matches the end of the string (which contains the mail's header without the trailing newline), or the next header line (which must not start with whitespace).

The first group also needs to match newline characters, and the reluctant nature of the pattern ensures that we do not match the next header line (if any).

derrickstolee:
Perhaps a test that demonstrates the previous bug so we know it's fixed and won't regress. Without adequate test coverage, we will not be able to safely edit this regex in the future.

derrickstolee:
Merging now so I can submit a patch with GGG. We will revisit the test next week.

dscho:
> Perhaps a few tests in
> this area would help. It might also help

Okay. Two esteemed colleagues are *well* enough to convince me of this.

Privately with @derrickstolee we agreed that it is not worth holding up this PR, but I will try to spend [OneWeek](https://news.microsoft.com/life/one-week-microsoftlife/) on GitGitGadget, including adding tests for this.

sunshineco:
FWIW: I do grok the regex and verified that it works as expected. Indeed, tests would be a good idea to protect against regression, but I'm reasonably convinced that this patch fixes the immediate problem.

In the long term, if other bugs crop up or other header management becomes necessary, it might make sense to do header processing in a more traditional fashion (line-by-line, slurping up continuation lines as needed, etc.), rather than via tricky regex's with their attendant high cognitive load. Of course, tests would need to exist before such a rewrite.

PR #27, opened 2018-08-03
Wrap the cover letters

dscho:
It was pointed out that long lines are hard to read, and the convention for Markdown (and hence, for PR descriptions on GitHub) is to have paragraphs in one long line, to be rendered (wrapped) when displayed.

When sending mails to the Git mailing list, it is expected that that rendering already happened. And this PR makes it so (as demonstrated by the updates to the tests).

### DIFF
--- a/lib/patch-series.ts
+++ b/lib/patch-series.ts
@@ -4,6 +4,7 @@ import {
 import { GitNotes } from "./git-notes";
 import { GitGitGadget } from "./gitgitgadget";
 import { IMailMetadata } from "./mail-metadata";
+import { md2text } from "./markdown-renderer";
 import { IPatchSeriesMetadata } from "./patch-series-metadata";
 import { PatchSeriesOptions } from "./patch-series-options";
 import { ProjectOptions } from "./project-options";
@@ -153,8 +154,10 @@ export class PatchSeries {
         const project = await ProjectOptions.get(workDir, headCommit, cc || [],
             basedOn, publishToRemote, baseCommit);
 
+        const wrapCoverLetterAtColumn = 76;
         return new PatchSeries(notes, options, project, metadata,
-            rangeDiff, coverLetter, senderName);
+            rangeDiff, md2text(coverLetter, wrapCoverLetterAtColumn),
+            senderName);
     }
 
     protected static parsePullRequestDescription(description: string): {
--- a/tests/gitgitgadget.test.ts
+++ b/tests/gitgitgadget.test.ts
@@ -25,8 +25,8 @@ MIME-Version: 1.0
 To: reviewer@example.com
 Cc: Some Body <somebody@example.com>
 
-This Pull Request contains some really important changes that I would love to
-have included in git.git.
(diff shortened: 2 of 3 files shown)


lib/patch-series.ts
```diff
@@ -154,7 +155,7 @@ export class PatchSeries {
             basedOn, publishToRemote, baseCommit);
 
         return new PatchSeries(notes, options, project, metadata,
-            rangeDiff, coverLetter, senderName);
+            rangeDiff, md2text(coverLetter, 76), senderName);
```
whoisj: `76` what? Looks like you're just moving this "magic number" down the road.

nit: avoid magic numbers

tests/gitgitgadget.test.ts
```diff
@@ -25,8 +25,8 @@ MIME-Version: 1.0
 To: reviewer@example.com
 Cc: Some Body <somebody@example.com>
 
-This Pull Request contains some really important changes that I would love to
-have included in git.git.
+This Pull Request contains some really important changes that I would love
+to have included in git.git [https://github.com/git/git].
```
whoisj: Did you mean `[git.git](https://github.com/git/git)` ?

tests/gitgitgadget.test.ts
```diff
@@ -215,8 +216,8 @@ Cc: Some Body <somebody@example.com>
 
     expect(patches.coverLetter).toEqual(`My first Pull Request!
 
-This Pull Request contains some really important changes that I would love to
-have included in git.git.`);
+This Pull Request contains some really important changes that I would love
+to have included in git.git [https://github.com/git/git].`);
```
whoisj: Why sometimes `git.git [https://github.com/git/git]` and others `[git.git](https://github.com,/git/git)` ?

Trying to understand the intention.

lib/patch-series.ts
dscho: It's the `columns` parameter of the `md2text()` function...

whoisj: Yeah, but why `76`?

tests/gitgitgadget.test.ts
dscho: No, this is the output of the renderer. The renderer specifically does not use Markdown to show links, but shows link texts as plain text, followed by the link in brackets.

It is the default of the [`html-to-text`](https://www.npmjs.com/package/html-to-text) package that I use to do all the hard work.

tests/gitgitgadget.test.ts
dscho: These tests verify the output to be the plain text (with link targets in square brackets), and it needs to provide the input (the "Pull Request description") as Markdown.

So you see the link provided in Markdown *once*, and all the other times it is the text that tries to assure the output format is correct.

lib/patch-series.ts
dscho: To clarify (it was already explained in the commit message, but that was not obvious from the diff), I introduced a constant.

PR #51, opened 2018-12-21
Switch GitGitGadget to be backed by an Azure Function and an Azure Pipeline

dscho:
The previous design let the GitHub App that is GitGitGadget be backed by an Azure WebApp. However:

- The free tier does not offer any always-on mode, therefore it always took something like 2-5 minutes for the web app to be spun up.
- Logging was also a slight problem, as only one person (yours truly) could look at it and tell what problems occurred.
- Finally, if two PRs were `/submit`ted at roughly the same time, two jobs were started that used the same Git worktree, causing the first one "to win".

All of these problems go away by switching to an Azure Function/Pipeline combo. Which I already did, BTW, it simply runs from this branch until it is merged. The simple idea is: GitHub App triggers Azure Function, Azure Function does some real light-weight validation (like checking that GitHub sent the payload, that this event comes from https://github.com/gitgitgadget/git, that we're looking at a PR comment that contains a command) and then simply triggers the Azure Pipeline. This Pipeline retrieves the comment by ID, verifies that the user who authored that comment is in the permitted list, parses the command and then handles it.

One additional important benefit is that users now also get visual feedback in the form of a Check on their PR (which also offers a convenient way to access the logs).


dscho:
The first PR that went through with the new design: https://github.com/gitgitgadget/git/pull/104#issuecomment-449365244

Issue #38, opened 2018-10-04
multiple "Mime-Version:" headers


sunshineco:
Emails generated by `gitgitgadget` may contain multiple `Mime-Version:` headers. For example:

```
Message-Id: <13032d847533cfecbd9a912312cab4f256e46b30.1538586732.git.gitgitgadget@gmail.com>
Date:   Wed, 03 Oct 2018 10:12:17 -0700 (PDT)
In-Reply-To: <pull.42.v2.git.gitgitgadget@gmail.com>
References: <pull.42.git.gitgitgadget@gmail.com>
        <pull.42.v2.git.gitgitgadget@gmail.com>
From:   "=?UTF-8?q?Martin=20=C3=85gren?= via GitGitGadget" 
        <gitgitgadget@gmail.com>
Subject: [PATCH v2 2/3] builtin/commit-graph.c: UNLEAK variables
MIME-Version: 1.0
Content-Type: text/plain; charset=UTF-8
Content-Transfer-Encoding: 8bit
Content-Type: text/plain; charset=UTF-8
Content-Transfer-Encoding: 8bit
MIME-Version: 1.0
To:     git@vger.kernel.org
Cc:     Junio C Hamano <gitster@pobox.com>,
        =?UTF-8?q?Martin=20=C3=85gren?= <martin.agren@gmail.com>
Sender: git-owner@vger.kernel.org
List-ID: <git.vger.kernel.org>
X-Mailing-List: git@vger.kernel.org
```

(Note that the `Content-Type:` and `Content-Transfer-Encoding:` headers are also unnecessarily duplicated.)

According to the BNF grammar of [RFC 2045 § 3](https://tools.ietf.org/html/rfc2045#section-3), it appears that only a single top-level `Mime-Version:` header is allowed.

Moreover, multiple `Mime-Version:` headers may cause delivery problems. For instance, anti-spam software may reject such email:

```
X-Amavis-Alert: BAD HEADER SECTION, Duplicate header field: "MIME-Version"
X-Quarantine-ID: <...>
```

dscho:
I had a look at the latest mail sent by GitGitGadget and it appears not to have that problem: https://public-inbox.org/git/2b127d9669aa7b73ced7611b6e77044f5efed11d.1538595818.git.gitgitgadget@gmail.com/raw

[...]

PR #54, opened 2018-12-21
Use correctly-encoded email headers for names with non-ASCII characters

dscho:
I ran into this issue a couple times when Cc:ing Ævar: if the Cc: header contains unencoded non-ASCII characters, the Git mailing list will simply reject that mail.

So let's encode those characters (using nodemailer's internal `encodeWords()` function).

This fixes https://github.com/gitgitgadget/gitgitgadget/issues/29.

PR #55, opened 2018-12-22
Drop extra `MIME-Version` headers

dscho:
As explained in https://github.com/gitgitgadget/gitgitgadget/issues/38, sometimes we end up with mails that have multiple `MIME-Version` headers.

In order to avoid being filed away in the spam folder, let's just post-process the generated mails to remove duplicate `MIME-Version` headers.


dscho:
Please note that this PR is based on #54 and therefore needs that PR to be merged first.

dscho:
@sunshineco any other headers you see that we should not duplicate, either (even if we don't do that right now anyway)?

sunshineco:
> @sunshineco any other headers you see that we should not duplicate, either?

Based upon my reading of [RFC 2045](https://tools.ietf.org/html/rfc2045#section-3), the following message headers must be singletons:

```
Content-Description:
Content-ID:
Content-Transfer-Encoding:
Content-Type:
MIME-Version:
```

Issue #63, opened 2018-12-26
Upgrade the Function and Pipelines to use node.js v10.x


dscho:
V8.x will be retired in Dec '19. Upgrade the Function and all the Pipelines.

dscho:
I upgraded the remaining (currently 2) Pipelines, too, so this ticket is closed.

PR #64, opened 2018-12-26
ci: upgrade to node.js v10.x

dscho:
In theory, the (very simple) GitGitGadget code should compile easily
with any node.js version that supports ES6. In practice, only a proper
test suite can give us enough confidence that that is the case.

Luckily, even GitGitGadget's rather sparse test suite covers enough to
give us that confidence, and we should be able to upgrade to node.js
v10.x without issues (because v8.x is expected to not receive any
regular updates anymore come April '19, and not even security updates
come December '19).

This addresses part of https://github.com/gitgitgadget/gitgitgadget/issues/63.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -12,7 +12,7 @@ pool:
 steps:
 - task: NodeTool@0
   inputs:
-    versionSpec: '8.x'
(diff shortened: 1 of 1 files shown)


dscho:
Whoa, that was fast!

Issue #52, opened 2018-12-21
Greet new users


dscho:
Now that GitGitGadget is backed by an Azure Pipeline that is registered as PR build, we can easily say something when a new PR is opened by somebody who is not (yet) permitted to `/submit`, with a nice lil' greeting and a short explainer and links and everything.

PR #69, opened 2019-01-14
Greet new users with a welcome message

dscho:
When a new contributor is told "just use GitGitGadget, you only need to open a PR at https://github.com/gitgitgadget/git", that's all nice and dandy, but said contributor is unlikely to know what to do next.

So let's tell them.

This fixes https://github.com/gitgitgadget/gitgitgadget/issues/52.


dscho:
I [pretended not to be allowed to use GitGitGadget](https://github.com/gitgitgadget/gitgitgadget/pull/69/commits/eee7e47ab2c7471b4aff366d9afc35b394885d49) and opened a new gitgitgadget/git PR, so you can see the welcome message here: https://github.com/gitgitgadget/git/pull/107#issuecomment-454013944

PR #72, opened 2019-01-25
Mirror answers on the Git mailing list to the corresponding PRs

dscho:
So far, GitGitGadget has been a one-way street: commits pushed to GitGitGadgets' Git fork could be sent to the Git mailing list, but the communication about the patches had to happen there, completely.

With this PR, the answers on the Git mailing list get added as PR comments.

While the format is far from perfect (the emails are rendered in a rather ugly way, non-ASCII senders are not displayed properly, for replies to individual patches, there is no obvious link to the respective commit, etc), it is a start.

This can be seen in action at https://github.com/gitgitgadget/git/pull/31#issuecomment-457692752

### DIFF
--- a/lib/ci-helper.ts
+++ b/lib/ci-helper.ts
@@ -7,6 +7,7 @@ import { GitHubGlue } from "./github-glue";
 import { MailCommitMapping } from "./mail-commit-mapping";
 import { IMailMetadata } from "./mail-metadata";
 import { IPatchSeriesMetadata } from "./patch-series-metadata";
+import { PublicInboxGitHelper } from "./public-inbox-helper";
 
 const readFile = util.promisify(fs.readFile);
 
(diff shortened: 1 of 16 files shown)


dscho:
@jeffhostetler the code in this PR generated https://github.com/gitgitgadget/git/pull/108#issuecomment-457708527 (if you wondered about that).

dscho:
@jeffhostetler I know you're not necessarily familiar with the GitGitGadget code base, but maybe you could give this PR a look-over?

dscho:
> I was wondering if there are any tests for gitgitgadget because there is a lot of parsing going on and tests would really help validate it.

Indeed, it gives a bit more confidence to have tests covering this functionality.

The part that I think is the only part that is testable, though, is the parsing of the mbox file (which is already tested). The more important part is where the code actually acts on mails sent to the mailing list, which I think is not really testable, not unless I mock quite a bit of stuff into the code... which would cost me a lot of time, and might actually be buggy in and of itself? So I'll try to be lazy here ;-)

Thank you for the review!

PR #84, opened 2019-04-12
Use `git range-diff` to identify upstream commits

dscho:
To determine when a PR made it into an integration branch in core Git, we need to identify corresponding commits (if any).

Earlier, we tried to do that by tracking the mails' progress, identifying pairs of mails on the GIt mailing list and their corresponding commits in core Git by the date, with the idea of fixing things manually if they go wrong. Turns out they go wrong a lot.

So let's use `range-diff` instead. This takes a bit more time, but should be a lot more robust, too.

### DIFF
--- a/lib/ci-helper.ts
+++ b/lib/ci-helper.ts
@@ -260,8 +260,11 @@ export class CIHelper {
             return [false, false];
         }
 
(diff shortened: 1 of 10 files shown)


dscho:
@kewillford thank you!

Issue #74, opened 2019-01-27
Fix duplicate mail headers


dscho:
As @sunshineco mentioned [here](https://github.com/gitgitgadget/gitgitgadget/pull/56#commitcomment-31781797) and [here](https://github.com/gitgitgadget/gitgitgadget/pull/55#commitcomment-31781797), there should be at most one `Content-Description:`, `Content-ID:`, `Content-Type:` and `Content-Transfer-Encoding:` header, [just like the `MIME-Version` header](https://github.com/gitgitgadget/gitgitgadget/pull/55).

sunshineco:
Thanks for opening this ticket. I was just wondering the other day if #38 should be re-opened, but opening a new ticket makes more sense.

webstech:
I have a proposed fix on the DupMailHdrs branch of my project. It expands the MIME-Version check to include the other headers mentioned.  If the values are different, an error is thrown.  There is support to allow 'known' values and not throw an error.

Should I create a pull request or can I get feedback on the changes in the branch before creating a pull request?  Still learning protocol.

thanks,
...chris.

sunshineco:
Thanks for working on this. I left some comments on your changes to the code in e8613484f556c204c608e2463ba40ee1baab7143. The comments are mostly minor and may or may not even be actionable (depending upon whether or not I'm understanding the code correctly).

dscho:
@webstech thank you for jumping on this!

> Should I create a pull request

Please do!

webstech:
@sunshineco Thanks for the comments.  I have added responses.

In response I have refactored the code to make the positioning clearer (I hope) and added tests to help prove it is working.

Is it preferred that the code updates be in a new commit or just amend the prior one?  Waiting for the answer to this before pushing.  Also outstanding is the question of whether the DRY stuff (~line 770) should be removed/reverted.

thanks,
...chris.

sunshineco:
> Is it preferred that the code updates be in a new commit or just amend the prior one? Waiting for the answer to this before pushing. 

Amending the existing commit makes it easier for people to review and understand the changes as a whole (rather than having to first understand one patch which itself has problems, followed by another patch which fixes those problems).

> Also outstanding is the question of whether the DRY stuff (~line 770) should be removed/reverted.

[...]

PR #89, opened 2019-05-18
Fix duplicate mail headers #74

webstech:
Changed removeDuplicateMimeVersionHeaders to process a table of headers that should be singletons (renamed method).  If a duplicate header value is different and not an allowed variant, a message is logged.  The same table is now used to generate the headers in generateMBox.

### DIFF
--- a/lib/patch-series.ts
+++ b/lib/patch-series.ts
@@ -16,6 +16,37 @@ export interface ILogger {
 
 export type SendFunction = (mail: string) => Promise<string>;
 
+interface ISingletonHeader {
+    key: string;
+    values: string[];
+}
+
+// NOTE: first values is used when emitting headers in addSingletonHeaders
+// unless it is an empty string
+const singletonHeaders: ISingletonHeader[] = [
+    {
+        key: "Content-Description",
+        values: [],
+    },
+    {
+        key: "Content-ID",
+        values: [],
+    },
+    {
+        key: "Content-Type",
+        values: ["text/plain; charset=UTF-8", "text/plain; charset=\"UTF-8\"",
+                 "text/plain; charset=utf-8", "text/plain"],
+    },
+    {
+        key: "Content-Transfer-Encoding",
+        values: ["8bit", "7bit"],
+    },
+    {
+        key: "MIME-Version",
+        values: ["1.0"],
+    },
+];
+
 export class PatchSeries {
     public static async getFromTag(options: PatchSeriesOptions,
                                    project: ProjectOptions):
@@ -231,29 +262,68 @@ export class PatchSeries {
         return mbox.split(separatorRegex);
     }
 
-    protected static removeDuplicateMimeVersionHeaders(mails: string[]): void {
+    protected static removeDuplicateHeaders(mails: string[]): void {
         mails.map((mail: string, i: number) => {
             const endOfHeader = mail.indexOf("\n\n");
             if (endOfHeader < 0) {
                 return;
             }
+
             let headers = mail.substr(0, endOfHeader + 1);
-            const needle = "\nMIME-Version: 1.0\n";
-            const offset = headers.indexOf(needle);
-            if (offset < 0) {
-                return;
-            }
-            let offset2 = headers.indexOf(needle, offset + 1);
-            if (offset2 < 0) {
-                return;
-            }
-            do {
-                headers = headers.substr(0, offset2 + 1)
-                    + headers.substr(offset2 + needle.length);
-                offset2 = headers.indexOf(needle, offset + 1);
-            } while (offset2 > 0);
+            singletonHeaders.forEach((header: ISingletonHeader) => {
+                headers = PatchSeries.stripDuplicateHeaders(headers, header);
+            });
+
             mails[i] = headers + mail.substr(endOfHeader + 1);
         });
+    }
+
+    protected static stripDuplicateHeaders(headers: string,
+                                           header: ISingletonHeader): string {
+        const needle = "\n" + header.key + ":";
+        let offset;
+
+        if (headers.startsWith(`${header.key}:`)) {
+            offset = 0;
+        } else {
+            offset = headers.indexOf(needle) + 1;
+            if (!offset) {
+                return headers;
+            }
+        }
+
+        let endOfKey = offset + needle.length - 1;
+        offset = headers.indexOf(needle, endOfKey);
+
+        if (offset < 0) {
+            return headers;
+        }
+
+        // extract values to determine if they match.
+        let endOfHdr = headers.indexOf("\n", endOfKey);
+        const value1 = headers.substr(endOfKey, endOfHdr - endOfKey).trim();
+
+        do {
+            endOfKey = offset + needle.length;
+            endOfHdr = headers.indexOf("\n", endOfKey);
+            const value2 = headers.substr(endOfKey,
+                                          endOfHdr - endOfKey).trim();
+
+            if (value1 !== value2) {
+                if (0 >= header.values.indexOf(value2)) {
+                    console.log("Found multiple headers where only one allowed"
+                        + `\n    ${header.key}: ${value1}\n    `
+                        + `${header.key}: ${value2}\nProcessing headers:\n`
+                        + headers);
+                }
+            }
+
+            // substr up to \n and concat from next \n
+            headers = headers.substr(0, offset) + headers.substr(endOfHdr);
+            offset = headers.indexOf(needle, offset);
+        } while (offset >= 0);
+
+        return headers;
     }
 
     protected static insertCcAndFromLines(mails: string[], thisAuthor: string,
@@ -433,6 +503,18 @@ export class PatchSeries {
         return count;
     }
 
+    protected static generateSingletonHeaders(): string[] {
+        const results: string[] = [];
+
+        for (const key of singletonHeaders) {
+            if (key.values.length) {
+                results.push(`--add-header=${key.key}: ${key.values[0]}`);
+            }
+        }
+
+        return results;
+    }
+
     public readonly notes: GitNotes;
     public readonly options: PatchSeriesOptions;
     public readonly project: ProjectOptions;
@@ -482,7 +564,7 @@ export class PatchSeries {
         logger.log("Generating mbox");
         const mbox = await this.generateMBox();
         const mails: string[] = PatchSeries.splitMails(mbox);
-        PatchSeries.removeDuplicateMimeVersionHeaders(mails);
+        PatchSeries.removeDuplicateHeaders(mails);
 
         const ident = await git(["var", "GIT_AUTHOR_IDENT"], {
             workDir: this.project.workDir,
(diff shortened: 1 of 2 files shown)


webstech:
Apologies for not doing the pull request before the first amended commit.  Updated based on comments (thank you @dscho and @sunshineco).

Set these in tsconfig.json to get around errors running tests before changes.
    "noImplicitAny": false,
     "strictNullChecks": false,

lib/patch-series.ts
```diff
@@ -231,31 +262,69 @@ export class PatchSeries {
...
+                return headers;
+            }
+        }
+
+        let endofKey = offset + needle.length - 1;
+        offset = headers.indexOf(needle, endofKey + 1);
+
+        if (offset < 0) {
+            return headers;
+        }
```
sunshineco: Now we check if `needle` appears a second time, following the first instance. If it doesn't, there's no duplicate, so we're done and exit early.

However, I think this logic (due to the `+1`) can fail in a pathological case, such as input like this:

```
MIME-Version:\n
MIME-Version: 1.0\n
```

In this case, `endofKey` points at the newline following the `:`, and the `+1` bumps it past the newline, so it won't match `needle` even though there clearly is a second instance of the header. The easy fix is simply to drop the `+1`.

lib/patch-series.ts
```diff
@@ -231,31 +262,69 @@ export class PatchSeries {
...
+
+        // extract values to determine if they match.
+        let endOfHdr = headers.indexOf("\n", endofKey);
+        const value1 = headers.substr(endofKey, endOfHdr - endofKey).trim();
+
+        do {
+            endofKey = offset + needle.length;
```
sunshineco: At this point, `offset` is pointing at the newline preceding the next instance of `key`, so `endofKey` points just after the trailing `:`. Okay.

lib/patch-series.ts
```diff
@@ -231,31 +262,69 @@ export class PatchSeries {
...
+            endOfHdr = headers.indexOf("\n", endofKey);
+            const value2 = headers.substr(endofKey,
+                                          endOfHdr - endofKey).trim();
+
+            if (value1 !== value2) {
+                if (0 >= header.values.indexOf(value2)) {
+                    console.log("Header values do not match for " +
+                        header.key + ": " + value1 +
+                        ":" + value2 + "\n" + headers);
```
sunshineco: Nit: The `:` separating the two values may be confusing for the reader:

```
Header values do not match for MIME-Version: 1.0:2.0
```

Further, the reader isn't told about the more serious problem, which is that multiple instance of what should be a singleton were found. Perhaps the entire message could be replaced with something like:

```
Found multiple headers where only one allowed:
    MIME-Version: 1.0
    MIME-Version: 2.0
```

lib/patch-series.ts
```diff
@@ -231,31 +262,69 @@ export class PatchSeries {
...
+                }
+            }
+
+            // substr up to \n and concat from next \n
+            headers = headers.substr(0, offset) + headers.substr(endOfHdr);
```
sunshineco: `offset` points at the newline just before the instance of `key` which we want to remove, so the first `substr()` grabs everything up to but not including that newline; `endOfHdr` points at the newline following the value of the instance of `key` we want to remove, so the second `substr()` grabs that newline and everything following it. Okay.

lib/patch-series.ts
```diff
@@ -231,31 +262,69 @@ export class PatchSeries {
...
+            offset = headers.indexOf(needle, offset);
```
sunshineco: At first glance, it seems like we should be skipping ahead and starting the next search at `endOfHdr`, however, we just removed `\nKey: value`, so `endOfHdr` is now meaningless. Therefore, we need to start the new search at the last point we've examined in the now-shortened string, which is `offset`. Okay.

lib/patch-series.ts
```diff
@@ -433,6 +502,17 @@ export class PatchSeries {
         return count;
     }
 
+    protected static addSingletonHeaders(): string[] {
```
sunshineco: Nit: I think I may be the one who suggested this name, but in the context where it's used as the argument to `concat()` in `const args = [...].concat(...)`, a more meaningful name would be simply `singletonHeaders()`.

sunshineco:
By the way, the cognitive load associated with this code is quite high due to its insistence upon performing these operations on a relatively complex _string_. The reader is required to maintain quite a lot of state in his/her head, and then painstakingly manually apply that state to various constructs to understand what is going on or to verify correction operation. This does not lend itself to code review or code changes.

A good way forward from here (if someone wants a project to work on) would be to treat headers as first-class entities rather than performing all this complex work on a string representing the entire mail message. For instance, it is easy to envision a function which takes the full email message as a string and decomposes it into three parts: (1) _envelope_, (2) _headers_, and (3) _body_, where _envelope_ is the leading "`From `" line, _headers_ is an array of already-parsed key/value tuples, and _body_ is the body as a simple string. Operations on headers would then become quite easy. For instance, removing duplicate headers becomes the simple operation of iterating over the key/value tuples looking for duplicates, and removing them when found. Finally, a separate function would reassemble the _envelope_, _headers_, and _body_ back into a string for final submission.

Parsing (and perhaps unparsing) the headers is slightly complicated by _continuation_ lines. Some decision needs to be made about how to treat them internally (whether to preserve the split point or to keep both split and glued representations), and consumers of headers should be prepared to deal with fallout from that decision.

dscho:
> By the way, the cognitive load associated with this code is quite high due to its insistence upon performing these operations on a relatively complex _string_.

That is my fault: the original code was written in Unix shell script, and it shows.

Somebody with a lot of experience in Typescript once looked over one method of this code base and offered that it would be a lot easier to read/maintain/debug if it was a *lot* more functional.

And I agree... I am not at all opposed to ripping apart the existing code, making it a lot more functional, exploding it into a gazillion small, well-defined functions.

lib/patch-series.ts
```diff
@@ -231,31 +262,69 @@ export class PatchSeries {
...
+        if (headers.startsWith(`${header.key}:`)) {
+            offset = 0;
+        } else {
+            offset = headers.indexOf(needle) + 1;
+            if (!offset) {
+                return headers;
+            }
+        }
+
+        let endofKey = offset + needle.length - 1;
```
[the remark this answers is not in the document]
webstech: Changing to endOfKey.

lib/patch-series.ts
webstech: Removing the +1.

lib/patch-series.ts
webstech: Changing to suggested message test.

lib/patch-series.ts
webstech: Changing to generateSingletonHeaders.

lib/patch-series.ts
```diff
@@ -16,6 +16,37 @@ export interface ILogger {
...
+    },
+    {
+        key: "Content-Transfer-Encoding",
+        values: ["8bit", "7bit"],
+    },
+    {
+        key: "MIME-Version",
+        values: ["1.0"],
+    },
+];
```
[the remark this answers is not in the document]
webstech: Empty arrays are supported and changed to that.  I did not convert to dictionary style.

webstech:
Updated based on comments.  Converted some strings to template strings for cleaner look.

lib/patch-series.ts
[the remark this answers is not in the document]
sunshineco: > I did not convert to dictionary style.

As one whose mail server complains about these duplicate headers, I'm happy to see this issue getting fixed, so this choice is not at all a blocker for me. As a reviewer/programmer, though, I'm curious to know your reasoning for this decision.

sunshineco:
> Re; code overload - the code is assuming the headers are mostly correct and simple (ie no multi-line for the headers being looked at).

Just to be clear, I wasn't suggesting that this (already written) patch should be changed to do major code refactoring; rather, I was just offering a way forward if someone (else) wants to tackle additional changes in the future.

Regarding multi-line headers, even though the headers consulted by this particular code won't be multi-line, `git format-patch` can and does generate multi-line headers, so code which parses headers would, in general, need to take that into account. But that's an issue for some future change; not relevant to this patch.

> A regex could have been used to extract the values but the current code to determine if there are duplicates first is probably more efficient.

For as infrequently as this code executes, efficiency is not likely a major concern, whereas ease of review and ease of understanding the code for someone modifying it, reduced cognitive load would be a win in the long run.

> There is a lot to look at, including removeDuplicateHeaders being called with a set of mails. I targeted one problem and wanted to report if there was a possible value conflict.

Thanks for working on this problem. As noted, the parsing/refactoring suggestion was just that, a suggestion for the future if someone wants to tackle it, not a request that you do so for this particular fix.

lib/patch-series.ts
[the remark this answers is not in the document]
webstech: singletonHeaders.forEach feels more object oriented to me, passing an object/structure to stripDuplicateHeaders.  The choice was to pass the key and array or just the key and have stipDuplicateHeaders access the global.

dscho:
> > A regex could have been used to extract the values but the current code to determine if there are duplicates first is probably more efficient.
> 
> For as infrequently as this code executes, efficiency is not likely a major concern, whereas ease of review and ease of understanding the code for someone modifying it, reduced cognitive load would be a win in the long run.

There is ample evidence that developers get regexes wrong _all the time_. As a maintainer/reviewer, I therefore have to inspect them all the time, to make sure that they are correct. And even regular readers have to wrap their heads around regular expressions (often getting it wrong).

In short: I don't think that regular expressions are a good way forward, unless they are *super simple*. (Which they are not in GitGitGadget, and neither in the proposed way forward.)

sunshineco:
> I don't think that regular expressions are a good way forward, unless they are _super simple_. (Which they are not in GitGitGadget, and neither in the proposed way forward.)

The proposal of a proper parser to extract headers into an array of key/value tuples need not be implemented via regular expressions (it was certainly never my intention to suggest such, and is not the way I would write a parser). And, once the headers are available as an array of key/value tuples, duplicate-detection should involve only simple string comparison, not regular expressions. So, regular expressions were never intended as a component of the proposal for future work on this topic.

webstech:
Are there any other changes needed for this pull request?

sunshineco:
> Are there any other changes needed for this pull request?

I don't think I found any blocking problems the last time I reviewed this patch. (There was a minor stylistic difference of opinion about whether `singletonHeaders` should be a simple array and `ISingletonHeaders` dropped altogether, but that doesn't warrant holding up the patch, and, as you're the one doing the actual work, your opinion matters more than mine there.)

As far as I understand, the only thing needed for this patch to be accepted is for someone with _write_ privilege to review the patch (either @dscho or @derrickstolee, I think), but I suppose it's relatively low-priority compared to other work they're doing. (@dscho, at least, tends to have a lot on his plate.)

dscho:
@webstech I owe you a honking big apology, sorry for letting this slip so long!

I will have a quick look over it right now.

dscho:
@webstech thank you so, so, so much for your contribution! Finally that bug is fixed!

@sunshineco thank you for your thorough and helpful review, too!

PR #98, opened 2019-06-25
Quote names in emails if necessary

dscho:
Hopefully this fixes the problem that prevented the mails from https://github.com/gitgitgadget/git/pull/151 to be delivered correctly.

If only vger provided better diagnostics to us.

### DIFF
--- a/tests/patch-series.test.ts
+++ b/tests/patch-series.test.ts
@@ -89,6 +89,13 @@ class PatchSeriesTest extends PatchSeries {
             expect(mails[0]).toEqual(expect.stringContaining(needle));
         });
 
+        test("sender name with a dot in it is enclosed in double quotes", () => {
+            const withDots = "wile e. coyote <coyote@desert.net>";
+            const expected = "\"wile e. coyote\" <coyote@desert.net>";
+            const quoted = PatchSeries.encodeSender(withDots);
+            expect(quoted).toEqual(expected);
+        });
+
         test("Cc: is inserted correctly", () => {
             expect(mails[1]).toMatch(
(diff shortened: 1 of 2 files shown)


dscho:
@sunshineco could I ask you to review this?

tests/patch-series.test.ts
```diff
@@ -89,6 +89,13 @@ class PatchSeriesTest extends PatchSeries {
             expect(mails[0]).toEqual(expect.stringContaining(needle));
         });
 
+        test("sender name with a dot in it is enclosed in double quotes", () => {
+            const withDots = "wile e. coyote <coyote@desert.net>";
```
derrickstolee: An excellent test!

sunshineco:
Are we confident that only _period_ is problematic in the name? My quick scan of [RFC2822 §3.2.1](https://tools.ietf.org/html/rfc2822#section-3.2.1) suggests that there are several other characters which might be problematic for _vger_.

Why not take the simpler approach of quoting the name unconditionally rather than doing so only if it contains a period? (Or, perhaps quote it only if it does not already appear to be quoted -- by checking if the first character is a double quote.)

Aside from the potential fragility of triggering only on period, the patch itself looks okay.

dscho:
> My quick scan of [RFC2822 §3.2.1](https://tools.ietf.org/html/rfc2822#section-3.2.1) suggests that there are several other characters which might be problematic for _vger_.

That's a good point. I tried to address this in #99.

> Why not take the simpler approach of quoting the name unconditionally rather than doing so only if it contains a period?

I _really_ dislike unnecessary quoting. That would land you in quoting hell. :-)

PR #99, opened 2019-06-28
Make email name quoting more complete

dscho:
This PR tries to address the shortcomings pointed out in https://github.com/gitgitgadget/gitgitgadget/pull/98#issuecomment-506617659.

### DIFF
--- a/lib/patch-series.ts
+++ b/lib/patch-series.ts
@@ -331,23 +331,20 @@ export class PatchSeries {
 
     protected static encodeSender(sender: string): string {
         const encoded = encodeWords(sender);
-        const dot = encoded.indexOf(".");
-        if (dot < 0) {
+
+        /* Don't quote if already quoted */
+        if (encoded.startsWith("\"") && encoded.match(/"\s*</)) {
             return encoded;
         }
 
-        let endOfName = encoded.indexOf("<");
-        if (endOfName < 0) {
-            endOfName = encoded.length;
-        } else {
-            while (endOfName > 0 && encoded[endOfName - 1] === " ") {
-                endOfName--;
-            }
-        }
-        if (dot >= endOfName) {
+        const match =
+            encoded.match(/^([^<]*[()<>[\]:;@\\,."][^<]*?)(\s*)(<.*)/);
+        if (!match) {
             return encoded;
         }
-        return `"${encoded.substr(0, endOfName)}"${encoded.substr(endOfName)}`;
+
+        // tslint:disable-next-line:max-line-length
+        return `"${match[1].replace(/["\\\\]/g, "\\$&")}"${match[2]}${match[3]}`;
     }
 
     protected static insertCcAndFromLines(mails: string[], thisAuthor: string,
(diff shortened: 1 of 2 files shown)


lib/patch-series.ts
```diff
@@ -328,23 +328,19 @@ export class PatchSeries {
 
     protected static encodeSender(sender: string): string {
         const encoded = encodeWords(sender);
-        const dot = encoded.indexOf(".");
-        if (dot < 0) {
+
+        /* Don't quote if already quoted */
+        if (encoded.startsWith("\"") && encoded.indexOf("\" <") > 0) {
```
sunshineco: I would have expected this to match whitespace loosely (just like the check later in the patch for a _special_ character in the name):

```
    ... && encoded.match(/"\s*</) {
```

lib/patch-series.ts
```diff
@@ -328,23 +328,19 @@ export class PatchSeries {
...
-        if (endOfName < 0) {
-            endOfName = encoded.length;
-        } else {
-            while (endOfName > 0 && encoded[endOfName - 1] === " ") {
-                endOfName--;
-            }
-        }
-        if (dot >= endOfName) {
+        const match =
+            encoded.match(/^([^<]*[()<>[\]:;@\\,."][^<]*?)(\s*)(<.*)/);
```
sunshineco: This is a bit difficult to reason about but I _think_ it's correct. To make it more obvious that this works as intended, perhaps in addition to the test @derrickstolee suggested, test cases could be added which verify that these do not get quoted:

```
name <email>
my name <email>
excited! <email> 
```

and these get quoted:

```
mr. name <email>
foo [bar] name <email>
bob@obo <email>
blug <nobody <email>
```

(I'm not sure the last one has any value; it almost certainly isn't even legal so perhaps it doesn't deserve to be tested.)

lib/patch-series.ts
```diff
@@ -328,23 +328,19 @@ export class PatchSeries {
...
+        if (!match) {
             return encoded;
         }
-        return `"${encoded.substr(0, endOfName)}"${encoded.substr(endOfName)}`;
+
+        return `"${match[1].replace(/"/g, "\\\"")}"${match[2]}${match[3]}`;
```
sunshineco: Escaping embedded double quotes make sense. Do we also need to escape backslashes?

lib/patch-series.ts
dscho: Makes sense! Thanks.

lib/patch-series.ts
dscho: I added all of them, and also one containing a backslash.

dscho: I added all of them, and also one containing a backslash.

dscho: I added all of them, and also one containing a backslash.

lib/patch-series.ts
dscho: Yes, I think so. I added it. And a test.

dscho:
> Do you have a test to check the "Don't quote if already quoted" logic?

Now I do ;-)

dscho:
> Do you have a test to check the "Don't quote if already quoted" logic?

Now I do ;-)

Issue #94, opened 2019-06-05
Be more lenient with white-space in PR comments


dscho:
When issuing an `/allow hello ` (with a trailing space), GitGitGadget is a bit dense, and adds "hello " (with a trailing space) to the list of allowed users.

We might even want to go one step further and make sure that that user exists.

Another step further might be to handle `/allow` without arguments to mean the author of the current PR.

Yet another step would add code to maintain an internal map from GitHub handles to clear names and email addresses, so that PR authors are auto-Cc:ed on their submitted patches.

webstech:
Pull request #101 was opened to address the first three parts of this.

The last step should probably be a separate issue.  Github may not have a public email address for the PR author.  GitHubGlue could provide an interface to what is known:
export interface IGHUser {
    email: string;                  // null if no public email
    login: string;
    name: string;
    type: string;
}

webstech:
@dscho Can this be closed now?

dscho:
> Can this be closed now?

Yes! Thanks!

PR #101, opened 2019-07-05
Trim spaces, default to PR author, validate name

webstech:
1. Trim leading and trailing spaces on pull request comment.
2. Default (dis)allow to PR author.
3. Verify user is valid for allow.

This addresses the first three items in issue #94 but not the last one.

### DIFF
--- a/lib/ci-helper.ts
+++ b/lib/ci-helper.ts
@@ -486,7 +486,7 @@ export class CIHelper {
      */
     public async handleComment(commentID: number): Promise<void> {
         const comment = await this.github.getPRComment(commentID);
-        const match = comment.body.match(/^(\/[-a-z]+)(\s+(.*))?$/);
+        const match = comment.body.match(/^\s*(\/[-a-z]+)(\s+(.*?))?\s*$/);
         if (!match) {
             console.log(`Not a command; doing nothing: '${comment.body}'`);
             return; /* nothing to do */
@@ -511,6 +511,11 @@ export class CIHelper {
                     comment.author} is not permitted to use GitGitGadget`);
             }
 
+            const getPRAuthor = async (): Promise<string> => {
+                const pr = await this.github.getPRInfo(comment.prNumber);
+                return pr.author;
+            };
+
             if (command === "/submit") {
                 if (argument && argument !== "") {
                     throw new Error(`/submit does not accept arguments ('${
@@ -523,7 +528,7 @@ export class CIHelper {
                     throw new Error("Only the owner of a PR can submit it!");
                 }
 
-                if (!pr || !pr.baseLabel || !pr.baseCommit ||
+                if (!pr.baseLabel || !pr.baseCommit ||
                     !pr.headLabel || !pr.headCommit) {
                     throw new Error(`Could not determine PR details for ${
                         pullRequestURL}`);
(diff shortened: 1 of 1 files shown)


lib/ci-helper.ts
```diff
@@ -486,13 +486,13 @@ export class CIHelper {
      */
     public async handleComment(commentID: number): Promise<void> {
         const comment = await this.github.getPRComment(commentID);
-        const match = comment.body.match(/^(\/[-a-z]+)(\s+(.*))?$/);
+        const match = comment.body.match(/^\s*(\/[-a-z]+)(\s+(.*?))\s*$/);
```
dscho: I think we still need the `?` after `(\s+(.*?))`, otherwise `/submit` would no longer be recognized...

lib/ci-helper.ts
```diff
@@ -511,19 +511,24 @@ export class CIHelper {
                     comment.author} is not permitted to use GitGitGadget`);
             }
 
+            const pr = await this.github.getPRInfo(comment.prNumber);
```
dscho: Hmm. This will add a roundtrip even if we do not need the information. Maybe we can just repeat that one line below, in only the places where we actually need `pr`?

lib/ci-helper.ts
```diff
@@ -511,19 +511,24 @@ export class CIHelper {
...
+
+            if (!pr || !pr.author) {
+                throw new Error(`Could not determine basic PR info for ${
+                    pullRequestURL}`);
+            }
+
```
dscho: As this is unlikely to happen, I don't know whether it is really necessary to add that error message. It's not like incorrect usage of GitGitGadget can cause this...

Since you already have the error message, though, I wonder whether we shouldn't just go with something like this:

```typescript
const getPRAuthor = () => {
    const pr = await this.github.getPRInfo(comment.prNumber);
    if (!pr || !pr.author) {
        throw new Error(`Could not determine basic PR info for ${pullRequestURL}`);
    }
    return pr.author;
}
```

and then use `await getPRAuthor()` in the places below?

lib/ci-helper.ts
```diff
@@ -554,6 +559,17 @@ export class CIHelper {
                 await addComment(`Submitted as [${
                     coverMid}](https://public-inbox.org/git/${coverMid})`);
             } else if (command === "/allow") {
+                if (!argument) {
```
dscho: It might make sense, for readability, to use something like `const accountName = argument || await getPRAuthor();`, and leave `argument` as a constant.

lib/ci-helper.ts
```diff
@@ -554,6 +559,17 @@ export class CIHelper {
...
+                    argument = pr.author;
+                } else {
+                    try {
+                        await this.github.getGitHubUserName(argument);
+                    } catch (reason) {
+                        throw new Error(`User ${
+                            argument} is not a valid GitHub username.`);
+                    }
+                }
```
dscho: Good idea! For simplicity, we could have this check even if we just obtained the account name via `await getPRAuthor()`.

lib/ci-helper.ts
```diff
@@ -562,6 +578,10 @@ export class CIHelper {
                         argument} already allowed to use GitGitGadget.`);
                 }
             } else if (command === "/disallow") {
+                if (!argument) {
+                    argument = pr.author;
+                }
+
```
dscho: Again, for readability, I would prefer `const accountName = argument || await getPRAuthor();`.

What do you think?

lib/ci-helper.ts
webstech: Sorry about that.  Forgot to retest with no trailing blanks after playing with third parameter support.

lib/ci-helper.ts
webstech: > As this is unlikely to happen, I don't know whether it is really necessary to add that error message. It's not like incorrect usage of GitGitGadget can cause this...

I will remove the check and error message.  I had just copied what was done for /submit (although pr was checked after pr.author was already used).  If there is a problem retrieving the PR info, an error will be thrown.

lib/ci-helper.ts
webstech: Okay.

lib/ci-helper.ts
webstech: Done.

lib/ci-helper.ts
dscho: No worries!

lib/ci-helper.ts
dscho: Okay.

lib/ci-helper.ts
dscho: Excellent!

PR #102, opened 2019-07-05
getPRInfo change deprecated number to pull_number

webstech:
Change to get rid of the deprecated warning message in the log.

### DIFF
--- a/lib/github-glue.ts
+++ b/lib/github-glue.ts
@@ -222,8 +222,8 @@ export class GitHubGlue {
      */
     public async getPRInfo(prNumber: number): Promise<IPullRequestInfo> {
         const response = await this.client.pulls.get({
-            number: prNumber,
             owner: "gitgitgadget",
+            pull_number: prNumber,
             repo: "git",
         });
         return {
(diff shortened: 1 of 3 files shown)


webstech:
It appears this requires an 'npm update'.  I don't know when the name change was done (not going to track it down).  I did not catch what version of octokit/rest I had.  After doing an update it bumped to 16.28.2.  Is there a process for getting an update done in the test/prod environment (to pick up fixes)?

@dscho How do you want to handle this?

dscho:
> @dscho How do you want to handle this?

I *think* you have to run `npm update octokit/rest` and then commit the `package-lock.json` file, but I am not really solid on npm usage...

webstech:
Updated @octokit/rest in package.json to 16.25.0.  V16.24.1 deprecated 'number' from several requests.  This is messy in that 16.25.0 must be explicitly installed to get the package-lock.json to show it instead of a later release.  Unfortunately some other packages that got updated do reflect a later release than actually required by the dependencies of @octokit/rest.

Original problem was my (cough cough) npm was (cough cough) out of date.  Can get node (and yarn) updates through chocolatey but npm requires manual updates.

dscho:
Thank you so much!

PR #104, opened 2019-07-11
Upgrade Typescript to latest version

dscho:
The latest octokit type definitions apparently are incompatible with the Typescript version we used before. Not sure why the CI build did not catch this...

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5648,9 +5648,9 @@
       }
     },
     "typescript": {
-      "version": "2.9.2",
-      "resolved": "https://registry.npmjs.org/typescript/-/typescript-2.9.2.tgz",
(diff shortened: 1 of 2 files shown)


dscho:
I'll just override the requirement to have one approving review, as this one really blocks pretty much everything in GitGitGadget.

webstech:
@dscho Really sorry about this.  A new form of dll-hell.  I had reinstalled all the packages after updating npm hoping to set things straight.  Removed a global typescript package but still do not get any errors.  No need to research - I will really try to avoid issues like this in the future.

dscho:
@webstech no worries, I could not repeat in my local tests, and our CI build also did not catch it. Probably has something to do with the node.js version I use in the build agent that actually runs GitGitGadget...

Issue #58, opened 2018-12-23
Improve `Cc:`ing


dscho:
We should *always* Cc: the Submitter. To that end, we need to record email addresses of the people allowed to use GitGitGadget (i.e. extend `/allow` to try to figure out a public email on GitHub, and if there is none, require one to be provided in the same `/allow` command, extending the syntax to `/allow Name Of Submitter [<email@address>]`).

While at it, `/allow` could also get the feature where calling it without argument automatically uses the owner of the PR as the user to add.

webstech:
Pull request #101 will default /allow to the owner of the PR.

I am not sure GitGitGadget should retain the email address.  It could just get it from github each time.  GitHubGlue::getGitHubUserInfo(login: string): Promise<IGHUser> (see issue #94 for interface) could replace getGitHubUserName use in CIHelper.  That would support the user changing/adding a public email address in github.

If someone does not have a public email address, they probably do not want to provide a private one to GitGitGadget and have it publicly available.  Considering how easy it is to have a throw away email address I wonder how many users do not have a public email address.  What GitGitGadget could do when the PR is opened is add a comment if the PR author does not have a public email.  The author can then make the choice to add one or not.  They can remove it from their profile after the PR is closed if they want.  That would still leave the email public in the repo but not as obvious as looking at a user profile.

dscho:
> I am not sure GitGitGadget should retain the email address. It could just get it from github each time.

My experience is that it is not public most of the time, and even the real name is not part of the public information. So I really would like to have a `.mailmap`-like record of mappings between the GitHub account name and the "Real Name <email@address>" information.

> If someone does not have a public email address, they probably do not want to provide a private one to GitGitGadget and have it publicly available.

True. I know of one such case. The problem with that is that Git does not accept such contributions. The real name and a working email address is required for the DCO to be remotely valid.

webstech:
> I really would like to have a `.mailmap`-like record of mappings between the GitHub account name and the "Real Name [email@address](mailto:email@address)" information

Having a separate copy means providing a method to update (yes, I know, disallow/allow but someone has to remember to do it).  On the other hand, if GitGitGadget just refused to submit requests without a public email in github, the github info can be used.  It would make it simpler to allow the user to change the email (and name) by using a single source.  Is the github public email always an address verified by github?

What do you think?

dscho:
Good points.

[...]

PR #110, opened 2019-08-05
Copy submitter to address issue 58

webstech:
This change is done in three commits.  It is intended to address issue #58 where the person submitting changes is not the author (this may be a simplification of the problem being addressed).

The first adds a warning comment in the /allow if a user does not have a public email address.  

The second does the same during a /submit.  It also changes the signature for gitgitgadget::submit to only pass the pull request and user info.  Passing the pull request object removes the need for the CIHelper to know what gitgitgadget needs from the pull request.

The third commit simply adds the submitter to the cc list.

### DIFF
--- a/lib/gitgitgadget.ts
+++ b/lib/gitgitgadget.ts
@@ -1,6 +1,7 @@
 import { isDirectory } from "./fs-util";
 import { git, gitConfig } from "./git";
 import { GitNotes } from "./git-notes";
+import { IGitHubUser, IPullRequestInfo } from "./github-glue";
 import { PatchSeries } from "./patch-series";
 import { IPatchSeriesMetadata } from "./patch-series-metadata";
 import { ISMTPOptions, parseHeadersAndSendMail } from "./send-mail";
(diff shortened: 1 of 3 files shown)


webstech:
Most of the testing was done using the changes in the ci-helper-test branch.  This is a basic setup to test the changes for this PR and the previous user verification PR.  Is there any interest in improving and formalizing these changes?  Just trying to get something in place for future testers.  The changes in ci-helper-test branch are  by no means complete.  Should an issue be opened to get more input?

fyi, I do not know how to use jest to mock the functions being overridden.

webstech:
Please do not review until I make an update.  Need to authenticate with github to obtain the user email address.  Initial tests had been done after authentication was already done.

webstech:
Okay, initial tests were not done after authentication because I do not have the key.  github-glue.ts has been updated in the first commit to authenticate.  Can't really test the code without getting a key.  The function is mocked out in testing.

Ready for review again.

lib/gitgitgadget.ts
```diff
@@ -163,50 +164,43 @@ export class GitGitGadget {
...
         return coverMid;
     }
 
     protected async updateNotesAndPullRef(pullRequestNumber: number,
                                           additionalRef?: string):
         Promise<string> {
-        if (!await isDirectory(this.workDir)) {
-            await git(["init", "--bare", this.workDir]);
-        }
-
```
dscho: Is this removal intentional?

webstech: The init is done in the get method at line 44.  I assume that was added at some point without removing the later one.  I can restore the init if I misunderstood.

dscho: But that `get()` method is static, while `updateNotesAndPullRef()` is not. Therefore, the latter could be called without having had the benefit of the former guaranteeing that the `workDir` exists...

I think we need to reinstate these lines.

webstech: The static get() method is used to construct a new instance of GitGitGadget.  This is currently the only way to access the protected constructor so the check would have been done?

Would it be better to have the constructor do the git init and call fetchAndReReadOptions()?  This would remove the git init and fetch and readOptions() from the get() method.  It would ensure the git init has been done (and options are current) if there is a new get() alternative added to create a GitGitGadget.  This would be a separate pull request of course.

dscho: Ah, you're right, of course, the `get()` method is kind of the constructor: we need this to be `async`, and a constructor cannot be `async`.

So your patch is of course fine as-is, as you patiently explained to me often enough for me to understand. Thanks!

PR #122, opened 2019-09-30
Accommodate Git GUI PRs

dscho:
The main aim of this PR is to fix the automated task that tries to match PRs' commits to commits in `upstream/pu` (which fails with an out-of-memory for Git GUI PRs).


dscho:
@webstech FYI I already use this branch on the agent that runs the automated jobs, but any review would be appreciated... I'd rather run reviewed `master` than a topic branch with a hot fix.

dscho:
Thank you!

Issue #83, opened 2019-04-05
Implement /test to do dry-run or send to alternate email address


nasamuffin:
As discussed in IRC today:

```
17:08:04    nasamuffin_ | I'm trying to figure out whether it makes sense to target the tutorial to using dscho's gitgitgadget & was curious whether there's a way to have it send 
                        | the mails to alternate email                                                                                                                             
17:08:41       jrnieder | nasamuffin_: interesting. Even outside the tutorial context, sounds useful as a sort of "dry run"                                                        
17:08:50        __rohit | nasamuffin_: Yes, a tutorial on gitgitgadget will benefit the beginners                                                                                  
17:09:22          dscho | jrnieder exciting news about trace2! Do you think you will be able to share some of your findings in the future?                                         
17:09:40          dscho | Welcome, nasamuffin_!                                                                                                                                    
17:09:44        __rohit | although current "documentation" of gitgit is good enough                                                                                                
17:10:07          dscho | Regarding alternate email: GitGitGadget is all open source ;-)                                                                                           
17:10:13       jrnieder | dscho: yep! I was happy to see the stats jobriggs described in Git Merge, would love to see more people report that kind of thing more regularly         
17:10:16    nasamuffin_ | it seems straightforward to use, but i don't want newbies spamming the list with a toy patch                                                             
17:10:18         rafasc | On the topic of helping beginners, I have recently discovered https://git-send-email.io/. (I am not related to the project in any shape or form)         
17:10:25    nasamuffin_ | dscho: i'll learn typescript and figure it out :P                                                                                                        
17:10:49          dscho | __rohit: please do open PRs to improve the documentation as you see fit.                                                                                 
17:11:11          dscho | nasamuffin_ I often hang out in this here channel, and try to do my best to answer questions ;-)                                                         
17:11:20  jonathantanmy | For me, more protocol v2 bug fixes and a small partial clone improvement (batch blobs fetched for diff/show). Also worked on some CDN offloading patches 
                        | for JGit, but I don't know when that will see production.                                                                                                
17:12:20          dscho | rafasc is that site really just accumulating links to real documentation?                                                                                
17:12:41        __rohit | dscho: I will.                                                                                                                                           
17:12:49       jrnieder | jonathantanmy: you're burying the lede. ;-) Interested reviewers can look at the series ending at https://git.eclipse.org/r/c/139993                     
[...]

PR #125, opened 2019-10-03
Add /preview PR comment

webstech:
This PR addresses issue #83.  An update to gitgitgadget/gitgitgadget.github.io will be done if/when approved.

The /preview comment will only send the emails to the commentor.

The email subject prefix will be PREVIEW instead of PATCH to
identify the email.

A new flag PatchSeriesOptions::noUpdate replaces some of the function
of the dryRun flag to prevent updates to the repos.

### DIFF
--- a/lib/ci-helper.ts
+++ b/lib/ci-helper.ts
@@ -540,6 +540,24 @@ export class CIHelper {
                 await addComment(`Submitted as [${
                     coverMid}](https://public-inbox.org/git/${coverMid})${
                         extraComment}`);
+            } else if (command === "/preview") {
+                if (argument && argument !== "") {
+                    throw new Error(`/preview does not accept arguments ('${
+                        argument}')`);
+                }
+
+                const pr = await this.getPRInfo(comment.prNumber,
+                                                pullRequestURL);
+                const userInfo = await this.getUserInfo(comment.author);
+
+                if (!userInfo.email) {
+                    throw new Error(`Could not determine public email of ${
+                        comment.author}`);
+                }
+
+                const coverMid =
+                    await gitGitGadget.preview(pr, userInfo);
+                await addComment(`Preview email sent as ${coverMid}`);
             } else if (command === "/allow") {
                 const accountName = argument || await getPRAuthor();
                 let extraComment = "";
--- a/lib/ci-helper.ts
+++ b/lib/ci-helper.ts
@@ -3,7 +3,7 @@ import * as util from "util";
 import { commitExists, git } from "./git";
 import { GitNotes } from "./git-notes";
 import { GitGitGadget, IGitGitGadgetOptions } from "./gitgitgadget";
-import { GitHubGlue } from "./github-glue";
+import { GitHubGlue, IGitHubUser, IPullRequestInfo } from "./github-glue";
 import { MailCommitMapping } from "./mail-commit-mapping";
 import { IMailMetadata } from "./mail-metadata";
 import { IPatchSeriesMetadata } from "./patch-series-metadata";
@@ -523,33 +523,14 @@ export class CIHelper {
                         argument}')`);
                 }
 
-                const pr = await this.github.getPRInfo(comment.prNumber);
-
+                const pr = await this.getPRInfo(comment.prNumber,
+                                                pullRequestURL);
                 if (pr.author !== comment.author) {
                     throw new Error("Only the owner of a PR can submit it!");
                 }
 
-                if (!pr.baseLabel || !pr.baseCommit ||
-                    !pr.headLabel || !pr.headCommit) {
-                    throw new Error(`Could not determine PR details for ${
-                        pullRequestURL}`);
-                }
+                const userInfo = await this.getUserInfo(comment.author);
 
-                if (!pr.title || !pr.body) {
-                    throw new Error("Ignoring PR with empty title and/or body");
-                }
-
-                if (!pr.mergeable) {
-                    throw new Error("Refusing to submit a patch series "
-                        + "that does not merge cleanly.");
-                }
-
-                const userInfo =
-                    await this.github.getGitHubUserInfo(comment.author);
-                if (!userInfo.name) {
-                    throw new Error(`Could not determine full name of ${
-                        comment.author}`);
-                }
                 const extraComment = userInfo.email === null ?
                     ( `\n\nWARNING: ${comment.author} has no public email` +
                     " address set on GitHub" ) : "";
@@ -629,6 +610,37 @@ export class CIHelper {
             await PublicInboxGitHelper.get(this.notes, publicInboxGitDir,
                                            this.github);
         return await publicInboxGit.processMails(prFilter);
+    }
+
+    private async getPRInfo(prNumber: number, pullRequestURL: string):
+        Promise<IPullRequestInfo> {
+        const pr = await this.github.getPRInfo(prNumber);
+
+        if (!pr.baseLabel || !pr.baseCommit ||
+            !pr.headLabel || !pr.headCommit) {
+            throw new Error(`Could not determine PR details for ${
+                pullRequestURL}`);
+        }
+
+        if (!pr.title || !pr.body) {
+            throw new Error("Ignoring PR with empty title and/or body");
+        }
+
+        if (!pr.mergeable) {
+            throw new Error("Refusing to submit a patch series "
+                + "that does not merge cleanly.");
+        }
+
+        return pr;
+    }
+
+    private async getUserInfo(author: string): Promise<IGitHubUser> {
+        const userInfo = await this.github.getGitHubUserInfo(author);
+        if (!userInfo.name) {
+            throw new Error(`Could not determine full name of ${author}`);
+        }
+
+        return userInfo;
     }
 
     private async maybeUpdateGGGNotes(): Promise<void> {
--- a/lib/gitgitgadget.ts
+++ b/lib/gitgitgadget.ts
@@ -5,7 +5,9 @@ import { IGitHubUser, IPullRequestInfo } from "./github-glue";
 import { PatchSeries, SendFunction } from "./patch-series";
 import { IPatchSeriesMetadata } from "./patch-series-metadata";
 import { PatchSeriesOptions } from "./patch-series-options";
-import { ISMTPOptions, parseHeadersAndSendMail } from "./send-mail";
+import {
+    ISMTPOptions, parseHeadersAndSendMail, parseMBox,
+    sendMail } from "./send-mail";
 
 export interface IGitGitGadgetOptions {
     allowedUsers: string[];
@@ -165,6 +167,21 @@ export class GitGitGadget {
         return true;
     }
 
+    // Send emails only to the user
+    public async preview(pr: IPullRequestInfo, userInfo: IGitHubUser):
+        Promise<string | undefined> {
+
+        const send = async (mail: string): Promise<string> => {
+            const mbox = await parseMBox(mail);
+            mbox.cc = [];
+            mbox.to = userInfo.email;
+            console.log(mbox);
+            return await sendMail(mbox, this.smtpOptions);
+        };
+
+        return await this.genAndSend( pr, userInfo, {noUpdate: true}, send);
+    }
+
     // Send emails out for review
     public async submit(pr: IPullRequestInfo, userInfo: IGitHubUser):
         Promise<string | undefined> {
@@ -221,7 +238,7 @@ export class GitGitGadget {
             await GitGitGadget.readOptions(this.notes);
     }
 
-    // Finish the job for submit
+    // Finish the job for preview and submit
     protected async genAndSend(pr: IPullRequestInfo, userInfo: IGitHubUser,
                                options: PatchSeriesOptions,
                                send: SendFunction):
--- a/lib/gitgitgadget.ts
+++ b/lib/gitgitgadget.ts
@@ -2,8 +2,9 @@ import { isDirectory } from "./fs-util";
 import { git, gitConfig } from "./git";
 import { GitNotes } from "./git-notes";
 import { IGitHubUser, IPullRequestInfo } from "./github-glue";
-import { PatchSeries } from "./patch-series";
+import { PatchSeries, SendFunction } from "./patch-series";
 import { IPatchSeriesMetadata } from "./patch-series-metadata";
+import { PatchSeriesOptions } from "./patch-series-options";
 import { ISMTPOptions, parseHeadersAndSendMail } from "./send-mail";
 
 export interface IGitGitGadgetOptions {
@@ -164,41 +165,15 @@ export class GitGitGadget {
         return true;
     }
 
-    // better parms would be to pass prinfo object
+    // Send emails out for review
     public async submit(pr: IPullRequestInfo, userInfo: IGitHubUser):
         Promise<string | undefined> {
-
-        // if known, add submitter to email chain
-        const ccSubmitter = userInfo.email ? `\nCc: ${
-            userInfo.name} <${userInfo.email}>` : "";
-        const description = `${pr.title}\n\n${pr.body}${ccSubmitter}`;
-
-        if (pr.baseOwner !== "gitgitgadget" || pr.baseRepo !== "git") {
-            throw new Error(`Unsupported repository: ${pr.pullRequestURL}`);
-        }
-
-        // get metadata in work repo
-        const metadata =
-            await this.notes.get<IPatchSeriesMetadata>(pr.pullRequestURL);
-        const previousTag = metadata && metadata.latestTag ?
-            `refs/tags/${metadata.latestTag}` : undefined;
-        // update work repo from base
-        await this.updateNotesAndPullRef(pr.number, previousTag);
-
-        const series =
-            await PatchSeries.getFromNotes(this.notes, pr.pullRequestURL,
-                                           description, pr.baseLabel,
-                                           pr.baseCommit, pr.headLabel,
-                                           pr.headCommit, userInfo.name);
 
         const send = async (mail: string): Promise<string> => {
             return await parseHeadersAndSendMail(mail, this.smtpOptions);
         };
-        const coverMid =
-            await series.generateAndSend(console, send,
-                                         this.publishTagsAndNotesToRemote,
-                                         pr.pullRequestURL, new Date());
-        return coverMid;
+
+        return await this.genAndSend(pr, userInfo, {}, send);
     }
 
     protected async updateNotesAndPullRef(pullRequestNumber: number,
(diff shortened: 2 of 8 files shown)


lib/gitgitgadget.ts
```diff
@@ -164,41 +169,29 @@ export class GitGitGadget {
...
+        return await this.genAndSend( pr, userInfo, {noUpdate: true}, send);
+    }
 
-        const series =
-            await PatchSeries.getFromNotes(this.notes, pr.pullRequestURL,
-                                           description, pr.baseLabel,
-                                           pr.baseCommit, pr.headLabel,
-                                           pr.headCommit, userInfo.name);
+    // Send emails out for review
+    public async submit(pr: IPullRequestInfo, userInfo: IGitHubUser):
```
dscho: I feel that it would make the patch more obvious if the refactoring of `submit()`'s code was done in a preparatory commit, and the `preview()` function was introduced only in a subsequent commit. It is a bit hard for me to be certain that I did not miss anything when reviewing such a large patch...

lib/patch-series.ts
```diff
@@ -140,10 +140,12 @@ export class PatchSeries {
                 pullRequestURL,
             };
         } else {
-            if (!await git(["rev-list",
-                            `${metadata.headCommit}...${headCommit}`],
-                           { workDir })) {
-                throw new Error(`${headCommit} was already submitted`);
+            if (!options.noUpdate) {    // allow reprint of submitted PRs
```
dscho: Good.

Two suggestions:

1. introduce the `noUpdate` field in its own, preparatory commit. That structures the commit history better, and makes the changes easier to review.
2. extend the condition in the already-existing `if`: 

   ```typescript
   if (!options.noUpdate &&
       !await git(["rev-list",
   [...]
   ```

lib/patch-series.ts
```diff
@@ -778,7 +777,7 @@ export class PatchSeries {
         }
 
         logger.log("Publishing branch and tag");
-        if (this.project.publishToRemote && !this.options.dryRun) {
+        if (this.project.publishToRemote && !this.options.noUpdate) {
```
dscho: I would like to keep the `--dry-run` option working, even if I have no plans on using the command-line version of GitGitGadget, ever. Therefore, I would have extended the condition, like so:

```typescript
         if (this.project.publishToRemote && !this.options.dryRun && !this.options.noUpdate) {
```

I do see, though, that you changed the command-line script to not only set `dryRun` upon the `--dry-run` option, but also `noUpdate`.

If this was done as a preparatory commit that introduces the `noUpdate` field, I would be totally on board with this, as it would make it a lot easier to review than one large patch...

lib/patch-series.ts
```diff
@@ -858,14 +857,10 @@ export class PatchSeries {
     }
 
     protected async publishBranch(tagName: string): Promise<void> {
-        if (!this.project.publishToRemote || this.options.dryRun) {
-            return;
-        }
-
```
dscho: In the command-line version, back when it was still a Unix shell script, and then a Javascript script, I used to use `--dry-run` a lot, and I really did not want anything to be published in that case.

So I would like to keep this here.

If all of the calls already have this guard, I would rather unguard the calls and keep the guard in the function, as that makes it much safer to use it in the future.

lib/patch-series.ts
```diff
@@ -858,14 +857,10 @@ export class PatchSeries {
...
         if (this.options.redo) {
             tagName = "+" + tagName;
         }
-        await git(["push", this.project.publishToRemote,
+        await git(["push", `${this.project.publishToRemote}`,
```
dscho: Is this really intentional?

res/WELCOME.md
```diff
@@ -21,6 +21,10 @@ Once on the list of permitted usernames, you can contribute the patches to the G
 
 After you submit, GitGitGadget will respond with another comment that contains the link to the cover letter mail in the Git mailing list archive. Please make sure to monitor the discussion in that thread and to address comments and suggestions.
 
+If you want to see what email(s) would be sent for a submit request, add
+a PR comment '/preview' to have the email(s) sent to you.  You must have
+a public GitHub email address for this.
+
```
dscho: In the `.md` file, please do not wrap the lines of a paragraph. I don't know for certain, but I imagine that GitHub's rendering would retain the wrapped lines, which I imagine would look less elegant.

lib/ci-helper.ts
```diff
@@ -559,6 +572,24 @@ export class CIHelper {
...
+
+                if (!userInfo.email) {
+                    throw new Error(`Could not determine public email of ${
+                        comment.author}`);
+                }
+
+                const coverMid =
+                    await gitGitGadget.preview(pr, userInfo);
+                await addComment(`Preview email sent as [${
+                    coverMid}](https://public-inbox.org/git/${coverMid})`);
```
dscho: Oh, completely forgot: do we want to provide a link here? Public-inbox won't have this mail, like, ever, because it was sent privately...

And now that I think of it, we should probably pay attention to _not_ Cc:ing the original authors during `/preview`. That could potentially be very irritating. I could imagine that `insertCcAndFromLines()` would need to be patched, but I might well be wrong.

lib/ci-helper.ts
```diff
@@ -505,6 +505,39 @@ export class CIHelper {
...
+            Promise<IGitHubUser> => {
+            const userInfo = await this.github.getGitHubUserInfo(author);
+            if (!userInfo.name) {
+                throw new Error(`Could not determine full name of ${
+                    author}`);
+            }
+
+            return userInfo;
+        };
+
```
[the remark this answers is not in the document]
webstech: @dscho Just to clarify before I continue to comment, do you want the PR broken into multiple commits or additional PRs for the commits?  Thanks for the quick review.

dscho:
That split looks very, very nice! Thank you!

lib/gitgitgadget.ts
```diff
@@ -164,41 +169,29 @@ export class GitGitGadget {
...
 
-        // get metadata in work repo
-        const metadata =
-            await this.notes.get<IPatchSeriesMetadata>(pr.pullRequestURL);
-        const previousTag = metadata && metadata.latestTag ?
-            `refs/tags/${metadata.latestTag}` : undefined;
-        // update work repo from base
-        await this.updateNotesAndPullRef(pr.number, previousTag);
+        return await this.genAndSend( pr, userInfo, {noUpdate: true}, send);
+    }
```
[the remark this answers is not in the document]
dscho: In other words, I think that we will need to pass through e.g. a `coverMidSuffix` (that I would initialize to some known unique value for `prefix()`, e.g. `${Date.now()}`) and append it to the Message-ID [here](https://github.com/gitgitgadget/gitgitgadget/blob/bdcebdffd98610b6d9e295cf160b189d4e1d7746/lib/patch-series.ts#L637-L638).

lib/ci-helper.ts
dscho: >  we should probably pay attention to _not_ Cc:ing the original authors during `/preview`.

I think we'll just need to guard [these lines](https://github.com/gitgitgadget/gitgitgadget/blob/bdcebdffd98610b6d9e295cf160b189d4e1d7746/lib/patch-series.ts#L600-L603) behind a new parameter `noCc: boolean` or something like it.

lib/patch-series.ts
```diff
@@ -141,9 +142,10 @@ export class PatchSeries {
                 pullRequestURL,
             };
         } else {
-            if (!await git(["rev-list",
+            if (!options.noUpdate &&   // allow reprint of submitted PRs
+                !await git(["rev-list",
                             `${metadata.headCommit}...${headCommit}`],
-                           { workDir })) {
+                            { workDir })) {
```
dscho: I think it is this extra space that [the build is complaining about](https://dev.azure.com/gitgitgadget/git/_build/results?buildId=17643&view=logs&jobId=12f1170f-54f2-53f3-20dd-22fc7dff55f9&taskId=9c939e41-62c2-5605-5e05-fc3554afc9f5&lineStart=36&lineEnd=36&colStart=1&colEnd=83).

webstech: Sorry, did not mean for you to have to look that up.  Forgot the push would kick off the CI test and notify you.  Fixed it.

lib/patch-series.ts
webstech: I had switched to the template literal because the linters decided publishToRemote could be null.  Reinstating the guard (as per your request) made the linters happy and the change has been removed.

res/WELCOME.md
webstech: Done and corrected how /preview is displayed.

lib/patch-series.ts
dscho: Just let me know when I should look at this PR again 😄

lib/ci-helper.ts
webstech: 1. Changed the comment to not be a markdown link.  It especially won't be valid if the message id is changed.
2. This is a preview so shouldn't all the names show up?  The provided send function in GitGitGadget::preview() removes the cc and changes the to for the email, so the envelope only has one to: address.

dscho: >     1. Changed the comment to not be a markdown link.  It especially won't be valid if the message id is changed.

:+1:



>     2\. This is a preview so shouldn't all the names show up?  The provided send function in GitGitGadget::preview() removes the cc and changes the to for the email, so the envelope only has one to: address.

Oh, I missed that! That's a much more elegant solution than the one I proposed, I agree.

lib/gitgitgadget.ts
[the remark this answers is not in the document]
webstech: @dscho Git generates the emails with a unique value in the message-id.  What do you thing about retaining that in the new cover letter message id?  It is also part of the individual fix message-ids.  This would remove the need for a new option.  This would ensure multiple /preview comments get different message-id values.
ie
_Git generates_:
Message-Id: <cover.1570600244.git.user@example.com>
In-Reply-To: <pull.59.git.1570600233.user@example.com>
_gitgitgadget changes it to:_
Message-ID: <pull.59.v2.git.1570600244.user@example.com>
In-Reply-To: <pull.59.git.1570600233.user@example.com>

I understand the reply comment for /submit and /preview will not be as clean as it currently is but is that a problem?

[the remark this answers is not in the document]
webstech: Instance extraction at ~627:
            const coverInst = coverMid.match(/cover(\.[0-9]+\.?)/);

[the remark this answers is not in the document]
dscho: > What do you thing about retaining that in the new cover letter message id?

Hmm. I rather like that the Message-IDs talk about GitGitGadget (or at least about the originating PullRequest).

> _gitgitgadget changes it to:_
> Message-ID: [pull.59.v2.git.1570600244.user@example.com](mailto:pull.59.v2.git.1570600244.user@example.com)
> In-Reply-To: [pull.59.git.1570600233.user@example.com](mailto:pull.59.git.1570600233.user@example.com)

Actually, it does not. It changes it to `pull.59.git.gitgitgadget@gmail.com`, see e.g. https://public-inbox.org/git/pull.380.git.gitgitgadget@gmail.com/

How about changing the Message-ID to include the Unix epoch always (or even easier: the milliseconds since Jan 1st 1970, which is what `Date.now()` returns)? I.e. something like `pull.59.git.1570609920817.gitgitgadget@gmail.com`?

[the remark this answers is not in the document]
webstech: > Hmm. I rather like that the Message-IDs talk about GitGitGadget (or at least about the originating PullRequest).

Agree.  I was just clumsily proposing to retain the milliseconds value provided by git.  That would keep it consistent with the rest of the patch emails.

> Actually, it does not. It changes it to `pull.59.git.gitgitgadget@gmail.com`, see e.g. https://public-inbox.org/git/pull.380.git.gitgitgadget@gmail.com/

Sorry, the example was with the proposed change and the unit tests require user@example.com.

> How about changing the Message-ID to include the Unix epoch always (or even easier: the milliseconds since Jan 1st 1970, which is what `Date.now()` returns)? I.e. something like `pull.59.git.1570609920817.gitgitgadget@gmail.com`?

Sounds like we agree on this.  I will use the timestamp from the git format-patch output.
Cover letter (user@example.com would be gitgitgadget on azure):
Message-Id: <pull.59.v2.git.1570638881.user@example.com
Patch email:
Message-Id: <f886cb837afaf10c720beae3ca88e740d93b3aea.1570638881.git.user@example.com

webstech:
@dscho Ready for review.  Tests updated to support message-id change.  Tests have extra checks to accommodate tsconfig errors.

lib/gitgitgadget.ts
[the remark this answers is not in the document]
dscho: > the example was with the proposed change

Sorry for being so thick! Yes, I agree with that change ;-)

lib/patch-series.ts
```diff
@@ -624,6 +623,8 @@ export class PatchSeries {
                 throw new Error(`Could not extract cover letter Message-ID`);
             }
 
+            const midTimeStamp = coverMid.match(/cover(\.[0-9]+\.?)/);
+            const midTS = midTimeStamp ? midTimeStamp[1] : ".";
```
dscho: Those are two `const` variables with very similar names, i.e. the same purpose, and they include the dots (but maybe not a trailing one?).

How about this instead?

```typescript
let timeStamp = coverMid.match(/cover\.([0-9]+\)/);
timeStamp = timeStamp ? timeStamp[1] : `${Date.now()}`;
```

Later on, you would use this as `.${timeStamp}.`.

tests/gitgitgadget.test.ts
```diff
@@ -246,14 +246,26 @@ to have included in git.git [https://github.com/git/git].`);
...
+    const seriesMeta = await notes.get<IPatchSeriesMetadata>(pullRequestURL);
+    if (seriesMeta && seriesMeta.coverLetterMessageId) {
+        seriesMeta.coverLetterMessageId = seriesMeta.coverLetterMessageId
+            .replace(midCleanupRegex, midCleanup);
+    }
+    if (seriesMeta && seriesMeta.referencesMessageIds) {
+        seriesMeta.referencesMessageIds[0] = seriesMeta.referencesMessageIds[0]
+            .replace(midCleanupRegex, midCleanup);
+    }
+    expect(seriesMeta).toEqual({
```
dscho: One part of this test tries to verify that the Message-ID of the cover letter is properly referenced in the replies. This patch weakens this a bit (it would fail to catch a regression where the `References` misses the timestamp, for example).

But it should be relatively easy to extract the cover letter's Message-ID, and for added brownie points verify that it matches the new format.

Something along these lines:

```typescript
const seriesMeta = await notes.get<IPatchSeriesMetadata>(pullRequestURL);

expect(seriesMeta.coverLetterMessageId).not.toBeNull();
expect(seriesMeta.coverLetterMessageId).toMatch(/pull\.1\.v2\.git\.\d+\.gitgitgadget@example\.com/);

expect(seriesMeta).toEqual({
    [...]
    coverLetterMessageId: seriesMeta.coverLetterMessageId,
    [...]
```

That would build a bit more confidence in me ;-)

dscho: Likewise, you would need the same trick later on:

```typescript
[...]
Submitted-As: https://dummy.com/?mid=${seriesMeta.coverLetterMessageId}
In-Reply-To: https://dummy.com/?mid=${seriesMeta.coverLetterMessageId}`);
```

dscho:
> Ready for review. Tests updated to support message-id change. Tests have extra checks to accommodate tsconfig errors.

Very, very nice!

I added two suggestions how I think the PR could be improved, still. This is getting really close. Exciting!

lib/patch-series.ts
webstech: timeStamp in the first assignment is an array and the second use is as a string.  Typescript throws an error - I will keep two vars for the moment (unless you have a better suggestion).

The original code with extra variable midTS was to overcome typescript rejecting midTimeStamp[1] as possibly not defined.  That was kind of the idea - get a runtime (not compile time) error if the regex for midTimeStamp failed.  The match will not fail (the value will always be there) but we have to write guard code for it (assuming git does not change).  Sorry for letting some coding philosophy slip in there.

tests/gitgitgadget.test.ts
webstech: Changes made and pushed.  May do a later PR to use the baseCommit var in the cat-file test to make it more flexible to changes.

lib/patch-series.ts
```diff
@@ -624,6 +623,8 @@ export class PatchSeries {
                 throw new Error(`Could not extract cover letter Message-ID`);
             }
 
+            const timeStamp = coverMid.match(/cover\.([0-9]+)\./);
+            const midTStamp = timeStamp ? timeStamp[1] : `${Date.now()}`;
```
dscho: Hmm. I really had hoped to avoid two variables here, and certainly I see still the possible confusion between them given that they have very, very similar names. I guess I could warm up to calling the first one `match` and the second one `timeStamp` instead.

tests/gitgitgadget.test.ts
```diff
@@ -246,24 +246,31 @@ to have included in git.git [https://github.com/git/git].`);
...
+        .toMatch(/pull\.1\.v2\.git\.\d+\.gitgitgadget@example\.com/);
     expect(mails.length).toEqual(5);
     if (await gitCommandExists("range-diff", repo.workDir)) {
         expect(mails[0]).toMatch(/Range-diff vs v1:\n[^]*\n -: .* 4: /);
     }
     expect(await repo.revParse("pr-1/somebody/master-v2")).toBeDefined();
 
-    expect(await notes.get(pullRequestURL)).toEqual({
+    const seriesMeta = await notes.get<IPatchSeriesMetadata>(pullRequestURL);
+    expect(seriesMeta.coverLetterMessageId).not.toBeNull();
```
dscho: I fear that my advice was not good... The Typescript compiler [complains](https://dev.azure.com/gitgitgadget/git/_build/results?buildId=18091&view=logs&jobId=12f1170f-54f2-53f3-20dd-22fc7dff55f9&taskId=9c939e41-62c2-5605-5e05-fc3554afc9f5&lineStart=39&lineEnd=45&colStart=1&colEnd=1). Maybe it works if you use `.not.toBeUndefined()`?

I fear that you still have to append `!` to the subsequent uses of `seriesMeta.coverLetterMessageId`...

lib/patch-series.ts
webstech: As noted earlier, two vars are needed due to different types.  `match` was already used so `tsMatch` was chosen.

tests/gitgitgadget.test.ts
webstech: Due to the problem raised in issue #130 I did not catch the error before the CI build.  I changed to extract the values to new vars to get past the problem.  Will look at using !.

lib/patch-series.ts
dscho: > timeStamp in the first assignment is an array and the second use is as a string. Typescript throws an error - I will keep two vars for the moment (unless you have a better suggestion).

Ah. Yes, I would have the suggestion to specify the type explicitly: `let timeStamp: string[] | string | null` ;-)

But as I stated elsewhere, I would also be fine with this:

```typescript
const match = coverMid.match(/cover\.([0-9]+\)/);
const timeStamp = match ? match[1] : `${Date.now()}`;
```

lib/patch-series.ts
dscho: Ah, okay!

lib/patch-series.ts
dscho: Oh, I missed the update... Nevermind! ;-)

dscho:
Oh, I found another problem: the tag was still created (even if it was not pushed).

I will fix this, and while at it, amend the test to use the `!` trick, then merge.

dscho:
Oh, and I used this opportunity to verify that GMail really hates the "GMT" in the `Date:` header. See e.g. [this mail's raw headers](https://public-inbox.org/git/37aee862e14b1352eb08485f15ea06bab33679df.1570826543.git.gitgitgadget@gmail.com/raw):

```mbox
[...]
Date:   Fri, 11 Oct 2019 13:42:25 -0700 (PDT)
X-Google-Original-Date: Fri, 11 Oct 2019 20:42:22 GMT
[...]
```

Replacing the "GMT" by "+0000" seems to do the trick, so that's what I slipped into this PR, too. (We need to override the `Date:` header to enforce a consistent ordering of the mails in a thread that contains more than one patch.)

tests/gitgitgadget.test.ts
dscho: Done ;-)

I hope you did not continue working on that, if so: my apologies for wasting your time!

PR #126, opened 2019-10-04
Cc: the correct maintainer for Git GUI contributions

dscho:
I have no doubt that there are more corners to fix for Git GUI contributions, but this is probably the most obvious one.

### DIFF
--- a/lib/project-options.ts
+++ b/lib/project-options.ts
@@ -1,5 +1,5 @@
 import {
-    commitExists, git, gitConfig, gitConfigForEach,
+    commitExists, git, gitConfig, gitConfigForEach, revParse,
 } from "./git";
 
 // For now, only the Git, Cygwin and BusyBox projects are supported
@@ -39,7 +39,13 @@ export class ProjectOptions {
         let to: string;
         let midUrlPrefix: string = " Message-ID: ";
 
-        if (await commitExists("e83c5163316f89bfbde", workDir)) {
+        if (await commitExists("cb07fc2a29c86d1bc11", workDir) &&
+            await revParse(`${baseCommit}:git-gui.sh`, workDir) !== undefined) {
+            // Git GUI
+            to = "--to=git@vger.kernel.org";
+            cc.push("Pratyush Yadav <me@yadavpratyush.com>");
+            upstreamBranch = "git-gui/master";
+        } else if (await commitExists("e83c5163316f89bfbde", workDir)) {
             // Git
             to = "--to=git@vger.kernel.org";
             cc.push("Junio C Hamano <gitster@pobox.com>");
--- a/lib/gitgitgadget.ts
+++ b/lib/gitgitgadget.ts
@@ -205,12 +205,14 @@ export class GitGitGadget {
                                           additionalRef?: string):
         Promise<string> {
         const pullRequestRef = `refs/pull/${pullRequestNumber}/head`;
+        const pullRequestMerge = `refs/pull/${pullRequestNumber}/merge`;
         const args = [
             "fetch",
             this.publishTagsAndNotesToRemote,
             "--",
             `+${this.notes.notesRef}:${this.notes.notesRef}`,
             `+${pullRequestRef}:${pullRequestRef}`,
+            `+${pullRequestMerge}:${pullRequestMerge}`,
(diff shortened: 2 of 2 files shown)


dscho:
Hrm. I just changed the Azure Pipeline that runs GitGitGadget, then submitted https://public-inbox.org/git/pull.361.v3.git.gitgitgadget@gmail.com/, and it seems that it did not find the correct base commit or something, therefore still Cc:ing Junio.

lib/project-options.ts
```diff
@@ -39,7 +39,13 @@ export class ProjectOptions {
         let to: string;
         let midUrlPrefix: string = " Message-ID: ";
 
-        if (await commitExists("e83c5163316f89bfbde", workDir)) {
+        if (await commitExists("cb07fc2a29c86d1bc11", workDir) &&
+            await revParse(`${baseCommit}:git-gui.sh`) !== undefined) {
```
webstech: @dscho Should workdir be passed to revParse?  Still figuring out where git-gui.sh comes from.  Insight would be appreciated.

dscho: Yes! I _just_ came to the same conclusion, force-pushed a fix, only to see that you identified the culprit already 20 hours ago! 😄

dscho: > Still figuring out where git-gui.sh comes from. Insight would be appreciated.

The Git GUI project would seem to many as if it is merged into Git, but it is not. It is a separate project that happens to be pulled into Git from time to time, but Git GUI contributions are supposed to be independent from the Git project.

I say "supposed to" because for a couple of years, Git GUI appeared to be unmaintained (the official Git GUI maintainer simply did not react to mails nor Pull Requests). That situation changed, with Pratyush Yadav stepping up to the task. But the new maintainer wants to reuse the Git mailing list for code contributions, so it is _still_ only a "supposed to".

As a separate project, it is merged via a "subtree merge", i.e. the code is pulled into the `git-gui/` subdirectory. In the Git GUI project's commit history, therefore, the file `git-gui.sh` (which lives in `git-gui/git-gui.sh` in git.git) is located at the top-level directory.

And that is what this `revParse()` call is all about: does the base commit reflect the Git GUI commit history (as opposed to being subtree-merged, which would then be the Git commit history)?

dscho: I will force-push in a few moments, munging my latest reply into an amended commit message.

dscho: Force-pushed.

webstech: Thanks for the explanation.  Slowly getting an understanding.  When I look at [pull 358](https://api.github.com/repos/gitgitgadget/git/pulls/358) I can't find the base sha using the github search.  I am guessing that is a merge (that will be fetched with the gitgitgadget change).  So, will a merge include all the objects in the subtree or are we just assuming that git-gui.sh will be updated in every merge?  Could the base label (not currently passed to get()) be used to determine if it is git-gui?

dscho: > When I look at [pull 358](https://api.github.com/repos/gitgitgadget/git/pulls/358) I can't find the base sha using the github search.

It is the tip of https://github.com/prati0100/git-gui/commit/60c60b627e81bf84e1cb01729d2ae882178f079d's `master` branch, mirrored into GitGitGadget's Git fork as `git-gui/master` via https://dev.azure.com/gitgitgadget/git/_build?definitionId=10

> So, will a merge include all the objects in the subtree or are we just assuming that git-gui.sh will be updated in every merge?

Keep in mind that Git is not patch-based, but snapshot-based. That is, `<commit>:<path>` will always return the contents of the path in that commit, if that path is tracked.

So even if the merge was fetched via a shallow fetch, it would contain all the objects of its tree.

> Could the base label (not currently passed to get()) be used to determine if it is git-gui?

It could theoretically be used, yes. But I'd rather test for `<commit>:git-gui.sh`, as it is safer than trying to get all future naming conventions correct. After all, it is conceivable that Pratyush will eventually put in enough work to enable GitGitGadget on their Git GUI repository. In which case all naming conventions would go to hell.

webstech: Thanks for the snapshot reminder and clarifying things.

dscho:
@webstech thank you for helping with GitGitGadget. You have no idea how happy it makes me.

PR #128, opened 2019-10-06
Avoid useless `range-diff`s

dscho:
The Git maintainer pointed out that GitGitGadget could do a lot better job at identifying matching patches between subsequent patch series iterations, as we can be more certain than otherwise that all patches are related.

This PR tries to implement that suggestion.


dscho:
> On a side note, is the check for the range-diff command still needed?

Better safe than sorry ;-) At the moment, I registered an agent on a machine with a new-enough Git version, so all is good...

Issue #130, opened 2019-10-10
Errors running tests on Windows


webstech:
Getting two different errors related to two tsconfig.json settings that I am not sure how to correct.  The Azure CI tests (running on linux) do not get these errors.

This error is avoided by setting **"noImplicitAny": false** in tsconfig.json.
    lib/patch-series.ts:185:15 - error TS7005: Variable 'publishToRemote' implicitly has an 'any' type.

    185         const publishToRemote = undefined;

This error is avoided by setting **"strictNullChecks": false** in tsconfig.json.
    lib/git.ts:65:17 - error TS2531: Object is possibly 'null'.

    65                 process.stdout.on("data", (chunk: any) => {

Changing the tsconfig.json locally can result in pushing code that gets errors in the CI build, so that is not a good option.

typescript is at 3.5.3.

Any ideas or suggestions?

dscho:
>  const publishToRemote = undefined;

How about making it explicitly `: string | undefined`?

> process.stdout.on("data", (chunk: any) => {

Not sure what object that might be... maybe `process.stdout`? In that case, appending a `!` might work around this: `process.stdout!.on(...`

dscho:
Is this addressed via #132?

PR #132, opened 2019-10-12
Suppress "error TS2531: Object is possibly 'null'." errors.

webstech:
Resolve part of issue #130.  The `childProcess` in git.ts gets errors on Windows.  Using postfix ! notation to avoid compile time errors (thanks @dscho).


dscho:
Thanks!

PR #133, opened 2019-10-12
Support additional nodemailer options

webstech:
For testing, an alternate SMTP server may be used to avoid sending
emails.  Setting "gitgitgadget.smtpOpts" in the git config will
allow other nodemailer options to be specified.

Ethereal.email is an example server which does not actually send the
emails.  It does require other options to be passed to nodemailer.

Sample setting:
    await worktree.git(["config",
        '--add', "gitgitgadget.smtpHost", "smtp.ethereal.email"]);
    await worktree.git(["config",
        '--add', "gitgitgadget.smtpOpts",
        '{ "port": 587, "secure": false,
		"tls": { "rejectUnauthorized": false } }']);

### DIFF
--- a/lib/send-mail.ts
+++ b/lib/send-mail.ts
@@ -1,4 +1,5 @@
 import { createTransport, SendMailOptions } from "nodemailer";
+import SMTPTransport = require("nodemailer/lib/smtp-transport");
 import { decode } from "rfc2047";
 
 export interface IParsedMBox {
@@ -16,6 +17,7 @@ export interface IParsedMBox {
 export interface ISMTPOptions {
     smtpUser: string;
     smtpHost: string;
+    smtpOpts?: string;
     smtpPass: string;
 }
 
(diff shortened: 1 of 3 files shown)


lib/send-mail.ts
```diff
@@ -147,15 +149,25 @@ export async function parseMBoxMessageIDAndReferences(mbox: string):
...
+            user: smtpOptions.smtpUser,
+        },
+        host: smtpOptions.smtpHost,
+        secure: true,
+    };
+
+    if (smtpOptions.smtpOpts) {
+        const options: any = JSON.parse(smtpOptions.smtpOpts);
+        for (const key of Object.keys(options)) {
+            // transportOpts.key = options[`${key}`];
```
dscho: Did you want to keep the commented-out line?

Apart from that, I think there is a "more functional" way to do this:

```typescript
Object.entries(JSON.parse(smtpOptions.smtpOpts)).forEach(([key, value]) => transportOpts[key] = value);
```

What do you think?

webstech: Totally agree.  That is pretty much where I started before running into typing problems (resolved by type any).  Thanks for pointing that out.  Changes have been pushed.

PR #141, opened 2019-11-05
Make it slightly clearer how to iterate on a patch

Utsav2:


res/WELCOME.md
```diff
@@ -2,7 +2,7 @@
 
 Hi @${username}, and welcome to GitGitGadget, the GitHub App to send patch series to the Git mailing list from GitHub Pull Requests.
 
-Please make sure that this Pull Request has a good description, as it will be used as cover letter.
+Please make sure that your Pull Request has a good description, as it will be used as cover letter (the commit message for your patch).
```
dscho: I don't think that the addition "(the commit message for your patch)" is correct... the cover letter is the cover letter, it will be sent as a separate email, and the patch(es) will be sent as reply (or replies) to that cover letter.

dscho: I don't think that the addition "(the commit message for your patch)" is correct... the cover letter is the cover letter, it will be sent as a separate email, and the patch(es) will be sent as reply (or replies) to that cover letter.

res/WELCOME.md
```diff
@@ -29,7 +29,7 @@ An alternative is the channel [`#git-devel`](https://webchat.freenode.net/#git-d
 
 Once on the list of permitted usernames, you can contribute the patches to the Git mailing list by adding a PR comment `/submit`.
 
-After you submit, GitGitGadget will respond with another comment that contains the link to the cover letter mail in the Git mailing list archive. Please make sure to monitor the discussion in that thread and to address comments and suggestions.
+After you submit, GitGitGadget will respond with another comment that contains the link to the cover letter mail in the Git mailing list archive. Please make sure to monitor the discussion in that thread and to address comments and suggestions. To iterate on your change, push to the same branch or modify your pull request title or description, and comment `/submit` again.
```
dscho: Maybe "(force-)push" instead of "push"?

Also, I don't think that GitGitGadget lets you submit a new iteration with only a changed PR description; it requires a change to at least one of the commits.

Finally, how about suggesting to summarize the changes in the cover letter, e.g. like this:

```
Changes since v1:
- Fixed a typo in the commit message (found by ...)
- Added a code comment to ... as suggested by ...
...
```

res/WELCOME.md
Utsav2: Ok. I think we can make that clearer as well, as I was under the wrong impression.

PR #142, opened 2019-11-12
Fix the "double double quotes" problem in the "<author> via GitGitGadget" From: lines

dscho:
This will hopefully prevent problems like the one that prevented https://github.com/gitgitgadget/git/pull/435 from arriving at the Git mailing list.

### DIFF
--- a/lib/patch-series.ts
+++ b/lib/patch-series.ts
@@ -376,7 +376,8 @@ export class PatchSeries {
                 // Special-case GitGitGadget to send from
                 // "<author> via GitGitGadget"
                 replaceSender = "\""
-                    + onBehalfOf
+                    + onBehalfOf.replace(/^"(.*)"$/, "$1")
+                                .replace(/"/g, "\\\"")
                     + " via GitGitGadget\" "
                     + thisAuthor.replace(/^GitGitGadget /, "");
             } else if (authorMatch[2] === thisAuthor) {
--- a/tests/gitgitgadget.test.ts
+++ b/tests/gitgitgadget.test.ts
@@ -10,7 +10,7 @@ import { testCreateRepo } from "./test-lib";
 jest.setTimeout(60000);
 
 const expectedMails = [
-    `From 07f68c195159518c5777ca4a7c1d07124e7a9956 Mon Sep 17 00:00:00 2001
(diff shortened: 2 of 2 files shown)


lib/patch-series.ts
```diff
@@ -376,7 +376,8 @@ export class PatchSeries {
                 // Special-case GitGitGadget to send from
                 // "<author> via GitGitGadget"
                 replaceSender = "\""
-                    + onBehalfOf
+                    + onBehalfOf.replace(/^"(.*)"$/, "$1")
```
webstech: Can `authorMatch[2]` ever be quoted (in the `isGitGitGadget` case)?  If not, should 
`PatchSeries.encodeSender(senderName)` at line 374 be `PatchSeries.encodeWords(senderName)`?  It would remove the need to dequote.

dscho: > Can `authorMatch[2]` ever be quoted (in the `isGitGitGadget` case)?

Yes. The `authorMatch` variable contains matches from the `From:` lines generated by `git format-patch`. If `user.name` contains a `.` (as was in the case in the PR that was not sent to the mailing list), then that `From:` line indeed contains double quotes.

webstech: > Yes. The `authorMatch` variable contains matches from the `From:` lines generated by `git format-patch`. If `user.name` contains a `.` (as was in the case in the PR that was not sent to the mailing list), then that `From:` line indeed contains double quotes.

Okay, so format-patch is also quoting.  The PR looks good.  Not sure of the chance of the second replace happening.  Would git allow double quotes in the user.name and if so, would it already be escaped?

This may explain why my initial PR for Git did not make it to the mailing list.  You had to forward it.  Looking at some of the emails I noticed an unquoted copy of my name so not sure if there is something else lurking but that would be a different problem.

dscho: > Not sure of the chance of the second replace happening. Would git allow double quotes in the user.name and if so, would it already be escaped?

Neither am I, and hopefully we will never have to find out. I just wanted to err on the side of caution.

> This may explain why my initial PR for Git did not make it to the mailing list. You had to forward it. Looking at some of the emails I noticed an unquoted copy of my name so not sure if there is something else lurking but that would be a different problem.

Indeed! That could have been the problem all along...

PR #147, opened 2019-11-18
Allow GitGitGadget to Work Without a PR Template

webstech:
To support other projects and testing, `try/catch` to avoid errors if there is no template file.

In `TestRepo.commit()`, create directories if they are present in the `fileName` parameter.

### DIFF
--- a/lib/gitgitgadget.ts
+++ b/lib/gitgitgadget.ts
@@ -260,14 +260,19 @@ export class GitGitGadget {
         // update work repo from base
         await this.updateNotesAndPullRef(pr.number, previousTag);
 
-        // Remove template from description
-        let prTemplate =
-            await git(["show",
-                       "upstream/master:.github/PULL_REQUEST_TEMPLATE.md"],
-                      { workDir: this.workDir });
-        // github uses \r\n so make sure it is set
-        prTemplate = prTemplate.replace(/([^\r])\n/, "$1\r\n");
-        const prBody = pr.body.replace(prTemplate, "");
+        // Remove template from description (if template exists)
+        let prBody: string;
+        try {
+            let prTemplate =
+                await git(["show",
+                           "upstream/master:.github/PULL_REQUEST_TEMPLATE.md"],
+                          { workDir: this.workDir });
+            // github uses \r\n so make sure it is set
+            prTemplate = prTemplate.replace(/\r?\n/g, "\r\n");
+            prBody = pr.body.replace(prTemplate, "");
+        } catch {
+            prBody = pr.body;
+        }
 
         if (!prBody.length) {       // reject empty description
             throw new Error("A pull request description must be provided");
--- a/tests/test-lib.ts
+++ b/tests/test-lib.ts
@@ -1,4 +1,5 @@
 import * as fs from "fs";
+import * as path from "path";
 import * as util from "util";
 import { isDirectory, isFile } from "../lib/fs-util";
 import { git, IGitOptions, revParse } from "../lib/git";
@@ -45,6 +46,11 @@ export class TestRepo {
 
         if (!fileName) {
             fileName = `${message}.t`;
+        }
+
+        const fPath = path.dirname(`${this.workDir}/${fileName}`);
+        if (fPath !== this.workDir) {
+            await mkdir(fPath, { recursive: true });
         }
 
         await writeFile(`${this.workDir}/${fileName}`, contents || message);


lib/gitgitgadget.ts
```diff
@@ -260,14 +260,19 @@ export class GitGitGadget {
...
-                      { workDir: this.workDir });
-        // github uses \r\n so make sure it is set
-        prTemplate = prTemplate.replace(/([^\r])\n/, "$1\r\n");
-        const prBody = pr.body.replace(prTemplate, "");
+        // Remove template from description (if template exists)
+        let prBody: string;
+        try {
+            let prTemplate =
+                await git(["show",
+                           "upstream/master:.github/PULL_REQUEST_TEMPLATE.md"],
```
dscho: It occurred to me last night that PR templates might differ depending on the base branch. Do you know?

dscho:
> To support other projects

I wonder what other projects those might be... 😄

lib/gitgitgadget.ts
```diff
@@ -260,14 +260,19 @@ export class GitGitGadget {
...
+                          { workDir: this.workDir });
+            // github uses \r\n so make sure it is set
+            prTemplate = prTemplate.replace(/([^\r])\n/, "$1\r\n");
```
dscho: Oh, and wouldn't this be better `.replace(/\r?\n/g, "\r\n")`? I.e. replace _all_ occurrences, and look for an optional `\r` before the `\n`?

lib/gitgitgadget.ts
webstech: Not sure what you asking.  This change assumes the base branch is master.  If it was not, then we would need to parse the `baseLabel` to get the name.  If I cloned and changed the base branch in my repo and did a pull request to my repo I could have a different template.

lib/gitgitgadget.ts
webstech: I had intended to do a global replace.  The regex excludes `\r` but I like yours better.  Pushed an update that only changes the regex.

lib/gitgitgadget.ts
dscho: > This change assumes the base branch is master.

Many are not, though...

> If it was not, then we would need to parse the `baseLabel` to get the name.

Yes, I think we will want to do that. IIRC the `baseLabel` starts with the `${repositoryOwner}:` prefix, though, so we will want to be careful to use the correct thing there. Maybe `baseCommit` would be a better idea?

> If I cloned and changed the base branch in my repo and did a pull request to my repo I could have a different template.

True. But how common is it that somebody opens a PR, _forgets_ to remove the template text, _and_ is savvy enough to change the target branch?

lib/gitgitgadget.ts
dscho: (Technically, it is probably poor practice to sneak in that fix in a commit that claims to only surround that code block in `try`/`catch`... but I won't object if you do not want to separate those two changes into two commits.)

lib/gitgitgadget.ts
webstech: > > This change assumes the base branch is master.
> 
> Many are not, though...

We may be talking about different bases here.  I was referring to the base/default branch for the repo, which in the case of git is master.  That is my understanding of what GitHub uses for the template.  Are you referring to the base for the pull request, which I agree may be different?

> > If it was not, then we would need to parse the `baseLabel` to get the name.
> 
> Yes, I think we will want to do that. IIRC the `baseLabel` starts with the `${repositoryOwner}:` prefix, though, so we will want to be careful to use the correct thing there. Maybe `baseCommit` would be a better idea?

It might actually be `base.repo.default_branch` that is needed.  It can be added to `IPullRequestInfo`.  Can this be deferred until it is actually needed?  Otherwise, the fetch of upstream will also need to ensure it is pulling that branch.

dscho: > > > This change assumes the base branch is master.
> > 
> > 
> > Many are not, though...
> 
> We may be talking about different bases here. I was referring to the base/default branch for the repo, which in the case of git is master. That is my understanding of what GitHub uses for the template.

I am not sure that the `master` version is used always, even when the user targets a different branch. Is this documented somewhere?

If it is _always_ the `master` version, then I retract my suggestion and declare the current version fine ;-)

dscho: >  Is this documented somewhere?

It is ;-) https://help.github.com/en/github/building-a-strong-community/about-issue-and-pull-request-templates#pull-request-templates says:

> You must create templates on the repository's default branch. Templates created in other branches are not available for collaborators to use.

So: I retract my suggestion to look at `baseCommit` ;-) And since we are mirroring git/git's `master` into gitgitgadget/git always, `upstream/master` is good enough and you do not have to do anything differently.

Sorry for the noise!

webstech: I recently found out if you push a new repo to github with branch foo having commits and no commits on master that branch foo is the default branch on github.  Had to reset in the settings.  😮

dscho: Oh yes, that happened to me, too. I was quite surprised by that, but it kinda makes sense.

Issue #140, opened 2019-11-05
Configuration Problem Testing CIHelper


webstech:
CIHelper expects to be in the gitgitgadget repo so it can reference the res/Welcome.MD file.  When testing, the workdir for CIHelper needs to point to different locations so the tests can run in parallel and the git config can be set outside of the tester's personal config.  

In CIHelper, when creating a GitGitGadget, it is done with `GitGitGadget.get(".")`.  The get method will query the config for `gitgitgadget.workDir` using the passed '.' directory (IOW, the git config for gitgitgadget).

1. This seems awkward because the CIHelper workdir is supposed to point to a git repository.  Why not just pass that to `GitGitGadget.get()`?  
2. The get method is using the passed directory to obtain git config info.  Is this supposed to be part of the git repo config or the gitgitgadget repo config?  The get method does support two directories being passed.
3. Concurrent testing has problems with different tests setting `gitgitgadget.workDir`.  This can be worked around by changing directories to a different git config but that also requires copying res/Welcome.MD.

Hopefully this makes some sense.  One solution is to provide CIHelper with a git repo (as is done now), and a location to get the config info (default to '.').

Questions?  Other suggestions?  Have I misunderstood everything?

dscho:
@webstech I think you are right on all accounts. To allow for easy deployment, I would still like to default to querying `gitgitgadget.workDir` in `.`, but you've correctly pointed out that it would make a lot of sense to allow overriding that with a pair of Git/GitGitGadget workDirs (we often need the Git workDir to look at commits and notes, and we do need the GitGitGadget workDir for the welcome message).

webstech:
Thanks for the quick reply.  I know you have been very busy.

For minimal change, the CIHelper would be 
`constructor(workDir?: string, skipUpdate?: boolean, gggDir = "." )`
and getting the gadget would be
`GitGitGadget.get(this.gggDir, this.workDir)`.

The oddity is the reversed order from the constructor to the get method but they are different objects.

I do have a question related to deployment.  How is the git configuration set up in azure?  Is it part of the vm setup?  Just trying to understand how this works to deploy on fresh vm.

dscho:
> For minimal change, the CIHelper would be
> `constructor(workDir?: string, skipUpdate?: boolean, gggDir = "." )`
> and getting the gadget would be
> `GitGitGadget.get(this.gggDir, this.workDir)`.
> 
> The oddity is the reversed order from the constructor to the get method but they are different objects.

[...]

PR #148, opened 2019-11-18
Support PRs over at https://github.com/git/git, too

dscho:
@peff asked for this a couple of times, e.g. at https://public-inbox.org/git/20191113011020.GB20431@sigill.intra.peff.net/

This PR should do that trick, together with a couple of changes to GitGitGadget's Azure Pipelines (and a new one).

Note that the `TO-DROP` patches at the end are there only for testing; I did send a test patch series (and a second iteration) to myself, and the resulting tags can be seen here: https://github.com/gitgitgadget/git/releases/tag/pr-dscho-12%2Fdscho%2Ffix-rebase-r-labels-v1 and https://github.com/gitgitgadget/git/releases/tag/pr-dscho-12%2Fdscho%2Ffix-rebase-r-labels-v2.

I would appreciate all the code reviews I can get, although I _am_ expecting having to work out the kinks after merging this PR.

### DIFF
--- a/azure-function/index.js
+++ b/azure-function/index.js
@@ -1,3 +1,4 @@
+// @ts-check
 /*
  * This is the Azure Function backing the GitGitGadget GitHub App.
  *
--- a/azure-function/index.js
+++ b/azure-function/index.js
@@ -91,12 +91,29 @@ module.exports = async (context, req) => {
     }
 
     try {
+        /*
+         * The Azure Pipeline needs to be installed as a PR build on _the very
+         * same_ repository that triggers this function. That is, when the
+         * Azure Function triggers GitGitGadget for gitgitgadget/git, it needs
+         * to know that pipelineId 3 is installed on gitgitgadget/git, and
+         * trigger that very pipeline.
+         *
+         * So whenever we extend GitGitGadget to handle another repository, we
+         * will have to add an Azure Pipeline, install it on that repository as
+         * a PR build, and add the information here.
+         */
+        const pipelines = {
+            'git': 13,
+            'gitgitgadget': 3,
+        };
+
         const eventType = context.req.headers['x-github-event'];
         context.log(`Got eventType: ${eventType}`);
-        if (req.body.repository.owner.login !== 'gitgitgadget') {
+        const repositoryOwner = req.body.repository.owner.login;
+        if (pipelines[repositoryOwner] === undefined) {
             context.res = {
                 status: 403,
-                body: 'Refusing to work on a repository other than gitgitgadget/git'
+                body: 'Refusing to work on a repository other than gitgitgadget/git or git/git'
             };
         } else if ((new Set(['check_run', 'status']).has(eventType))) {
             context.res = {
@@ -128,8 +145,11 @@ module.exports = async (context, req) => {
             const parameters = {
                 'pr.comment.id': comment.id,
             };
+            const pipelineId = pipelines[repositoryOwner];
+            if (!pipelineId || pipelineId < 1)
+                throw new Error(`No pipeline set up for org ${repositoryOwner}`);
             context.log(`Queuing with branch ${sourceBranch} and parameters ${JSON.stringify(parameters)}`);
-            await triggerAzurePipeline(triggerToken, 'gitgitgadget', 'git', 3, sourceBranch, parameters);
+            await triggerAzurePipeline(triggerToken, 'gitgitgadget', 'git', pipelineId, sourceBranch, parameters);
 
             context.res = {
                 // status: 200, /* Defaults to 200 */
--- a/azure-function/index.js
+++ b/azure-function/index.js
@@ -103,6 +103,7 @@ module.exports = async (context, req) => {
          * a PR build, and add the information here.
          */
         const pipelines = {
+            'dscho': 12,
             'git': 13,
             'gitgitgadget': 3,
         };
@@ -130,6 +131,11 @@ module.exports = async (context, req) => {
             if (!comment || !comment.id || !prNumber) {
                 context.log(`Invalid payload:\n${JSON.stringify(req.body, null, 4)}`);
                 throw new Error('Invalid payload');
+            }
+
+            /* GitGitGadget works on dscho/git only for testing */
+            if (repositoryOwner === 'dscho' && comment.user.login !== 'dscho') {
+                throw new Error(`Ignoring comment from ${comment.user.login}`);
             }
 
             /* Only trigger the Pipeline for valid commands */
--- a/azure-function/index.js
+++ b/azure-function/index.js
@@ -109,8 +109,7 @@ module.exports = async (context, req) => {
             }
 
             const comment = req.body.comment;
-            const pullRequestURL = req.body.issue.pull_request.html_url;
-            const prNumber = pullRequestURL.match(/https:\/\/github\.com\/gitgitgadget\/git\/pull\/([0-9]*)$/)
+            const prNumber = req.body.issue.number;
             if (!comment || !comment.id || !prNumber) {
                 context.log(`Invalid payload:\n${JSON.stringify(req.body, null, 4)}`);
                 throw new Error('Invalid payload');
@@ -125,7 +124,7 @@ module.exports = async (context, req) => {
                 return;
             }
 
-            const sourceBranch = `refs/pull/${prNumber[1]}/head`;
+            const sourceBranch = `refs/pull/${prNumber}/head`;
             const parameters = {
                 'pr.comment.id': comment.id,
             };
--- a/azure-function/index.js
+++ b/azure-function/index.js
@@ -111,7 +111,7 @@ module.exports = async (context, req) => {
             const comment = req.body.comment;
             const pullRequestURL = req.body.issue.pull_request.html_url;
             const prNumber = pullRequestURL.match(/https:\/\/github\.com\/gitgitgadget\/git\/pull\/([0-9]*)$/)
-            if (!comment.id || !prNumber) {
+            if (!comment || !comment.id || !prNumber) {
                 context.log(`Invalid payload:\n${JSON.stringify(req.body, null, 4)}`);
                 throw new Error('Invalid payload');
             }
--- a/lib/ci-helper.ts
+++ b/lib/ci-helper.ts
@@ -117,7 +117,9 @@ export class CIHelper {
         }
         await this.notes.set(messageID, mailMeta, true);
 
-        if (!this.testing && mailMeta.pullRequestURL) {
+        if (!this.testing && mailMeta.pullRequestURL &&
+            mailMeta.pullRequestURL
+            .startsWith("https://github.com/gitgitgadget/") ) {
             await this.github.annotateCommit(mailMeta.originalCommit,
                                              upstreamCommit);
         }
--- a/lib/ci-helper.ts
+++ b/lib/ci-helper.ts
@@ -487,8 +487,10 @@ export class CIHelper {
      *
      * @param commentID the ID of the PR comment to handle
      */
-    public async handleComment(commentID: number): Promise<void> {
-        const comment = await this.github.getPRComment(commentID);
+    public async handleComment(repositoryOwner: string, commentID: number):
+        Promise<void> {
+        const comment =
+            await this.github.getPRComment(repositoryOwner, commentID);
         const match = comment.body.match(/^\s*(\/[-a-z]+)(\s+(.*?))?\s*$/);
         if (!match) {
             console.log(`Not a command; doing nothing: '${comment.body}'`);
@@ -497,8 +499,8 @@ export class CIHelper {
         const command = match[1];
         const argument = match[3];
 
-        const pullRequestURL =
-            `https://github.com/gitgitgadget/git/pull/${comment.prNumber}`;
+        const pullRequestURL = `https://github.com/${
+            repositoryOwner}/git/pull/${comment.prNumber}`;
         console.log(`Handling command ${command} with argument ${argument} at ${
             pullRequestURL}#issuecomment-${commentID}`);
 
@@ -515,7 +517,8 @@ export class CIHelper {
             }
 
             const getPRAuthor = async (): Promise<string> => {
-                const pr = await this.github.getPRInfo(comment.prNumber);
+                const pr = await this.github.getPRInfo(repositoryOwner,
+                                                       comment.prNumber);
                 return pr.author;
             };
 
@@ -605,12 +608,12 @@ export class CIHelper {
         }
     }
 
-    public async handlePush(prNumber: number) {
-        const pr = await this.github.getPRInfo(prNumber);
+    public async handlePush(repositoryOwner: string, prNumber: number) {
+        const pr = await this.github.getPRInfo(repositoryOwner, prNumber);
         const gitGitGadget = await GitGitGadget.get(".");
         if (!pr.hasComments && !gitGitGadget.isUserAllowed(pr.author)) {
             const pullRequestURL =
-                `https://github.com/gitgitgadget/git/pull/${prNumber}`;
+                `https://github.com/${repositoryOwner}/git/pull/${prNumber}`;
             const welcome = (await readFile("res/WELCOME.md")).toString()
                 .replace(/\${username}/g, pr.author);
             this.github.addPRComment(pullRequestURL, welcome);
@@ -634,7 +637,9 @@ export class CIHelper {
 
     private async getPRInfo(prNumber: number, pullRequestURL: string):
         Promise<IPullRequestInfo> {
-        const pr = await this.github.getPRInfo(prNumber);
+        const [owner] =
+                GitGitGadget.parsePullRequestURL(pullRequestURL);
+        const pr = await this.github.getPRInfo(owner, prNumber);
 
         if (!pr.baseLabel || !pr.baseCommit ||
             !pr.headLabel || !pr.headCommit) {
--- a/lib/ci-helper.ts
+++ b/lib/ci-helper.ts
@@ -121,7 +121,7 @@ export class CIHelper {
             mailMeta.pullRequestURL
             .startsWith("https://github.com/gitgitgadget/") ) {
             await this.github.annotateCommit(mailMeta.originalCommit,
-                                             upstreamCommit);
+                                             upstreamCommit, "gitgitgadget");
         }
 
         return true;
--- a/lib/gitgitgadget.ts
+++ b/lib/gitgitgadget.ts
@@ -195,7 +195,8 @@ export class GitGitGadget {
         return await this.genAndSend(pr, userInfo, {}, send);
     }
 
-    protected async updateNotesAndPullRef(pullRequestNumber: number,
+    protected async updateNotesAndPullRef(repositoryOwner: string,
+                                          pullRequestNumber: number,
                                           additionalRef?: string):
         Promise<string> {
         const pullRequestRef = `refs/pull/${pullRequestNumber}/head`;
@@ -205,15 +206,24 @@ export class GitGitGadget {
             this.publishTagsAndNotesToRemote,
             "--",
             `+${this.notes.notesRef}:${this.notes.notesRef}`,
-            `+${pullRequestRef}:${pullRequestRef}`,
-            `+${pullRequestMerge}:${pullRequestMerge}`,
             `+refs/heads/maint:refs/remotes/upstream/maint`,
             `+refs/heads/master:refs/remotes/upstream/master`,
             `+refs/heads/next:refs/remotes/upstream/next`,
             `+refs/heads/pu:refs/remotes/upstream/pu`,
         ];
+        const prArgs = [
+            `+${pullRequestRef}:${pullRequestRef}`,
+            `+${pullRequestMerge}:${pullRequestMerge}`,
+        ];
         if (additionalRef) {
             args.push(`+${additionalRef}:${additionalRef}`);
+        }
+        if (repositoryOwner === "gitgitgadget") {
+            args.push(...prArgs);
+        } else {
+            prArgs.unshift("fetch", `https://github.com/${repositoryOwner}/git`,
+                           "--");
+            await git(prArgs, { workDir: this.workDir });
         }
         await git(args, { workDir: this.workDir });
 
@@ -263,7 +273,7 @@ export class GitGitGadget {
         const previousTag = metadata && metadata.latestTag ?
             `refs/tags/${metadata.latestTag}` : undefined;
         // update work repo from base
-        await this.updateNotesAndPullRef(pr.number, previousTag);
+        await this.updateNotesAndPullRef(pr.baseOwner, pr.number, previousTag);
 
         const series =
             await PatchSeries.getFromNotes(this.notes, pr.pullRequestURL,
--- a/lib/gitgitgadget.ts
+++ b/lib/gitgitgadget.ts
@@ -258,7 +258,8 @@ export class GitGitGadget {
                                send: SendFunction):
         Promise<string | undefined> {
 
-        if (pr.baseOwner !== "gitgitgadget" || pr.baseRepo !== "git") {
+        if (!new Set(["gitgitgadget", "git"]).has(pr.baseOwner) ||
+            pr.baseRepo !== "git") {
             throw new Error(`Unsupported repository: ${pr.pullRequestURL}`);
         }
 
--- a/lib/gitgitgadget.ts
+++ b/lib/gitgitgadget.ts
@@ -258,7 +258,7 @@ export class GitGitGadget {
                                send: SendFunction):
         Promise<string | undefined> {
 
-        if (!new Set(["gitgitgadget", "git"]).has(pr.baseOwner) ||
+        if (!new Set(["gitgitgadget", "dscho", "git"]).has(pr.baseOwner) ||
             pr.baseRepo !== "git") {
             throw new Error(`Unsupported repository: ${pr.pullRequestURL}`);
         }
--- a/lib/github-glue.ts
+++ b/lib/github-glue.ts
@@ -234,9 +234,10 @@ export class GitHubGlue {
      * @param prNumber the Pull Request's number
      * @returns information about that Pull Request
      */
-    public async getPRInfo(prNumber: number): Promise<IPullRequestInfo> {
+    public async getPRInfo(repositoryOwner: string, prNumber: number):
+        Promise<IPullRequestInfo> {
         const response = await this.client.pulls.get({
-            owner: "gitgitgadget",
+            owner: repositoryOwner,
             pull_number: prNumber,
             repo: "git",
         });
@@ -263,10 +264,11 @@ export class GitHubGlue {
      * @param commentID the ID of the PR/issue comment
      * @returns the text in the comment
      */
-    public async getPRComment(commentID: number): Promise<IPRComment> {
+    public async getPRComment(repositoryOwner: string, commentID: number):
+        Promise<IPRComment> {
         const response = await this.client.issues.getComment({
             comment_id: commentID,
-            owner: "gitgitgadget",
+            owner: repositoryOwner,
             repo: "git",
         });
         const match = response.data.html_url.match(/\/pull\/([0-9]+)/);
--- a/lib/github-glue.ts
+++ b/lib/github-glue.ts
@@ -33,8 +33,8 @@ export interface IGitHubUser {
 
 export class GitHubGlue {
     public workDir?: string;
-    protected readonly client = new octokit();
-    protected authenticated = false;
+    protected client = new octokit();
+    protected authenticated?: string;
 
     public constructor(workDir?: string) {
         this.workDir = workDir;
@@ -52,7 +52,7 @@ export class GitHubGlue {
         const [, short, completedAt] = match;
         const url = `https://github.com/git/git/commit/${gitGitCommit}`;
 
-        await this.ensureAuthenticated();
+        await this.ensureAuthenticated("gitgitgadget");
         const checks = await this.client.checks.create({
             completed_at: completedAt,
             conclusion: "success",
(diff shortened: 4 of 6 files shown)


dscho:
As usual, GitHub's UI juggles the commits around because it orders by date rather than by topology. And the test is failing because I removed the Cc: for my testing ;-)

dscho:
> And the test is failing because I removed the Cc: for my testing ;-)

Note: I pushed those `TO-DROP` commits to the `work-on-git.git-debug` branch in my fork, and dropped from this here PR.

webstech:
Lots to look at here.  I will post comments as I go through it.  Hopefully I won't add too many more as I circle between sources.

script/misc-helper.ts
```diff
@@ -332,31 +332,52 @@ async function getCIHelper(): Promise<CIHelper> {
         const glue = new GitHubGlue();
         await glue.addPRComment(pullRequestURL, comment);
     } else if (command === "set-app-token") {
-        if (commander.args.length !== 1) {
+        if (commander.args.length < 1) {
```
webstech: Is this check still needed?  The code will run with no args won't it (assuming slice will have no issues)?  If it stays, maybe the message should be `unexpected` to `missing`.

script/misc-helper.ts
```diff
@@ -332,31 +332,52 @@ async function getCIHelper(): Promise<CIHelper> {
...
+                options.installationID =
+                    (await client.apps.getRepoInstallation({
+                        owner: options.name,
+                        repo: "git",
+                })).data.id;
+            }
+            const result = await client.apps.createInstallationToken({
+                installation_id: options.installationID!,
+            });
+            const configKey = options.name === "gitgitgadget" ?
```
webstech: Why have different values for `configKey`?  ie why not always set to `gitgitgadget.${options.name}.githubToken`?  This would make it simpler in github-glue.ts?

lib/github-glue.ts
```diff
@@ -52,7 +52,7 @@ export class GitHubGlue {
         const [, short, completedAt] = match;
         const url = `https://github.com/git/git/commit/${gitGitCommit}`;
 
-        await this.ensureAuthenticated();
+        await this.ensureAuthenticated(this.authenticated || "gitgitgadget");
```
webstech: Not clear why `this.authenticated` would be used.  Isn't this an update to the gitgitgadget repo?  Or, for testing, should `owner` in the json use this.authenticated?

lib/github-glue.ts
```diff
@@ -80,9 +80,9 @@ export class GitHubGlue {
      */
     public async addPRComment(pullRequestURL: string, comment: string):
         Promise<{id: number, url: string}> {
-        await this.ensureAuthenticated();
         const [owner, repo, nr] =
             GitGitGadget.parsePullRequestURL(pullRequestURL);
```
webstech: For future consideration, change `pullRequestURL` to an object with `[owner, repo, nr]`.  There seems to be a lot of reparsing of the url.  I understand the original script did not have the benefit of objects.

azure-function/index.js
```diff
@@ -109,12 +109,17 @@ module.exports = async (context, req) => {
 
             const comment = req.body.comment;
             const pullRequestURL = req.body.issue.pull_request.html_url;
-            const prNumber = pullRequestURL.match(/https:\/\/github\.com\/gitgitgadget\/git\/pull\/([0-9]*)$/)
-            if (!comment.id || !prNumber) {
+            const match = pullRequestURL.match(/https:\/\/github\.com\/([^/]*)\/git\/pull\/([0-9]*)$/);
+            if (!comment || !comment.id || !match || match[1] != req.body.repository.owner.login) {
```
webstech: Why not use `req.body.issue.pull_request.number` and `req.body.repository.owner.login` instead of extracting from the url?

azure-function/index.js
```diff
@@ -124,12 +129,16 @@ module.exports = async (context, req) => {
                 return;
             }
 
-            const sourceBranch = `refs/pull/${prNumber[1]}/head`;
+            const prNumber = match[2];
+            const sourceBranch = `refs/pull/${prNumber}/head`;
             const parameters = {
                 'pr.comment.id': comment.id,
```
webstech: Why not pass the owner as part of the parms to the pipeline?  It just looks like a magic parameter in misc-helper.ts at the moment.  Of course, it this was to ever support a repo besides git, the repo name would need to be passed as well.

lib/ci-helper.ts
```diff
@@ -117,7 +117,9 @@ export class CIHelper {
         }
         await this.notes.set(messageID, mailMeta, true);
 
-        if (!this.testing && mailMeta.pullRequestURL) {
+        if (!this.testing && mailMeta.pullRequestURL &&
+            mailMeta.pullRequestURL.match(
+                /^https:\/\/github\.com\/gitgitgadget\//) ) {
```
sgn: I'm not familiar with typescript.
Does it have `string.startsWith`? If yes, it would be 100% clearer over regex.

azure-function/index.js
dscho: > Why not pass the owner as part of the parms to the pipeline?

It is a magic parameter. And it kind of has to be: the Azure Pipeline needs to be installed _on that very same repository_. That is, when the Azure Function triggers GitGitGadget for git/git, it needs to know that pipelineId 13 is installed on git/git, and trigger that pipeline, if GitGitGadget should handle a PR at dscho/git, it needs to use the pipeline with the ID 12, and for regular PRs at gitgitgadget/git, the pipeline with ID 3 must be used.

The Azure Function therefore has to have intricate knowledge which pipelines were installed and which repository they target. I kinda hoped that the long commit message of https://github.com/gitgitgadget/gitgitgadget/pull/148/commits/8b08068e27aae1d27fb0195cbe3a776aae23b8bb was detailed enough?

lib/ci-helper.ts
dscho: > I'm not familiar with typescript.

Neither am I.

> Does it have `string.startsWith`?

What I do in such cases is scour the Javascript documentation (because Typescript is a superset of Javascript, therefore if Javascript has that method, so does Typescript). And so can you!

*clicketyclick*

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith suggests that this method is available. So I will make that change.

script/misc-helper.ts
dscho: Well spotted. Will fix.

script/misc-helper.ts
dscho: Legacy reasons. The Azure Pipeline uses the same token also for pushing, outside of the Typescript code, in a Bash script step.

dscho: (And I do want that Azure Pipeline to continue working seamlessly both before and after this PR is merged.)

lib/github-glue.ts
dscho: You're right. Will fix.

lib/github-glue.ts
dscho: Good point. I opened https://github.com/gitgitgadget/gitgitgadget/issues/150 as a reminder.

azure-function/index.js
dscho: You know what, I'll just add a code comment about that to `index.js`.

azure-function/index.js
dscho: I remember now why I did not do it that way: it does not work. You might _think_ that `req.body.issue.pull_request` is a fully-populated JSON as e.g. described at https://developer.github.com/v3/pulls/#get-a-single-pull-request, but it is not. The full `req.body` from an example run:

```json
{
  action: 'edited',
  changes: { body: { from: '/test again' } },
  issue: 
   {
     url: 'https://api.github.com/repos/dscho/git/issues/12',
     repository_url: 'https://api.github.com/repos/dscho/git',
     labels_url: 'https://api.github.com/repos/dscho/git/issues/12/labels{/name}',
     comments_url: 'https://api.github.com/repos/dscho/git/issues/12/comments',
     events_url: 'https://api.github.com/repos/dscho/git/issues/12/events',
     html_url: 'https://github.com/dscho/git/pull/12',
     id: 524054282,
     node_id: 'MDExOlB1bGxSZXF1ZXN0MzQxOTAyNDYw',
     number: 12,
     title: '[TEST] rebase -r: let `label` generate safer labels',
     user: 
      {
        login: 'dscho',
        id: 127790,
        node_id: 'MDQ6VXNlcjEyNzc5MA==',
        avatar_url: 'https://avatars0.githubusercontent.com/u/127790?v=4',
        gravatar_id: '',
        url: 'https://api.github.com/users/dscho',
        html_url: 'https://github.com/dscho',
        followers_url: 'https://api.github.com/users/dscho/followers',
        following_url: 'https://api.github.com/users/dscho/following{/other_user}',
        gists_url: 'https://api.github.com/users/dscho/gists{/gist_id}',
        starred_url: 'https://api.github.com/users/dscho/starred{/owner}{/repo}',
        subscriptions_url: 'https://api.github.com/users/dscho/subscriptions',
        organizations_url: 'https://api.github.com/users/dscho/orgs',
        repos_url: 'https://api.github.com/users/dscho/repos',
        events_url: 'https://api.github.com/users/dscho/events{/privacy}',
        received_events_url: 'https://api.github.com/users/dscho/received_events',
        type: 'User',
        site_admin: false },
     labels: [],
     state: 'open',
     locked: false,
     assignee: null,
     assignees: [],
     milestone: null,
     comments: 13,
     created_at: '2019-11-17T21:57:46Z',
     updated_at: '2019-11-19T16:30:14Z',
     closed_at: null,
     author_association: 'OWNER',
     pull_request: 
      {
        url: 'https://api.github.com/repos/dscho/git/pulls/12',
        html_url: 'https://github.com/dscho/git/pull/12',
        diff_url: 'https://github.com/dscho/git/pull/12.diff',
        patch_url: 'https://github.com/dscho/git/pull/12.patch' },
     body: 'This is a test for GitGitGadget.' },
  comment: 
   {
     url: 'https://api.github.com/repos/dscho/git/issues/comments/555142445',
     html_url: 'https://github.com/dscho/git/pull/12#issuecomment-555142445',
     issue_url: 'https://api.github.com/repos/dscho/git/issues/12',
     id: 555142445,
     node_id: 'MDEyOklzc3VlQ29tbWVudDU1NTE0MjQ0NQ==',
     user: 
      {
        login: 'dscho',
        id: 127790,
        node_id: 'MDQ6VXNlcjEyNzc5MA==',
        avatar_url: 'https://avatars0.githubusercontent.com/u/127790?v=4',
        gravatar_id: '',
        url: 'https://api.github.com/users/dscho',
        html_url: 'https://github.com/dscho',
        followers_url: 'https://api.github.com/users/dscho/followers',
        following_url: 'https://api.github.com/users/dscho/following{/other_user}',
        gists_url: 'https://api.github.com/users/dscho/gists{/gist_id}',
        starred_url: 'https://api.github.com/users/dscho/starred{/owner}{/repo}',
        subscriptions_url: 'https://api.github.com/users/dscho/subscriptions',
        organizations_url: 'https://api.github.com/users/dscho/orgs',
        repos_url: 'https://api.github.com/users/dscho/repos',
        events_url: 'https://api.github.com/users/dscho/events{/privacy}',
        received_events_url: 'https://api.github.com/users/dscho/received_events',
        type: 'User',
        site_admin: false },
     created_at: '2019-11-18T18:18:38Z',
     updated_at: '2019-11-19T16:30:14Z',
     author_association: 'OWNER',
     body: '/testtest again' },
  repository: 
   {
     id: 2326047,
     node_id: 'MDEwOlJlcG9zaXRvcnkyMzI2MDQ3',
     name: 'git',
     full_name: 'dscho/git',
     private: false,
     owner: 
      {
        login: 'dscho',
        id: 127790,
        node_id: 'MDQ6VXNlcjEyNzc5MA==',
        avatar_url: 'https://avatars0.githubusercontent.com/u/127790?v=4',
        gravatar_id: '',
        url: 'https://api.github.com/users/dscho',
        html_url: 'https://github.com/dscho',
        followers_url: 'https://api.github.com/users/dscho/followers',
        following_url: 'https://api.github.com/users/dscho/following{/other_user}',
        gists_url: 'https://api.github.com/users/dscho/gists{/gist_id}',
        starred_url: 'https://api.github.com/users/dscho/starred{/owner}{/repo}',
        subscriptions_url: 'https://api.github.com/users/dscho/subscriptions',
        organizations_url: 'https://api.github.com/users/dscho/orgs',
        repos_url: 'https://api.github.com/users/dscho/repos',
        events_url: 'https://api.github.com/users/dscho/events{/privacy}',
        received_events_url: 'https://api.github.com/users/dscho/received_events',
        type: 'User',
        site_admin: false },
     html_url: 'https://github.com/dscho/git',
     description: 'Git Source Code Mirror',
     fork: true,
     url: 'https://api.github.com/repos/dscho/git',
     forks_url: 'https://api.github.com/repos/dscho/git/forks',
     keys_url: 'https://api.github.com/repos/dscho/git/keys{/key_id}',
     collaborators_url: 'https://api.github.com/repos/dscho/git/collaborators{/collaborator}',
     teams_url: 'https://api.github.com/repos/dscho/git/teams',
     hooks_url: 'https://api.github.com/repos/dscho/git/hooks',
     issue_events_url: 'https://api.github.com/repos/dscho/git/issues/events{/number}',
     events_url: 'https://api.github.com/repos/dscho/git/events',
     assignees_url: 'https://api.github.com/repos/dscho/git/assignees{/user}',
     branches_url: 'https://api.github.com/repos/dscho/git/branches{/branch}',
     tags_url: 'https://api.github.com/repos/dscho/git/tags',
     blobs_url: 'https://api.github.com/repos/dscho/git/git/blobs{/sha}',
     git_tags_url: 'https://api.github.com/repos/dscho/git/git/tags{/sha}',
     git_refs_url: 'https://api.github.com/repos/dscho/git/git/refs{/sha}',
     trees_url: 'https://api.github.com/repos/dscho/git/git/trees{/sha}',
     statuses_url: 'https://api.github.com/repos/dscho/git/statuses/{sha}',
     languages_url: 'https://api.github.com/repos/dscho/git/languages',
     stargazers_url: 'https://api.github.com/repos/dscho/git/stargazers',
     contributors_url: 'https://api.github.com/repos/dscho/git/contributors',
     subscribers_url: 'https://api.github.com/repos/dscho/git/subscribers',
     subscription_url: 'https://api.github.com/repos/dscho/git/subscription',
     commits_url: 'https://api.github.com/repos/dscho/git/commits{/sha}',
     git_commits_url: 'https://api.github.com/repos/dscho/git/git/commits{/sha}',
     comments_url: 'https://api.github.com/repos/dscho/git/comments{/number}',
     issue_comment_url: 'https://api.github.com/repos/dscho/git/issues/comments{/number}',
     contents_url: 'https://api.github.com/repos/dscho/git/contents/{+path}',
     compare_url: 'https://api.github.com/repos/dscho/git/compare/{base}...{head}',
     merges_url: 'https://api.github.com/repos/dscho/git/merges',
     archive_url: 'https://api.github.com/repos/dscho/git/{archive_format}{/ref}',
     downloads_url: 'https://api.github.com/repos/dscho/git/downloads',
     issues_url: 'https://api.github.com/repos/dscho/git/issues{/number}',
     pulls_url: 'https://api.github.com/repos/dscho/git/pulls{/number}',
     milestones_url: 'https://api.github.com/repos/dscho/git/milestones{/number}',
     notifications_url: 'https://api.github.com/repos/dscho/git/notifications{?since,all,participating}',
     labels_url: 'https://api.github.com/repos/dscho/git/labels{/name}',
     releases_url: 'https://api.github.com/repos/dscho/git/releases{/id}',
     deployments_url: 'https://api.github.com/repos/dscho/git/deployments',
     created_at: '2011-09-05T03:18:21Z',
     updated_at: '2019-09-16T20:37:39Z',
     pushed_at: '2019-11-18T22:02:18Z',
     git_url: 'git://github.com/dscho/git.git',
     ssh_url: 'git@github.com:dscho/git.git',
     clone_url: 'https://github.com/dscho/git.git',
     svn_url: 'https://github.com/dscho/git',
     homepage: 'http://git-scm.com',
     size: 192197,
     stargazers_count: 3,
     watchers_count: 3,
     language: 'C',
     has_issues: true,
     has_projects: true,
     has_downloads: true,
     has_wiki: true,
     has_pages: true,
     forks_count: 3,
     mirror_url: null,
     archived: false,
     disabled: false,
     open_issues_count: 1,
     license: 
      {
        key: 'other',
        name: 'Other',
        spdx_id: 'NOASSERTION',
        url: null,
        node_id: 'MDc6TGljZW5zZTA=' },
     forks: 3,
     open_issues: 1,
     watchers: 3,
     default_branch: 'master' },
  sender: 
   {
     login: 'dscho',
     id: 127790,
     node_id: 'MDQ6VXNlcjEyNzc5MA==',
     avatar_url: 'https://avatars0.githubusercontent.com/u/127790?v=4',
     gravatar_id: '',
     url: 'https://api.github.com/users/dscho',
     html_url: 'https://github.com/dscho',
     followers_url: 'https://api.github.com/users/dscho/followers',
     following_url: 'https://api.github.com/users/dscho/following{/other_user}',
     gists_url: 'https://api.github.com/users/dscho/gists{/gist_id}',
     starred_url: 'https://api.github.com/users/dscho/starred{/owner}{/repo}',
     subscriptions_url: 'https://api.github.com/users/dscho/subscriptions',
     organizations_url: 'https://api.github.com/users/dscho/orgs',
     repos_url: 'https://api.github.com/users/dscho/repos',
     events_url: 'https://api.github.com/users/dscho/events{/privacy}',
     received_events_url: 'https://api.github.com/users/dscho/received_events',
     type: 'User',
     site_admin: false
   },
  installation: 
   {
     id: 5187531,
     node_id: 'MDIzOkludGVncmF0aW9uSW5zdGFsbGF0aW9uNTE4NzUzMQ=='
   }
}
```
But just as I stare at this, I realize that I was stupid, and it is the `req.body.issue.number` I was looking for...

dscho:
@webstech @sgn could you have another look, please? In particular at the interdiff: https://github.com/gitgitgadget/gitgitgadget/compare/f8026c1a05c04760f35f95ddc216014e58bc62c7..374319ab56bce3729b70f360ecc357686c72dab6

script/misc-helper.ts
```diff
@@ -375,20 +375,29 @@ async function getCIHelper(): Promise<CIHelper> {
             await set({ appID: 46807, name: org});
         }
     } else if (command === "handle-pr-comment") {
-        if (commander.args.length !== 2) {
-            process.stderr.write(`${command}: Expected one comment ID\n`);
+        if (commander.args.length !== 2 && commander.args.length !== 3) {
+            process.stderr.write(`${command}: optionally takes  a ${
```
webstech: `handle-pr-push` says `need` and not `optionally`.  I assume the gitgitgadget pipeline will be updated after this is merged and the code can then be changed to require/need both args (ie cleaned up).  If my assumption is wrong then change line 391?

dscho: > `handle-pr-push` says `need` and not `optionally`

Whoops.

> I assume the gitgitgadget pipeline will be updated after this is merged and the code can then be changed to require/need both args (ie cleaned up).

Actually, I do not plan on updating the pipeline with ID 3 (i.e. the one servicing gitgitgadget/git).

Will change line 391, as you have suggested.

lib/gitgitgadget.ts
```diff
@@ -205,16 +206,25 @@ export class GitGitGadget {
...
+        ];
         if (additionalRef) {
             args.push(`+${additionalRef}:${additionalRef}`);
         }
+        if (repositoryOwner === "gitgitgadget") {
+            args.push(...prArgs);
+        } else {
+            prArgs.unshift("fetch", `https://github.com/${repositoryOwner}/git`,
+                           "--");
+            await git(prArgs, { workDir: this.workDir });
```
webstech: Sorry if this is a silly question.  How tightly in synch are gitgitgadget/git and git/git?  Just want to make sure `pr.baseCommit` will always be in the gitgitgadget repo for someone opening a PR on the git repo.

script/misc-helper.ts
webstech: I was reviewing individual commits when I first commented.  Looking at the merged set now and seeing the optional owner name (and error message wording issue) happens many times.  Would a commander option of --owner make this cleaner than an optional positional parameter?

btw, just noticed the use of `commander.gitGitWorkDir`.  Seems to ignore the `--work-dir` option (ie commander.workDir).

webstech:
I have now reviewed the commits and the set.  Think I may have commented on something earlier in a single commit that was updated in a later commit, so sorry about that.  I have done a basic scan of the project looking for anything else that may have been missed but you are much more familiar with it than I.  Just had a couple more comments above.

dscho:
> I have done a basic scan of the project looking for anything else that may have been missed but you are much more familiar with it than I.

@webstech thank you _so much_. The more eyes we have on these changes, the less likely bugs can hide.

lib/gitgitgadget.ts
dscho: > Sorry if this is a silly question. How tightly in synch are gitgitgadget/git and git/git?

I am glad you asked: gitgitgadget/git is kept up to date by two Azure Pipelines: [Synchronize gitster.git to GitGitGadget](https://dev.azure.com/gitgitgadget/git/_build?definitionId=8) and [Synchronize git.git to GitGitGadget](https://dev.azure.com/gitgitgadget/git/_build?definitionId=7).

The existence of the `git/git` -> `gitgitgadget/git` pipeline is probably obvious.

The `gitster/git` repository contains the individual branches for code contributions, and the first reason why I added the corresponding pipeline was that I got a bit sick and tired of `pu` failing all the time without any actionable information: it has been always unclear _which_ of those patch series that made it into `pu` was responsible for the test failures. Now with the individual branches being mirrored into `gitgitgadget/git`, and obviously triggering [CI builds](https://dev.azure.com/gitgitgadget/git/_build?definitionId=4), it is a lot easier to find the culprits.

Of course, from time to time the `pu` failures are caused by unfortunate interactions of separate patch series' changes, in which case the CI builds of the individual branches may very well succeed but `pu` still causes failures, which is a very useful piece of information in and of itself.

A secondary benefit of mirroring the `gitster/git` branches is that PRs at `gitgitgadget/git` can target more fine-grained base commits. I use this a lot, e.g. in my six "built-in add -i/-p" patch series which build on top of each other.

The pipeline that mirrors gitster/git into gitgitgadget/git (the pipeline names cannot contain slashes, that's why the `/` was replaced by a `.`) has two tasks:

1. Synchronize branch (the pipeline is triggered by all updates to `refs/heads/*`, polling every 180 seconds, i.e. every three minutes):
   ```bash
   case "$(Build.SourceBranch)" in
   refs/heads/*)
       refspec="+HEAD:$(Build.SourceBranch)"
       ;;
   refs/tags/*)
       refspec="$(Build.SourceBranch)"
       ;;
   *)
       echo "Cannot handle '$(Build.SourceBranch)'" >&2
       exit 1
       ;;
   esac

   git -c  http."https://github.com/gitgitgadget/git".extraheader="Authorization: Basic $(gitgitgadget.push.token.base64)" \
       push https://github.com/gitgitgadget/git "$refspec"
   ```
2. Synchronize tags (if necessary):
   ```bash
   die () {
       echo "$*" >&2
       exit 1
   }

   for d in git gitgitgadget
   do
       git ls-remote --tags https://github.com/$d/git | grep -v '\^{}$' | sort >tags.$d ||
       die "Could not enumerate tags in $d"
   done

   refspec="$(comm -23 tags.git tags.gitgitgadget | tr '\t' :)" ||
   die "Could figure out missing tags"

   if test -z "$refspec"
   then
       echo "##vso[task.complete result=Skipped]No tags to synchronize!"
   else
       git -c  http."https://github.com/gitgitgadget/git".extraheader="Authorization: Basic $(gitgitgadget.push.token.base64)" \
           push https://github.com/gitgitgadget/git $refspec
   fi
   ```

That second task is necessary because there is currently no way to trigger non-YAML Azure Pipelines from tag updates. Of course this means that new tags are only synchronized together with branch updates, but in practice that's okay because Junio always pushes out the tags with corresponding branches.

You will also notice that this task mirrors tags from the git/git repository, even if the pipeline purports to mirror gitster/git to gitgitgadget/git; This is a deliberate decision, to make sure that we only mirror the "official tags" from the authoritative repisotyr.

The pipeline that synchronizes git/git to gitgitgadget/git predates the other pipeline, and is probably made somewhat redundant by it. I still keep it around, and not only for sentimental reasons: it used to poll every 1800 seconds (i.e. every half hour), but the gitgitgadget builds recently got access to the webhooks, so it no longer polls (I *just* changed it, having noticed the behavior while writing this answer, and while at it, I also added a task that synchronizes the tags, which had been previously missing from that pipeline).

Side note: there is also the [Synchronize git-gui.git (branches only) to GitGitGadget](https://dev.azure.com/gitgitgadget/git/_build?definitionId=10) pipeline. Its purpose is obvious: to mirror the branches of Pratyush Yadav's repository (i.e. the current Git GUI maintainer's authoritative repository) into the `git-gui/*` namespace on gitgitgadget/git.

script/misc-helper.ts
dscho: > I was reviewing individual commits when I first commented. Looking at the merged set now and seeing the optional owner name (and error message wording issue) happens many times. Would a commander option of --owner make this cleaner than an optional positional parameter?

Having thought about your earlier comment about changing the pipeline to specify even `gitgitgadget`, I have come around to agreeing with your proposal.

As a consequence, I would like to keep this as a positional parameter, with the plan to make it mandatory after this PR was merged.

> btw, just noticed the use of `commander.gitGitWorkDir`. Seems to ignore the `--work-dir` option (ie commander.workDir).

You mean something like this?

```diff
diff --git a/script/misc-helper.ts b/script/misc-helper.ts
index 0446a9a..7af6591 100644
--- a/script/misc-helper.ts
+++ b/script/misc-helper.ts
@@ -24,7 +24,8 @@ if (commander.args.length === 0) {
 
 async function getGitGitWorkDir(): Promise<string> {
     if (!commander.gitGitWorkDir) {
-        commander.gitGitWorkDir = await gitConfig("gitgitgadget.workDir");
+        commander.gitGitWorkDir =
+            await gitConfig("gitgitgadget.workDir", commander.workDir || ".");
         if (!commander.gitGitWorkDir) {
             throw new Error(`Could not determine gitgitgadget.workDir`);
         }
```

I think you're right. I can sneak that in.

lib/gitgitgadget.ts
dscho: I copy-edited this into https://github.com/gitgitgadget/gitgitgadget/wiki/GitGitGadget's-Azure-Function-and-Azure-Pipelines.

dscho:
@webstech does [this interdiff](https://github.com/gitgitgadget/gitgitgadget/compare/1039c91dc2b3a9ea098163181b8401156cbff462..5abd8258cf679f5ca264fc60baa075d2726df3e7) look good to you?

lib/gitgitgadget.ts
webstech: Thanks for the detailed explanation and adding it to the wiki.  Always good to have the info in case you take vacation. 😀

webstech:
> @webstech does [this interdiff](https://github.com/gitgitgadget/gitgitgadget/compare/1039c91dc2b3a9ea098163181b8401156cbff462..5abd8258cf679f5ca264fc60baa075d2726df3e7) look good to you?

The old code relied on the git config setting and the new code overrides the git config setting.  Is that what is wanted?  Maybe not since this is the git work dir and not the gitgitgadget work dir.  Maybe defer this for now?  It could break things.  Maybe --work-dir is not needed and the config should be relied on.  What do you think?

May not matter but the default for `commander.workDir` is `"."` so the `|| "."` seems redundant.

dscho:
> > @webstech does [this interdiff](https://github.com/gitgitgadget/gitgitgadget/compare/1039c91dc2b3a9ea098163181b8401156cbff462..5abd8258cf679f5ca264fc60baa075d2726df3e7) look good to you?
> 
> The old code relied on the git config setting and the new code overrides the git config setting.

Oh, but all that was changed is that we now look in the specified workDir for that config setting. By default, it is still the same: `.`, i.e. the current directory.

> Is that what is wanted? Maybe not since this is the git work dir and not the gitgitgadget work dir.

I know, it _is_ confusing. There is `commander.workDir`, which is supposed to be the GitGitGadget worktree, and `commander.gitGitWorkDir`, which is the git.git worktree.

The config setting `gitgitgadget.workDir` is actually expected to be configured in the GitGitGadget worktree, and it points _to the git.git worktree_. Not the best naming on my part, I admit.

> Maybe defer this for now? It could break things. Maybe --work-dir is not needed and the config should be relied on. What do you think?

Sure, I should not have snuck it in. But I still think it is the right thing to do.

> May not matter but the default for `commander.workDir` is `"."` so the `|| "."` seems redundant.

True. And it also nudged me into looking what `commander.workDir` is actually used for. And the answer is: nothing.

I opened https://github.com/gitgitgadget/gitgitgadget/pull/156 to re-define what `--work-dir` means, and to make it actually do something in the first place ;-)

dscho:
@webstech I dropped that workdir patch again. Do you have any more concerns?

dscho:
FWIW I installed GitGitGadget-Git on git/git, and tested with a first patch series: https://lore.kernel.org/git/pull.670.git.git.1574433665.gitgitgadget@gmail.com/. So far, it seems to work fine.

webstech:
> > The old code relied on the git config setting and the new code overrides the git config setting.
> 
> Oh, but all that was changed is that we now look in the specified workDir for that config setting. By default, it is still the same: `.`, i.e. the current directory.

Sorry, my mistake confusing the workDir parameter  and the `gitConfig()` as a setter function when given a second parm.  Note to self: Drink ☕ before commenting.

> > Is that what is wanted? Maybe not since this is the git work dir and not the gitgitgadget work dir.
> 
> I know, it _is_ confusing. There is `commander.workDir`, which is supposed to be the GitGitGadget worktree, and `commander.gitGitWorkDir`, which is the git.git worktree.
>
> The config setting `gitgitgadget.workDir` is actually expected to be configured in the GitGitGadget worktree, and it points _to the git.git worktree_. Not the best naming on my part, I admit.

Agree on the confusing part when `--work-dir` was not used.

> I opened #156 to re-define what `--work-dir` means, and to make it actually do something in the first place ;-)

This ties in to issue #140 and the need to pass `--work-dir` to CIHelper (and thus GitGitGadget).

PR #156, opened 2019-11-22
Replace the confusing `--work-dir` option by two options (that actually work)

dscho:
`--work-dir` did not actually do anything.

Besides, we need _two_ options, one to override the current working directory as the one to be used as GitGitGadget worktree, and one to override the config setting in that worktree that points to the _git.git_ working directory.

This PR makes it so. It is based on #165.

### DIFF
--- a/lib/ci-helper.ts
+++ b/lib/ci-helper.ts
@@ -23,15 +23,15 @@ export class CIHelper {
     public readonly notes: GitNotes;
     protected readonly mail2commit: MailCommitMapping;
     protected readonly github: GitHubGlue;
-    protected readonly gggConfig: string;
+    protected readonly gggConfigDir: string;
     protected commit2mailNotes: GitNotes | undefined;
     protected testing: boolean;
     private gggNotesUpdated: boolean;
     private mail2CommitMapUpdated: boolean;
 
     public constructor(workDir?: string, skipUpdate?: boolean,
-                       gggConfig = ".") {
-        this.gggConfig = gggConfig;
+                       gggConfigDir = ".") {
+        this.gggConfigDir = gggConfigDir;
         this.workDir = workDir;
         this.notes = new GitNotes(workDir);
         this.gggNotesUpdated = !!skipUpdate;
@@ -527,7 +527,7 @@ export class CIHelper {
         };
 
         try {
-            const gitGitGadget = await GitGitGadget.get(this.gggConfig,
+            const gitGitGadget = await GitGitGadget.get(this.gggConfigDir,
                                                         this.workDir);
             if (!gitGitGadget.isUserAllowed(comment.author)) {
                 throw new Error(`User ${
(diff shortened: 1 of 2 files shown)


webstech:
I feel really bad about this.  I actually prefer the change you made [earlier](https://github.com/gitgitgadget/gitgitgadget/compare/1039c91dc2b3a9ea098163181b8401156cbff462..5abd8258cf679f5ca264fc60baa075d2726df3e7) which did not change the purpose of `--work-dir`.  That would position `getCIHelper()` to pass `commander.workDir` as part of the change for #140.

If you prefer this change, I can approve but I suspect the other change was your original intention.

dscho:
> I actually prefer the change you made [earlier](https://github.com/gitgitgadget/gitgitgadget/compare/1039c91dc2b3a9ea098163181b8401156cbff462..5abd8258cf679f5ca264fc60baa075d2726df3e7) which did not change the purpose of `--work-dir`.

From my analysis, it looks as if `--work-dir`  does not actually _do_ anything. Or do you find any effect of passing that option to `misc-helper`?

dscho:
> > I actually prefer the change you made [earlier](https://github.com/gitgitgadget/gitgitgadget/compare/1039c91dc2b3a9ea098163181b8401156cbff462..5abd8258cf679f5ca264fc60baa075d2726df3e7) which did not change the purpose of `--work-dir`.
> 
> From my analysis, it looks as if `--work-dir` does not actually _do_ anything. Or do you find any effect of passing that option to `misc-helper`?

Argh. I managed to forget about https://github.com/gitgitgadget/gitgitgadget/issues/140, where you suggested to make it do something: to set the GitGitGadget working directory.

So now I understand why we need this working directory: to be able to find `res/WELCOME.md`.

Given that we need to pass _two_ working directories to `CIHelper`, the option name `--work-dir` looks awfully ambiguous and confusing... I think I will want to clarify the purposes by dropping the `--work-dir` option, and by introducing `--gitgitgadget-work-dir` and `--git-git-work-dir` instead.

@webstech what do you think about this idea?

webstech:
> Argh. I managed to forget about #140, where you suggested to make it do something: to set the GitGitGadget working directory.
> 
> So now I understand why we need this working directory: to be able to find `res/WELCOME.md`.

The code is expecting to run in a GitGitGadget working directory.  I think that is okay.

> Given that we need to pass _two_ working directories to `CIHelper`, the option name `--work-dir` looks awfully ambiguous and confusing... I think I will want to clarify the purposes by dropping the `--work-dir` option, and by introducing `--gitgitgadget-work-dir` and `--git-git-work-dir` instead.

What is being passed as `--work-dir` is really a config location.  From that config, the git work dir is extracted and GitGitGadget obtains the SMTP info.  `CIHelper` passes `"."` as the config location to GitGitGadget at the moment and #140 is intended to change that.  So, the clarification is that `--work-dir` is a git config location. I don't think it is used for anything else.

dscho:
@webstech I rebased this on top of your PR that I just merged, reworked the patch substantially, and updated the PR's description. Could you have another look, please?

dscho:
> Can you explain the rational for specifying `commander.gitgitgadgetWorkDir || "."` on line 53 versus defaulting the value in commander? I see similar code in **git.ts** with `workDir?: string` that could be `workDir = "*"`.

The explanation is simple: I am not really used to concept of default values in Typescript ;-)

I just fixed the `gitgitgadgetGitDir` issue.

dscho:
@webstech thank you for your patient reviews!

PR #159, opened 2019-11-24
Detect stale `commitInGitGit` and force-update it

dscho:
I just noticed that https://github.com/gitgitgadget/git/pull/453 has not been updated to reflect that the patch was merged into `next`. Looks like a stale `commitInGitGit` for the head commit is still recorded, and therefore GitGitGadget is no longer able to service that PR correctly.

Let's try to detect such a situation by verifying that the commit in question is actually in `pu`.

### DIFF
--- a/lib/ci-helper.ts
+++ b/lib/ci-helper.ts
@@ -131,6 +131,7 @@ export class CIHelper {
         if (!this.gggNotesUpdated) {
             await git(["fetch", "https://github.com/gitgitgadget/git",
                        `+refs/notes/gitgitgadget:refs/notes/gitgitgadget`,
+                       `+refs/heads/maint:refs/remotes/upstream/maint`,
                        `+refs/heads/pu:refs/remotes/upstream/pu`],
                       { workDir: this.workDir });
             this.gggNotesUpdated = true;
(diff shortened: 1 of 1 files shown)


lib/ci-helper.ts
```diff
@@ -144,6 +144,12 @@ export class CIHelper {
             return false;
         }
 
+        const commitsInPu: Set<string> = new Set<string>(
+            (await git(["rev-list", "--no-merges",
+                        "^refs/remotes/upstream/maint~100",
```
webstech: Should `refs/heads/maint` be added to the git fetch at the start of this method?

lib/ci-helper.ts
```diff
@@ -160,7 +166,14 @@ export class CIHelper {
             }
             const meta = await this.getMailMetadata(messageID);
             if (!meta || meta.commitInGitGit !== undefined) {
-                continue;
+                if (!meta || commitsInPu.has(meta.commitInGitGit!)) {
+                    continue;
+                }
+                console.log(`Upstream commit ${meta.commitInGitGit} for ${
+                    info.headCommit} of ${
+                    info.pullRequestURL} no found longer in pu`);
```
webstech: Should probably be `no longer found in pu`.

webstech:
The commit message has a typo: `preformance`.  May as well correct when making updates.

lib/ci-helper.ts
dscho: Good point!

lib/ci-helper.ts
dscho: Oy. 🤦‍♂

Issue #158, opened 2019-11-23
Weird character encoding issue in a Gitgitgadget comment


phil-blain:
[This email](https://public-inbox.org/git/85CAC41A-A0EA-4E22-BC22-F57A59D76678@me.com/) was sent to the list and is mirrored in https://github.com/gitgitgadget/git/pull/430#issuecomment-547400942.
It seems that some characters are not showing correctly in the PR, even though the text on public-inbox looks OK.

Copy of the Gitgitgadget comment (notice the `=`, `=20`, `=C3=A0`, `=C3=A9`):

> [On the Git mailing list](https://public-inbox.org/git/85CAC41A-A0EA-4E22-BC22-F57A59D76678@me.com), Philippe Blain wrote ([reply to this](https://github.com/gitgitgadget/gitgitgadget/wiki/ReplyToThis)):
> 
> ```
> Hi Junio,
> 
> That indeed makes the trick. Thanks!=20
> Should I send a separate patch series with this patch ?  How would that =
> work ? "Signed-off by" me and "Based-on-patch-by" you ?
> 
> Philippe.
> 
> > Le 27 oct. 2019 =C3=A0 22:26, Junio C Hamano <gitster@pobox.com> a =
> =C3=A9crit :
> >=20
> > I am not sure if the compiler needs fixing in this case, but the
> > following may work it around.
> >=20
> > builtin/merge.c | 9 ++++++---
> > 1 file changed, 6 insertions(+), 3 deletions(-)
> >=20
> > diff --git a/builtin/merge.c b/builtin/merge.c
> > index e2ccbc44e2..0746f11df2 100644
> > --- a/builtin/merge.c
> > +++ b/builtin/merge.c
> ...
> ```

Copy of the text at public-inbox : 

```
Hi Junio,

That indeed makes the trick. Thanks! 
Should I send a separate patch series with this patch ?  How would that work ? "Signed-off by" me and "Based-on-patch-by" you ?

Philippe.

> Le 27 oct. 2019 à 22:26, Junio C Hamano <gitster@pobox.com> a écrit :
> 
> I am not sure if the compiler needs fixing in this case, but the
> following may work it around.
> 
> builtin/merge.c | 9 ++++++---
> 1 file changed, 6 insertions(+), 3 deletions(-)
> 
> diff --git a/builtin/merge.c b/builtin/merge.c
> index e2ccbc44e2..0746f11df2 100644
> --- a/builtin/merge.c
[...]

Issue #139, opened 2019-10-30
Handle emails in base64 encoding


dscho:
[This reply on the Git mailing list to a GitGitGadget PR](https://public-inbox.org/git/BN6PR21MB078689FFE9ED95A31998F41A91610@BN6PR21MB0786.namprd21.prod.outlook.com/) was encoded in base64:

```
Content-Transfer-Encoding: base64
```

and GitGitGadget did not decode it when mirroring it here: https://github.com/gitgitgadget/git/pull/424#discussion_r340301046

Obviously, we would want to decode such mails (and while at it, also handle quoted-printable encoded ones).

dscho:
Closed via https://github.com/gitgitgadget/gitgitgadget/pull/160

PR #160, opened 2019-11-27
Support base64 and quoted-printable emails

webstech:
Decode emails based on Content-Transfer-Encoding value when adding to the PR thread.

This is intended to address #139 and #158.

### DIFF
--- a/lib/public-inbox-helper.ts
+++ b/lib/public-inbox-helper.ts
@@ -3,7 +3,8 @@ import { git, revParse } from "./git";
 import { GitNotes } from "./git-notes";
 import { GitHubGlue } from "./github-glue";
 import { IMailMetadata } from "./mail-metadata";
-import { parseMBox, parseMBoxMessageIDAndReferences } from "./send-mail";
+import { IParsedMBox, parseMBox,
+    parseMBoxMessageIDAndReferences } from "./send-mail";
 
 const stateKey = "git@vger.kernel.org <-> GitGitGadget";
 const replyToThisURL =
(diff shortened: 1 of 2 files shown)


lib/public-inbox-helper.ts
```diff
@@ -116,7 +135,7 @@ export class PublicInboxGitHelper {
                      parsed.from.replace(/ *<.*>/, "") : "Somebody") +
                      ` wrote ([reply to this](${replyToThisURL})):\n\n`;
                 const comment = header +
-                    PublicInboxGitHelper.mbox2markdown(mbox);
+                    PublicInboxGitHelper.mbox2markdown(parsed);
```
dscho: Nice!

tests/send-mail.test.ts
```diff
@@ -56,3 +57,73 @@ test("parse mbox", async () => {
...
+Subject: [PATCH 0/3] My first Pull Request!
+Fcc: Sent
+Content-Type: text/plain; charset=UTF-8
+Content-Transfer-Encoding: BaSe64
+MIME-Version: 1.0
+To: reviewer@example.com
+Cc: Some Body <somebody@example.com>,
+ And Somebody Else <somebody@else.org>
+
+` + Buffer.from("Base 64 Data").toString("base64");
```
dscho: For my curiosity: why not embed it within the backticked part via `${Buffer.from("Base 64 Data")}`?

webstech: I am not always remembering to take full advantage of template literals.  I pushed up a change to do as you suggested and made it a little more DRY.

Issue #155, opened 2019-11-21
Switch from public-inbox.org to lore.kernel.org


dscho:
Seems that public-inbox.org will go the same way as gmane.org went, or almost: a new archive has been set up at lore.kernel.org. See https://lore.kernel.org/git/20191120195556.GA25189@dcvr/t/#u (which is identical to https://public-inbox.org/git/20191120195556.GA25189@dcvr/t/#u, and has the message by the public-inbox.org maintainer).

webstech:
In _lookup-commit.sh_, should this be changed to somewhere else?
`git clone --bare https://git-for-windows.visualstudio.com/git.public-inbox/_git/git.public-inbox "$PUBLIC_INBOX_DIR"`

webstech:
I have a [branch](https://github.com/webstech/gitgitgadget/compare/mail-archive) for this if you are interested.  Not sure how soon you want to get this done.  Pending the previous comment.  Also, need to run some tests and verify this:
`            this.state.latestRevision =
                "cf3590b3a1ce08a52b01142307b8fcc089acb6a6";
`

dscho:
>  Also, need to run some tests and verify this:
> ` this.state.latestRevision = "cf3590b3a1ce08a52b01142307b8fcc089acb6a6";`

I just tried to clone from lore.kernel.org, no dice. So it is definitely not a full replacement yet.

> I have a [branch](https://github.com/webstech/gitgitgadget/compare/mail-archive) for this if you are interested.

Nice!

About https://github.com/gitgitgadget/gitgitgadget/compare/master...webstech:mail-archive#diff-cf3021f43051c9b73cb47ee8437f78faL87-R92: I think we need to fall back to looking for `public-inbox.org` links if `lore.kernel.org` was not a match.

> Not sure how soon you want to get this done.

Well, seems that efforts are under way to move git/git to using the new URL, see https://lore.kernel.org/git/20191127200421.GA27456@dcvr/T/#u and https://lore.kernel.org/git/14167012a13125fe3c31af27afe68f84c9cdd2a2.1574882812.git.liu.denton@gmail.com/T/#u

So I'd say as soon as they have a working Git repository to clone from, we should switch.

dscho:
Seems like https://lore.kernel.org/git/0 is the URL for the repository. But the commit hashes do not match... I guess lore.kernel.org set up their own email account to receive those mails from...

```sh
$ git ls-remote https://lore.kernel.org/git/0 master
[...]

PR #164, opened 2019-12-05
Switch from public-inbox.org to lore.kernel.org

dscho:
Seems that public-inbox.org might go the same way as gmane.org went, or almost: a new archive has been set up at lore.kernel.org. See https://lore.kernel.org/git/20191120195556.GA25189@dcvr/t/#u (which is identical to https://public-inbox.org/git/20191120195556.GA25189@dcvr/t/#u, and has the message by the public-inbox.org maintainer).

A mirror was already set up at https://dev.azure.com/gitgitgadget/git/_git/lore-git?_a=history, and an Azure Pipeline to keep that mirror up to date: https://dev.azure.com/gitgitgadget/git/_build?definitionId=14&_a=summary.

These patches will (hopefully) take care of the rest.

This addresses https://github.com/gitgitgadget/gitgitgadget/issues/155

### DIFF
--- a/lib/mail-archive-helper.ts
+++ b/lib/mail-archive-helper.ts
@@ -14,13 +14,13 @@ export interface IGitMailingListMirrorState {
     latestRevision?: string;
 }
 
-export class PublicInboxGitHelper {
-    public static async get(gggNotes: GitNotes, publicInboxGitDir: string,
+export class MailArchiveGitHelper {
+    public static async get(gggNotes: GitNotes, mailArchiveGitDir: string,
                             githubGlue: GitHubGlue):
-        Promise<PublicInboxGitHelper> {
+        Promise<MailArchiveGitHelper> {
         const state: IGitMailingListMirrorState =
             await gggNotes.get<IGitMailingListMirrorState>(stateKey) || {};
-        return new PublicInboxGitHelper(gggNotes, publicInboxGitDir, githubGlue,
+        return new MailArchiveGitHelper(gggNotes, mailArchiveGitDir, githubGlue,
                                         state);
     }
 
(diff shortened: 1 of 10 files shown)


lib/mail-archive-helper.ts
```diff
@@ -199,7 +199,43 @@ export class MailArchiveGitHelper {
...
+                commit = await git(["log", "-1", "--format=%H", `-S${match[1]}`,
+                                    commit, "--"],
+                                   { workDir: this.mailArchiveGitDir});
+                if (!commit) {
+                    throw new Error(`Could not find ${this.state.latestRevision
+                                }'s equivalent in ${this.mailArchiveGitDir}`);
+                }
+                commitDiff = await git(["show", commit, "--"],
+                                       {workDir: this.mailArchiveGitDir});
+                if (commitDiff.indexOf(`\n+${match[1]}\n`) >= 0) {
```
webstech: I am not advocating for the removal but is the `+` really needed for the regex or was it just left in when the replacement mode of updates was found?

webstech:
Sorry about not updating send-mail.test.ts.  Don't think it was in master at the time but I should have remembered it.

lib/mail-archive-helper.ts
dscho: It is actually necessary: in the public-inbox.org repository, every mail was added as its own file, but in the lore.kernel.org one, every new mail replaces the previously-new mail in the file `m`. Therefore, if we look only for the `Message-ID: <id>`, we may very well find a commit that _removes_ those contents, i.e. a commit that corresponds to the child of the commit we are looking for.

dscho:
> Sorry about not updating send-mail.test.ts. Don't think it was in master at the time but I should have remembered it.

Oh, don't worry even 1ns about it. This is the reason why PR builds exist: to relieve us from that particular cognitive load.

lib/mail-archive-helper.ts
webstech: So the `\n+` should be `\n\+`, to actually include the literal `+` or the match at line 216 could have included the `+` - ie `\n(\+Message-ID...)`.

dscho: Oy. Oy, oy, oy. I got it right in line 216 and not here.

Good eyes! 👓

dscho: Fixed ;-)

webstech: Actually bad eyes. Sorry about this.  The second check is using `indexOf` so it is not a regex.  I was too focused on the search string and missed the method being used.  Realized my error while you were making the update.

dscho: Oh. Okay, I reverted this. I clearly need more sleep.

PR #165, opened 2019-12-09
Support alternate config dir for GitGitGadget

webstech:
Positioning change for tests which need to provide a separate
config to GitGitGadget.get().  Currently, the current directory
is passed.  This will remain the default but user testing should
not need to update the local repo config.  This change will allow
tests to provide an alternate location of the config which
GitGitGadget uses to obtain SMTP information.

Issue #140 and PR #156 will be able to use this change.  A work in progress at [webstech/ci-helper-test](https://github.com/webstech/gitgitgadget/commit/13c01a817577f1107e4b16d7320c0ee2371315bc) is a sample use of this change.


dscho:
Thank you!

Issue #161, opened 2019-11-27
Multi line quotes in the cover letter not working properly


prati0100:
Hi,

In [this](https://github.com/gitgitgadget/git/pull/436) thread, the author replies to my review in the cover letter (the PR description). He edits it after every round of review.

But, when the cover letter is sent as email, that quoting gets messed up and the result looks like [this](https://lore.kernel.org/git/pull.436.v5.git.1574627876.gitgitgadget@gmail.com/).

The first line of a multi-line quoted text is prefixed with '> ', but the rest of the lines aren't. This makes it very difficult to quickly differentiate between the quoted text and reply text.

webstech:
GitHub appears to keep the PR comments in markdown [format.](https://api.github.com/repos/gitgitgadget/git/pulls/436).  GitGitGadget converts the markdown to html and then back to text.  The multi line quotes in markdown become blockquotes in html, with multiple quoted lines converted to a single line.  The conversion from html back to text does not seem to be wrapping as expected and maintaining the quoting.

Some initial feedback while research continues.

webstech:
The fromString function used to convert from html to text does _lazy_ quoting where the `> ` is only on the start of the quoted block.  This relies on mail formatting to support or not support it.

@dscho I have a local override of the blockquote handler if it is of interest.  The [change](https://github.com/gitgitgadget/gitgitgadget/compare/master...webstech:block-quote) does not produce an exact match due to multiple conversions but is pretty close (blank lines between quote levels).  Tested with some actual comments obtained from the api at https://api.github.com/repos/gitgitgadget/gitgitgadget/issues/_nnn_/comments.  Quoting may not happen a lot in a PR (in fact, it is gone from the thread that prompted this issue) but it could be handled better.

dscho:
>  I have a local override of the blockquote handler if it is of interest.

@webstech yes!

> The [change](https://github.com/gitgitgadget/gitgitgadget/compare/master...webstech:block-quote) does not produce an exact match due to multiple conversions but is pretty close (blank lines between quote levels).

I am always a fan of "better than before" 😄

dscho:
```typescript
            const quoteOptions = options;
            quoteOptions.wordwrap -= 2;
```

I think this does not do what you think it does... because `quoteOptions = options` does not do a deep copy, it only copies the reference. See for example:

```sh
$ node -e 'var x = {a: 1, b: 2}; var y = x; y.b = 4; console.log(x); console.log(y);'
[...]

PR #168, opened 2019-12-11
Improve block quote formatting

webstech:
The fromString function from "html-to-text" does "lazy"
quoting with only a leading > for the quote.  This makes
cover letters that include quoting hard to read.

This change adds a blockquote formatter that will put `> ` at
the start of each line of the quote.  Quoting inside of quotes
will only add a `>`.  Not perfect but better than relying on
mail formatters to handle lazy quoting.

Signed-off-by: Chris. Webster <chris@webstech.net>

### DIFF
--- a/lib/markdown-renderer.ts
+++ b/lib/markdown-renderer.ts
@@ -17,6 +17,16 @@ class MyOptions implements HtmlToTextOptions {
                 .replace(/./g, "=");
             return `${heading}\n${underline}\n\n`;
         },
+        blockquote(elem: any, fn: any, options: any): string {
+            const indentOptions = Object.assign({}, options);
+            // decrease word wrap, but only to a minimum of 20 columns/line
+            indentOptions.wordwrap = Math.max(20, options.wordwrap - 2);
+
+            const block: string = fn(elem.children, indentOptions);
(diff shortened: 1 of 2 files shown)


lib/markdown-renderer.ts
```diff
@@ -17,6 +17,16 @@ class MyOptions implements HtmlToTextOptions {
...
         },
+        blockquote(elem: any, fn: any, options: any): string {
+            const indentOptions = Object.assign({}, options);
+            // decrease word wrap, but only to a minimum of 20 columns/line
+            indentOptions.wordwrap = Math.max(20, options.wordwrap - 2);
+
+            const block: string = fn(elem.children, indentOptions);
+            return block.replace(/^>/mg, ">>") // add to quote
+                .replace(/^(?!>|$)/mg, "> ")   // new quote
+                .replace(/(^|\n)(\n)/g, "$1>$2"); // quote empty lines
```
dscho: If we add `(?!$)` to this regex, do we get rid of the ugly trailing quotes?

webstech: Yes it does.  Done!

webstech:
This is a fix for issue #161.

dscho:
Very nice, thank you!

Issue #171, opened 2019-12-16
Using multiple CC's in one line can fail


SyntevoAlex:
Before https://github.com/gitgitgadget/gitgitgadget/pull/163 the only way of specifying multiple CC's was to have them on the same line without trailing newline.

Some developers seem to have found that approach, for example:
https://github.com/gitgitgadget/git/pull/450
https://github.com/gitgitgadget/git/pull/428
https://github.com/gitgitgadget/git/pull/415

However, doing so can "eat" the first CC, I myself ran into this problem in https://github.com/gitgitgadget/git/pull/490

The problem is demonstrated by this test:

	test("multiple CC's in one line", () => {
		const multipleCCs = "Some Dev <dev@example.com>, Some Other Dev <other@example.com>, Another Dev <another@example.com>";

		const encoded = PatchSeries.encodeSender(multipleCCs);

		const expected = "\"Some Dev\" <dev@example.com>, Some Other Dev <other@example.com>, Another Dev <another@example.com>"
		expect(encoded).toEqual(expected);
	});

The results are (formatted for ease of reading)

	"Some Dev <dev@example.com>, Some Other Dev" <other@example.com>,
	Another Dev <another@example.com>

There, the first mail was quoted and therefore ignored. Also, it seems that neither `GitGitGadget` nor `git format-patch` expect multiple items in one go. It "worked" by coincidence.

One solution is to split multi-email `Cc:` footer in `parsePullRequestBody()`, so that the rest of GitGitGadget works with expected single-mail items. Splitting could be non-trivial, though, because in worst case there could be some name with a comma like `"Someone, the team leader" <teamlead@example.com>`.

An alternative solution is to forbid using multiple emails in one `Cc:`, simply detecting multiple `@` should be good enough.

dscho:
Addressed via https://github.com/gitgitgadget/gitgitgadget/commit/bb46bfc6dbab366a527491030b6611cb1cbb0ae8

PR #172, opened 2019-12-16
Handle multiple entries in the same Cc: line again

dscho:
This got broken somewhere along the lines, most likely when we started to use `encodeWords()` to quote names when necessary.

This addresses https://github.com/gitgitgadget/gitgitgadget/issues/171

cc @SyntevoAlex

### DIFF
--- a/lib/patch-series.ts
+++ b/lib/patch-series.ts
@@ -1,3 +1,4 @@
+import addressparser = require("nodemailer/lib/addressparser");
 import { encodeWords } from "nodemailer/lib/mime-funcs";
 import {
     commitExists, git, gitCommandExists, gitConfig, revParse,
@@ -286,7 +287,13 @@ export class PatchSeries {
                             basedOn = match2[2];
                             break;
                         case "cc:":
-                            cc.push(match2[2]);
+                            addressparser(match2[2]).forEach((e) => {
+                                if (e.name) {
+                                    cc.push(`${e.name} <${e.address}>`);
+                                } else {
+                                    cc.push(e.address);
+                                }
+                            });
                             break;
                         default:
                             footer.push(line);
--- a/tests/patch-series.test.ts
+++ b/tests/patch-series.test.ts
@@ -423,7 +423,9 @@ Fetch-It-Via: git fetch ${repoUrl} my-series-v1
                 "some description goes here",
                 "",
                 "Cc: Some Contributor <contributor@example.com>",
-                "CC: Capital Letters <shout@out.loud>",
+                "CC: Capital Letters <shout@out.loud>, Hello <hello@wor.ld>"
+                + ", without@any.explicit.name"
+                + "; Semi Cologne <semi@col.on>",
                 "Cc: Git Maintainer <maintainer@gmail.com>",
                 "This is PR template",
                 "Please read our guide to continue",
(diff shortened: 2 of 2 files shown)


lib/patch-series.ts
```diff
@@ -286,7 +287,14 @@ export class PatchSeries {
                             basedOn = match2[2];
                             break;
                         case "cc:":
-                            cc.push(match2[2]);
+                            addressparser(match2[2]).forEach((e) => {
+                                if (e.name) {
+                                    cc.push(encodeWords(e.name)
```
SyntevoAlex: Is it expected that `generateMBox()` also calls `encodeWords` ?

dscho: Oh, I guess not... Good point. Will fix.

SyntevoAlex: Thanks, looks reasonable to me now (not that I have any experience to be sure, though)

dscho:
No idea why the check does not update, but the build is green: https://dev.azure.com/gitgitgadget/git/_build/results?buildId=24119&view=results

dscho:
@webstech could I ask you to review this PR?

lib/patch-series.ts
```diff
@@ -286,7 +287,13 @@ export class PatchSeries {
                             basedOn = match2[2];
                             break;
                         case "cc:":
-                            cc.push(match2[2]);
+                            addressparser(match2[2]).forEach((e) => {
+                                if (e.name) {
+                                    cc.push(e.name + " <" + e.address + ">");
```
webstech: Maybe use a template here instead of `+`? 
```
`${e.name} <${e.address}>`
```
Or remove the if with a conditional
```
cc.push(`${e.name ? `${e.name} <${e.address}>` : `${e.address}`}`);
```

tests/patch-series.test.ts
```diff
@@ -423,7 +423,8 @@ Fetch-It-Via: git fetch ${repoUrl} my-series-v1
                 "some description goes here",
                 "",
                 "Cc: Some Contributor <contributor@example.com>",
-                "CC: Capital Letters <shout@out.loud>",
+                "CC: Capital Letters <shout@out.loud>, Hello <hello@wor.ld>"
+                + ", without@any.explicit.name",
                 "Cc: Git Maintainer <maintainer@gmail.com>",
```
webstech: Add a test with a semicolon delimiter?  It is supported by addressparser() so maybe we want to know if support is removed.  Some people may be used to using semicolon.

lib/patch-series.ts
dscho: Good idea about the simplification. Personally, I like ternaries, but in this instance it seems a bit harder to read, so I refrained from doing it.

tests/patch-series.test.ts
dscho: Good idea! I added the test (the code did not need any change, as you probably expected: `addressparser()` does all the hard work for us here.

For future reference: if we ever introduce that mapping from GitHub handles to email addresses, we would want to introduce that handling also in the code in the `patch-series.ts` file that was touched above, allowing e.g. `Cc: @dscho` and/or `Cc: dscho`.

SyntevoAlex: Spaces vs tabs in line 428?

lib/patch-series.ts
webstech: My example didn't help when it should have been `: e.address` but no problem.  It is already well indented.

tests/patch-series.test.ts
dscho: Oops, sorry. Force-pushed a fix.

dscho:
Thank you for the reviews!

SyntevoAlex:
Thanks for the fix! I wouldn't have found that parser library myself

Issue #82, opened 2019-03-15
Prevent stupidly large patch series, as well as PRs with merge commits


dscho:
It's probably safe to say that something like three dozen patches in one go are too much for any reviewer, even for the most exciting contributions.

So let's disallow it, also notifying the user of that fact when they push such a thing.

While at it, also holler about merge commits in the PR, as that cannot be reflected by `git format-patch`.

webstech:
Should this be a github action for a PR/push update?  If not, why not?  Just looking for preferred direction.

dscho:
I would _love_ for this to happen, but GitHub Actions need to be added to the target branch, and we do not control `master`.

I think it would be much easier to implement a simple check, such as generating the `git rev-list --parents ${baseCommit}..${tipCommit}`, then counting the number of lines (I think we should disallow patch series larger than 30 commits by default, probably with some way to exempt a given PR), and then verifying that there are no lines with two spaces in them (that would be a merge commit).

webstech:
So webhooks would be added for pull_request and push events?  The azure-function/index.js would handle these and trigger a next step.  misc-helper.ts and ci-helper.ts would provide support for handling the events.  This would be the same for issue #120.  I am not clear on where misc-helper is identified to be run.  Is that in azure setup?  Let me know if this is the design you had in mind.

dscho:
> So webhooks would be added for pull_request and push events?

That is already the case...

>  The azure-function/index.js would handle these and trigger a next step.

The `issue_comment` events are handled by the Azure Function, triggering the [Azure Pipeline](https://dev.azure.com/gitgitgadget/git/_build?definitionId=3).

But the `push` event (and also the `pull_request` event) is not handled by the Azure Function. It is handled via the Azure Pipeline being configured as "Pull request validation". That also allows it to be shown in the Checks part of the PR.

You can see the difference on https://dev.azure.com/gitgitgadget/git/_build?definitionId=3: the `push`-triggered builds are labeled as "Pull request build", and the `issue_comment` ones as "Manual build" (because they are triggered using a Personal Access Token ("PAT") that I created).

> misc-helper.ts and ci-helper.ts would provide support for handling the events.

Right. `misc-helper.ts` is called with `handle-pr-comment` or with `handle-pr-push`, respectively. (This already works.)

webstech:
[...]

PR #178, opened 2020-01-02
Limit the number of commits on a pull request.

webstech:
A large number of patches is too much for reviewers.

A comment is added to the PR and the pipeline build fails for
pull request updates.

This implements part of #82.

The limit is arbitrary and easily changed.  Support for overrides could be added but really?

Suggestions for better messaging are welcome.

Tests are included with the PR.

### DIFF
--- a/lib/ci-helper.ts
+++ b/lib/ci-helper.ts
@@ -10,6 +10,7 @@ import { IMailMetadata } from "./mail-metadata";
 import { IPatchSeriesMetadata } from "./patch-series-metadata";
 
 const readFile = util.promisify(fs.readFile);
+type CommentFunction = (comment: string) => Promise<void>;
 
 /*
  * This class offers functions to support the operations we want to perform from
@@ -554,15 +555,19 @@ export class CIHelper {
 
                 const userInfo = await this.getUserInfo(comment.author);
 
-                const extraComment = userInfo.email === null ?
-                    ( `\n\nWARNING: ${comment.author} has no public email` +
-                    " address set on GitHub" ) : "";
+                const commitOkay = await this.checkCommits(pr, addComment);
 
-                const coverMid =
-                    await gitGitGadget.submit(pr, userInfo);
-                await addComment(`Submitted as [${
-                    coverMid}](https://lore.kernel.org/git/${coverMid})${
-                        extraComment}`);
+                if (commitOkay) {
+                    const extraComment = userInfo.email === null ?
+                        ( `\n\nWARNING: ${comment.author} has no public email` +
+                        " address set on GitHub" ) : "";
+
+                    const coverMid = await gitGitGadget.submit(pr, userInfo);
+                    await addComment(`Submitted as [${
+                        coverMid}](https://lore.kernel.org/git/${coverMid})${
+                            extraComment}`);
+                }
+
             } else if (command === "/preview") {
                 if (argument && argument !== "") {
                     throw new Error(`/preview does not accept arguments ('${
@@ -578,9 +583,13 @@ export class CIHelper {
                         comment.author}`);
                 }
 
-                const coverMid =
-                    await gitGitGadget.preview(pr, userInfo);
-                await addComment(`Preview email sent as ${coverMid}`);
+                const commitOkay = await this.checkCommits(pr, addComment);
+
+                if (commitOkay) {
+                    const coverMid = await gitGitGadget.preview(pr, userInfo);
+                    await addComment(`Preview email sent as ${coverMid}`);
+                }
+
             } else if (command === "/allow") {
                 const accountName = argument || await getPRAuthor();
                 let extraComment = "";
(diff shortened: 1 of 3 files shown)


lib/ci-helper.ts
```diff
@@ -554,6 +554,12 @@ export class CIHelper {
 
                 const userInfo = await this.getUserInfo(comment.author);
 
+                const commitProblems = await this.checkCommits(pr);
+                if (commitProblems) {
+                    this.github.addPRComment(pullRequestURL, commitProblems);
+                    throw new Error(`Max commits exceeded.`);
```
dscho: Maybe the error should repeat the `commitProblems` rather than assuming that the problem was "too many commits"?

lib/ci-helper.ts
```diff
@@ -578,6 +584,12 @@ export class CIHelper {
                         comment.author}`);
                 }
 
+                const commitProblems = await this.checkCommits(pr);
+                if (commitProblems) {
+                    this.github.addPRComment(pullRequestURL, commitProblems);
+                    throw new Error(`Max commits exceeded.`);
```
dscho: Same here: maybe `throw new Error(commitProblems)` instead?

lib/ci-helper.ts
```diff
@@ -626,15 +638,43 @@ export class CIHelper {
...
     }
 
+    public async checkCommits(pr: IPullRequestInfo):
+        Promise<string> {
+        let comment: string = "";
+        const maxCommits = 30;
+        if (pr.commits && pr.commits > maxCommits) {
+            comment = `The pull request has ${pr.commits
+                } commits.  The max allowed is ${maxCommits
+                }.  Please squash some commits and 'git push --force'`;
```
dscho: I would rather suggest to split the patch series.

Squashing commits would probably result in multiple unrelated changes within the same commits (which is something that the reviewers will point out as a problem later on).

lib/ci-helper.ts
```diff
@@ -626,15 +638,43 @@ export class CIHelper {
...
-                `https://github.com/${repositoryOwner}/git/pull/${prNumber}`;
-            const welcome = (await readFile("res/WELCOME.md")).toString()
-                .replace(/\${username}/g, pr.author);
-            this.github.addPRComment(pullRequestURL, welcome);
+        const pullRequestURL = `https://github.com/${repositoryOwner
+                                }/git/pull/${prNumber}`;
+
+        try {
+            const gitGitGadget = await GitGitGadget.get(this.gggConfigDir,
+                                                        this.workDir);
```
dscho: Sneaky ;-) It's a good change to replace `"."` here, of course.

lib/ci-helper.ts
```diff
@@ -626,15 +638,43 @@ export class CIHelper {
...
+            const commitProblems = await this.checkCommits(pr);
+
+            if (commitProblems) {
+                this.github.addPRComment(pullRequestURL, commitProblems);
+                throw new Error(`Max commits exceeded.`);
+            }
+        } catch (e) {
+            const body = e.toString();
+            console.log(`Adding comment to ${pullRequestURL}:\n${body}`);
+            await this.github.addPRComment(pullRequestURL, body);
```
dscho: Should we really throw an error here? It would appear to me as if this would have been enough:

```diff
+    if (commitProblems) {
+        this.github.addPRComment(pullRequestURL, commitProblems);
+    }
```

tests/ci-helper.test.ts
```diff
@@ -654,3 +654,90 @@ test("handle comment preview email success", async () => {
...
+    prinfo.author = "starfish";
+    comment.author = "starfish";
+    user.login = "starfish";
+    ci.setGHgetPRInfo(prinfo);
+    ci.setGHgetPRComment(comment);
+    ci.setGHgetGitHubUserInfo(user);
+
+    await ci.handlePush("gitgitgadget", 433865360);
+    expect(ci.addPRComment.mock.calls[0][1]).toMatch(/Welcome/);
+    expect(ci.addPRComment.mock.calls[2][1]).toMatch(/Max commits exceeded/);
```
dscho: Maybe we should test more concretely that it talks about 40 commits being too many?

lib/ci-helper.ts
webstech: I am open to any suggested verbiage.

_Please split the patch series into multiple pull requests.  Also consider squashing related commits._

Key here is to provide direction and use wording that is clear to everyone.  Would it be better to have a wiki page to discuss limiting the number of commits?  The message could then just point to the wiki page (the wiki page could be as detailed as needed or I am just punting responsibility).  Hmm, thinking a submission guidelines page may be useful so the linting errors can also point to it.  What do you think?  Don't want the rejection messages to be too verbose so pointing to (soon to be) existing doc may be a good way to go.

lib/ci-helper.ts
webstech: This (the catch) is equivalent to the `addComment` function used in the big catch of  `handleComment`.  The change you noticed in your previous comment was the real reason for this change.

I agree throwing an error at 672 is not the best way to go since that is really intended for the job log and not as a PR comment.  This applies to all callers of checkCommits.

lib/ci-helper.ts
webstech: Missed it last time around.

lib/ci-helper.ts
webstech: This is the first pass in the linting and handling other errors was considered but not present.  Need to make sure we agree on reporting.  I was heading in the direction of one comment per commit which has all the suggestions for improving the commit.  This could cause multiple comments but they can be addressed individually.  One comment per suggestion seems excessive since the comments will show up each time a change is pushed.

So, the second pass in linting would pass a call back to checkCommits for use during the various tests.  I like the idea of passing an object to the call back with a short log message and long comment.

lib/ci-helper.ts
webstech: I know there is doc on git patch submission but a wiki page would be more specific on the gitgitgadget guidelines/linting.

tests/ci-helper.test.ts
webstech: This comment is being removed (becoming a log message).  Will change the test to validate the commit count (while avoiding the internal max).

lib/ci-helper.ts
dscho: > Please split the patch series into multiple pull requests. Also consider squashing related commits.

I like that.

The wiki page idea is also good, maybe have both the message _and_ a link to the wiki page?

tests/ci-helper.test.ts
dscho: :+1:

lib/ci-helper.ts
webstech: Added the callback and changed `checkCommits` to return true if no errors.  No error is thrown for comment problems.  The log will include the PR comment data.

lib/ci-helper.ts
webstech: Removed throw.

lib/ci-helper.ts
webstech: There is a wiki page on git-for-windows/git regarding commits.  Should it be expanded regarding linting or should the gitgitgadget wiki have this info?

lib/ci-helper.ts
webstech: try/catch removed - let misc-helper catch the errors.  If checkCommit returns false an error is thrown to make the CI check fail.  That will get the user's attention.

lib/github-glue.ts
```diff
@@ -11,6 +11,7 @@ export interface IPullRequestInfo {
     baseCommit: string;
     baseOwner: string;
     baseRepo: string;
+    commits?: number;
```
dscho: Oh crap, I forgot about that in #181... I called the same field `patchCount`... Do you have a strong opinion on the name (I guess the GitHub response data calls it `commits`...)? I will be happy to change my PR.

lib/ci-helper.ts
dscho: Quite honestly, I'd rather move the wiki page from git-for-windows to gitgitgadget, leaving a link to the latter in place of the former.

lib/github-glue.ts
dscho: On second thought, I don't think I can rely on this field to be set, and it is also more accurate to call it "patch count" ("commits" would also include merge commits, after all), so I'll just keep my `patchCount`...

webstech: Yes, I followed the GitHub api field name.  The field is not set for `GitHubGlue::getOpenPRs()` responses but it is for `getPRInfo()`.  Yes it does include merges but those will be rejected at the front end soon.

dscho:
It works: https://github.com/git/git/pull/681#issuecomment-572220547 😃

PR #180, opened 2020-01-06
Remove obsolete files

dscho:
I just noticed some left-over files we no longer need.


dscho:
> Not sure if other shell scripts are used but script/monitor-upstream-commits.ts is probably not.

You're absolutely correct! I removed that one, too.

While one Unix shell script of the three we carry in our tree is not used regularly (`search-commit-on-mail-archive.sh` is intended to try to find mails corresponding to a give commit when the automatic mail<->commit mapping failed), the other two (`update-mail-to-commit-notes.sh` and `lookup-commit.sh`) are used regularly in the [Update GitGitGadget's commit to mail notes](https://dev.azure.com/gitgitgadget/git/_build/index?definitionId=9&_a=completed) Pipeline. I would actually like to even keep the rarely-used one.

PR #185, opened 2020-01-08
Stop Cc:ing the Git maintainer

dscho:
In https://lore.kernel.org/git/xmqqv9pn5hgl.fsf@gitster-ct.c.googlers.com/, we got yet another piece of disapproval of GitGitGadget. This PR intends to address the feasible part of the wishes offered in that mail thread.

Let's hope that this stream of complaints ends soon.


dscho:
> On a bright note, this is a sign gitgitgadget is growing and useful.

Yeah, that's the bright side. I should not forget that, shouldn't I?

> You are hearing voices? 😮

Yes! :-) Right now, I hear a tenor voice humming. It's my voice.

Issue #120, opened 2019-09-26
Rudimentary commit message linting


dscho:
To make it easier for new contributors, we should probably have our own "linter" for commit messages, i.e. validate that at least some of the conventions are met:

- the first line should not exceed 76 characters
- the first line should be followed by an empty line
- if the first line starts with a prefix (e.g. `tests:`), it should continue in lower-case
- the last line should be a Signed-off-by: line (our DCO check does this already, but it managed to baffle one long-time contributor already because its error message was not clear enough: the Signed-off-by: line was indented)
- the body of the commit message should not be empty.
- the body of the commit message should not consist of a hyperlink, without any other explanation.

There is also the convention that the commit message should talk in the imperative tone (as opposed to the past tense, as many other projects prefer). I am not sure that there is an easy way to test for that.

If any of these conventions are not met, GitGitGadget should talk about that, in a way that helps the contributors (e.g. suggest amending the commit, or using an interactive rebase if it is not the latest commit, force-push to update the PR, etc).

webstech:
There are some github apps for commit linting but they do not appear to match these requirements.  Not sure if they can be extended to include these requirements but a config file should be able to control what is needed for a generic linter.  Just wondering if this issue is better addressed with a github action responding to a pull request or a push to a pull request versus a specific PR comment.  There is also the question of whether it would not be a good commit hook for git in general.

dscho:
> Just wondering if this issue is better addressed with a github action responding to a pull request or a push to a pull request versus a specific PR comment.

I would be in favor, but that GitHub Action needs to be in `master`, which is controlled by the Git maintainer, not by GitGitGadget.

> There is also the question of whether it would not be a good commit hook for git in general.

Sure. But there is an entire ecosystem of `npm` packages out there that Unix shell scripts typically cannot access (read: the common Git hooks), but GitGitGadget can.

webstech:
I have an initial POC/WIP at https://github.com/webstech/gitgitgadget/commit/888c4f53429f6f5801c89f6e7319e9f24a380713.  The message linting is done in a separate class with methods to perform one or more checks based on complexity.  Just want to make sure you agree on the basic design before proceeding further.  As always, suggestions are welcome.

dscho:
> I have an initial POC/WIP at [webstech@888c4f5](https://github.com/webstech/gitgitgadget/commit/888c4f53429f6f5801c89f6e7319e9f24a380713).

Very nice! I left a couple of comments there.

[...]

PR #186, opened 2020-01-12
Commit Message Linting Phase 1

webstech:
The commits in a pull request are obtained from GitHub and run
through some basic validation.  The check job will fail if the messages
do not pass the tests.  The intention is to pre-empt problems that the
git maintainers will raise.

Add new file commit-lint.ts.  It contains several commit message
validation checks.  These should grow over tine.

Add new file commit-lint.test.ts which contains unit tests for
the Linting checks.

This is the start of resolving issue #120.

### DIFF
--- a/lib/ci-helper.ts
+++ b/lib/ci-helper.ts
@@ -1,5 +1,6 @@
 import * as fs from "fs";
 import * as util from "util";
+import { ILintError, LintCommit } from "./commit-lint";
 import { commitExists, git } from "./git";
 import { GitNotes } from "./git-notes";
 import { GitGitGadget, IGitGitGadgetOptions } from "./gitgitgadget";
@@ -685,6 +686,24 @@ export class CIHelper {
                     merges.join("\n    ")
                     }\n\nPlease rebase the branch and force-push.`);
             result = false;
+        }
+
+        // if no initial failure, run linting checks
+
+        if (result) {
+            const results = await Promise.all(
+                commits.map((commit: IPRCommit) => {
+                const linter = new LintCommit(commit);
+                return linter.lint();
+            }));
+
+            for (const lintError of
+                 (results.filter((el) => el) as ILintError[])) {
+                await addComment(lintError.message);
+                if (lintError.checkFailed) {
+                    result = false;
+                }
+            }
         }
 
         return result;
--- a/lib/ci-helper.ts
+++ b/lib/ci-helper.ts
@@ -646,24 +646,25 @@ export class CIHelper {
         let result: boolean = true;
         const maxCommits = 30;
         if (pr.commits && pr.commits > maxCommits) {
-            addComment(`The pull request has ${pr.commits
-                       } commits.  The max allowed is ${maxCommits
-                       }.  Please split the patch series into multiple pull ${
-                       ""}requests. Also consider squashing related commits.`);
+            await addComment(`The pull request has ${pr.commits
+                             } commits.  The max allowed is ${maxCommits
+                             }.  Please split the patch series into ${
+                             ""}multiple pull requests. Also consider ${
+                             ""}squashing related commits.`);
             result = false;
         }
 
         const commits = await this.github.getPRCommits(pr.baseOwner, pr.number);
 
         const merges: string[] = [];
-        commits.map((cm: IPRCommit) => {
+        await Promise.all(commits.map(async (cm: IPRCommit) => {
             if (cm.parentCount > 1) {
                 merges.push(cm.commit);
             }
 
             if (cm.author.email.endsWith("@users.noreply.github.com")) {
                 addComment(`Invalid author email in ${cm.commit}: "${
-                    cm.author.email}"`);
+                                 cm.author.email}"`);
                 result = false;
                 return;
             }
@@ -676,15 +677,15 @@ export class CIHelper {
                     userInfo.email = cm.committer.email;
                 }
             }
-        });
+        }));
 
         if (merges.length) {
-            addComment(`There ${
-                    merges.length > 1 ?
-                    "are merge commits" : "is a merge commit"
-                    } in this Pull Request:\n\n    ${
-                    merges.join("\n    ")
-                    }\n\nPlease rebase the branch and force-push.`);
+            await addComment(`There ${
+                             merges.length > 1 ?
+                             "are merge commits" : "is a merge commit"
+                             } in this Pull Request:\n\n    ${
+                             merges.join("\n    ")
+                             }\n\nPlease rebase the branch and force-push.`);
             result = false;
         }
 
--- a/lib/commit-lint.ts
+++ b/lib/commit-lint.ts
@@ -1,0 +1,134 @@
+import { IPRCommit } from "./github-glue";
+
+export interface ILintError {
+    checkFailed: boolean;           // true if check failed
+    message: string;
+}
+
+/*
+ * Simple class to drive lint tests on commit messages.
+ */
+export class LintCommit {
+    private blocked: boolean;
+    private lines: string[];
+    private patch: IPRCommit;
+    private messages: string[] = [];
+
+    public constructor(patch: IPRCommit) {
+        this.blocked = false;
+        this.lines =  patch.message.split("\n");
+        this.patch = patch;
+    }
+
+    /*
+     * Linter method to run checks on the commit message.
+     * @param {IPRCommit} the patch to be checked
+     *
+     * The lintings are in methods called from here.  If
+     * a lint check is too severe to continue, it will throw
+     * an error.
+     */
+    public async lint(): Promise<ILintError | void> {
+
+        // Basic test before all others
+
+        if (await this.commitViable()) {
+            await Promise.all([
+                this.commitMessageLength(),
+                this.lowerCaseAfterPrefix(),
+                this.signedOffBy(),
+                this.moreThanAHyperlink(),
+            ]);
+        }
+
+        if (this.messages.length) {
+            return { checkFailed: this.blocked,
+                     message: `There ${this.messages.length > 1 ?
+                     "are issues" : "is an issue"} in commit ${
+                     this.patch.commit}:\n${this.messages.join("\n")}`,
+                };
+        }
+    }
+
+    private addMessage(message: string): void {
+        this.messages.push(message);
+    }
+
+    private block(message: string): void {
+        this.blocked = true;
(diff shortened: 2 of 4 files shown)


dscho:
Thank you for opening the PR!

lib/commit-lint.ts
```diff
@@ -0,0 +1,152 @@
...
+        const msgs: string[] = [];
+
+        try {
+           // add a line for each test
+
+            // tslint:disable:curly
+           if (!await this.lint1(lines, msgs)) result = false;
+           if (!await this.lint2(lines, msgs)) result = false;
+           if (!await this.lint3(lines, msgs)) result = false;
+           if (!await this.lint4(lines, msgs)) result = false;
```
dscho: As I mentioned, I would like to make it the duty of the _caller_ to keep track of `result`.

Also, can we please use more descriptive names for the methods?

Finally, I think that this could be made easier to read by using a combination of `Promise.all()` and the `.bind()` method, i.e. something like:

```javascript
await Promise.all([
    commitMessageLength, lowerCaseAfterPrefix, signedOffBy, moreThanAHyperlink,
].map((method) => async {
    await (method.bind(this, lines, msgs)).();
});
```

lib/commit-lint.ts
```diff
@@ -0,0 +1,152 @@
...
+                patch.commit}:\n${msgs.join("\n")}`);
+        }
+
+        return result;
+    }
+/*
+TODO:
+There is also the convention that the commit message should talk in the
+imperative tone (as opposed to the past tense, as many other projects prefer).
+I am not sure that there is an easy way to test for that.
```
dscho: I think we can (and should) leave this to a future commit. Either we can use the Stanford Parser, as the `node-commit-msg` package does (which would require Java, but that's fine, I think), or we can go for something much, much simpler, e.g. imitating how [ELIZA](https://en.wikipedia.org/wiki/ELIZA) parses the English language to detect whether the sentence uses the present or past tense.

lib/commit-lint.ts
```diff
@@ -0,0 +1,152 @@
...
+*/
+    // This method tests:
+    // the body of the commit message should not be empty
+    // the first line should not exceed 76 characters
+    // the first line should be followed by an empty line
+
+    private async lint1(lines: string[], msgs: string[]):
+        Promise<boolean> {
```
dscho: If we keep `msgs` as a field, then we could actually make this a `Promise<void>` method family, and even record in another field whether _any_ blocking problems were detected. That would make the design more pleasant, I would think.

lib/commit-lint.ts
```diff
@@ -0,0 +1,152 @@
...
+        if (match) {
+            if (match[1].length) {
+                msgs.push(`Leading whitespace in sign off: ${
+                    line}`);
+                result = false;
+            }
+        } else {
+            msgs.push("Commit not signed off");
+            result = false;
+        }
```
dscho: I feel that this might be simpler if we do two things: introduce a method recording the message and whether there were blocking problems, and reordering the conditional arms:

```javascript
private async blocker(message: string): Promise<void> {
    this.addComment(message);
    this.result = false;
}
[...]
    if (!match) {
        blocker("Commit not signed off");
    } else if (match[1].length) {
        blocker(`Leading whitespace in sign off: ${line}`);
    }
```

lib/commit-lint.ts
webstech: Absolutely.  I had copied the initial list from the issue and moved the items as they were implemented.  I am removing the comment as there will still be an issue open for it.

lib/ci-helper.ts
```diff
@@ -687,6 +688,18 @@ export class CIHelper {
...
+        if (result) {
+            const linter = new LintCommit(addComment);
+
+            for (const cm of commits) {
+                if (!await linter.lint(cm)) {
+                    result = false;
+                }
+            }
+        }
+
```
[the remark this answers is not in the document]
webstech: Please see the update.  Not sure if it meets your severity level concern.

lib/commit-lint.ts
```diff
@@ -0,0 +1,152 @@
...
+     *
+     * The lintings are in methods called from here.  If
+     * a lint check is too severe to continue, it will throw
+     * an error.
+     */
+    public async lint(patch: IPRCommit):
+        Promise<boolean> {
+        let result: boolean = true;
+        const lines =  patch.message.split("\n");
+        const msgs: string[] = [];
```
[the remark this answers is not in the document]
webstech: Done.  Please see update.

lib/commit-lint.ts
webstech: Let me know if the changes resolve this.

webstech:
Reworked to use promises as suggested.  I think most of the comments have been addressed.  The `lint()` method will return a `ILintError` if there is a problem.  `addComment`, which was passed before can not be used in the async functions so its use has been moved back into ci-helper.ts.  This is to avoid possible problems with GitHubGlue being reused while it is waiting for I/O to complete.

lib/commit-lint.ts
```diff
@@ -0,0 +1,134 @@
...
+                };
+        }
+    }
+
+    private addMessage(message: string, failed: boolean = false): void {
+        this.messages.push(message);
+        if (failed) {
+            this.failed = true;
+        }
+    }
```
dscho: Maybe split this into two methods? For example, keep `addMessage()` as pushing into the array, but then also have `block()` which calls `this.addMessage()` and then sets `this.blocking = true;`? That way, nobody has to remember what that Boolean parameter is about.

lib/commit-lint.ts
```diff
@@ -0,0 +1,134 @@
...
+
+    // Verify if the first line starts with a prefix (e.g. tests:), it continues
+    // in lower-case
+
+    private async lowerCaseAfterPrefix(): Promise<void> {
+        const match = this.lines[0].match(/^([a-z]+)+?:.*?([A-Z])/);
+
+        if (match && typeof(match[2]) === "string" ) {
```
dscho: I don't think we have to verify that `match[2]` is of type `string` here.

lib/commit-lint.ts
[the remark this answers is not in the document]
dscho: 👍

lib/commit-lint.ts
dscho: Yes!

lib/commit-lint.ts
webstech: I think that was left over - removed with no effect on the unit tests.

lib/ci-helper.ts
```diff
@@ -687,6 +688,25 @@ export class CIHelper {
...
+        // if no initial failure, run linting checks
+
+        if (result) {
+            const results = await Promise.all(
+                commits.map((commit: IPRCommit) => {
+                const linter = new LintCommit(commit);
+                return linter.lint();
+            }));
+
+            for (const lintError of results) {
```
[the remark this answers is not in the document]
webstech: > I have a hunch that this would look even slicker if you used `.filter()` and `.map()`. For example, I could imagine that filtering out the `void`s could be done via `results.filter((el) => { el })` or some such.

Added `.filter()`.  Not sure creating a new array is more efficient but the code is cleaner.  The problem with `.map()` is the function is not async (and I have not seen a way to do it), so I left the for loop in place.  btw, once you approve this change I would like to add another commit before merging.  It would add the missing async to `addComment()` calls I missed in a previous PR. 😳

> And while we already have all the comments in one place, maybe we want to accumulate them into a single PR comment? Dunno... what do you think?

My concern with making a single comment is that it is easy to ignore parts of the comment.  Putting each commit in a separate comment allows them to be addressed individually.  OTOH, if there is a preference for different treatment based on `push` versus `submit comment`, that could be handled by `addComment()`.  If people complain about too many comments they need to improve their commit messages.  😉

lib/commit-lint.ts
dscho: Excellent!

lib/ci-helper.ts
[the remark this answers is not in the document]
dscho: > The problem with `.map()` is the function is not async (and I have not seen a way to do it),

I _think_ that `await Promise.all(something.map(...))` would work, but I am pretty new to this `async`/`await` business, I only started using it with GitGitGadget.

>  It would add the missing async to `addComment()` calls I missed in a previous PR.

I just added a commit to that end.

> My concern with making a single comment is that it is easy to ignore parts of the comment.

True.

dscho:
Thanks!

lib/ci-helper.ts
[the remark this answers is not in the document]
webstech: > > The problem with `.map()` is the function is not async (and I have not seen a way to do it),
> 
> I _think_ that `await Promise.all(something.map(...))` would work, but I am pretty new to this `async`/`await` business, I only started using it with GitGitGadget.

I see you added `Promise.all()` to your commit.  I have a concern with it.

1. The function is not returning the promise.  ie it should `return (or await) addComment()` so `Promise.all()` has a promise to wait on.
2. There appears to be a possible object reuse problem (see code below):
2.1 in the map() function addComment() is not run yet because it returns a promise
2.2 this can result in multiple pending addComment()
2.3 Promise.all() causes the addComment() functions to be started
2.4 if an addComment() is suspended in github communications, another addComment() will be started (normal node processing)
2.5 all the addComment() functions share a single GitHubGlue and octokit.  Could this lead to internal state errors in oktokit (two calls at the same time)?

Code to demo 2 above.  In this test, if addComment() is not returned/await'ed, jest gets errors.
```ts
test("Test async map", async () => {
    const addComment = async (body: string) => {
        console.log(`Adding comment to ${body}`);
        await setTimeoutPromise(4000, "foobar"); // pretend GutHubGlue.addPRComment()
        console.log(`Added comment to ${body}`);
    };

    const commits = [ "1", "2", "3"];
    console.log("starting");
    await Promise.all(commits.map(async (cm: string) => {
        return addComment(cm);
    }));
    console.log("promises done");
});
```
Does this make sense?
> I just added a commit to that end.

Thanks!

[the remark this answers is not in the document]
dscho: True. So we have to convert to a `for (... of ...) { ... await addComment(...) ...}` loop?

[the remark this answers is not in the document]
webstech: Probably the simplest.  The initial code had no async calls.  Are you doing it or shall I?

PR #188, opened 2020-01-13
Do not omit From: line for single-patch contributions

dscho:
When sending a patch on behalf of somebody else, the first line of the mail body should be the appropriate `From:` line.

In GitGitGadget, we special-case the cover letter _not_ to include such a line. However, if there is only a single mail without any cover letter, the current code mistakes it for being the cover letter, and consequently omits the `From:` line, misattributing the authorship.

That happened [in this contribution](https://public-inbox.org/git/pull.690.git.git.1578576634678.gitgitgadget@gmail.com/) (which made it into git.git as https://github.com/git/git/commit/49e268e23e313e8c6b009cd8ebbbaae4316cf6cc).

This PR fixes that.

### DIFF
--- a/lib/patch-series.ts
+++ b/lib/patch-series.ts
@@ -468,7 +468,7 @@ export class PatchSeries {
             }
 
             header = authorMatch[1] + replaceSender + authorMatch[3];
-            if (i === 0 && senderName) {
+            if (mails.length > 1 && i === 0 && senderName) {
                 // skip Cc:ing and From:ing in the cover letter
                 mails[i] = header + match[2];
                 return;


lib/patch-series.ts
```diff
@@ -468,7 +468,7 @@ export class PatchSeries {
             }
 
             header = authorMatch[1] + replaceSender + authorMatch[3];
-            if (i === 0 && senderName) {
+            if (mails.length > 1 && i === 0 && senderName) {
```
webstech: Does this also need to be taken into account when setting `onBehalfOf` at line 456?

dscho: No (I tried to explain that in the commit message, but admittedly that explanation is a bit convoluted).

Imagine that you write a patch, open a PR at git-for-windows/git and I merge it. The implicit agreement is that it is now _my_ responsibility to shepherd that patch into Git. So I open a PR at gitgitgadget/git. You are the author, I am the sender. The patch should still be sent on my behalf, but the `From:` line with your authorship should be prepended to the mail. Read: `onBehalfOf` is still correct (at least in my opinion).

dscho:
Thanks @webstech!

PR #189, opened 2020-01-15
Fix overzealous linting

dscho:
The linter indicated a problem in https://github.com/gitgitgadget/git/pull/527, but there is not actually a problem in that PR, there is a problem in the linter ;-)

Let's fix this.

### DIFF
--- a/lib/commit-lint.ts
+++ b/lib/commit-lint.ts
@@ -92,7 +92,7 @@ export class LintCommit {
     // in lower-case
 
     private async lowerCaseAfterPrefix(): Promise<void> {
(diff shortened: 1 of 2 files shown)


dscho:
> The linter indicated a problem in [gitgitgadget/git#527](https://github.com/gitgitgadget/git/pull/527)

I marked the comment as resolved, but it can still be viewed: https://github.com/gitgitgadget/git/pull/527#issuecomment-574867260

dscho:
> Misinterpreted 'continue in lower case' to mean the rest of the first line of the message.

No worries, I was not quite clear, and even worse, I missed this in the code review. I'm just glad that it did not bite anyone else but me ;-)

> Thinking about a change to the message but get this fix in first.

Thank you!

PR #191, opened 2020-01-16
Linting should not require signed-off be the last line

webstech:
Reviewed-by may be added after Signed-off-by.

### DIFF
--- a/lib/commit-lint.ts
+++ b/lib/commit-lint.ts
@@ -100,17 +100,25 @@ export class LintCommit {
         }
     }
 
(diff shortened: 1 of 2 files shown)


newren:
> [Timely](https://github.com/git/git/pull/692)! Thank you so much!

Indeed, thanks!

webstech:
I noticed the failed job so it was more reactive than timely.  I have opened #192 to allow devs to run linting against existing commits and decide if the linting is breaking anything.

Azure is making it hard to see jobs at the moment.  Not relevant to anyone but me whining.

Issue #193, opened 2020-01-17
Accommodate for authors/committers without GitHub login information


dscho:
Apparently we're now getting incomplete data (or "not as complete as we want") upon pushes (see [here](https://dev.azure.com/gitgitgadget/git/_build/results?buildId=27120&view=logs&j=f4a2bb40-7e70-5f50-d570-7d7c5c8ca123&t=47c053e2-7eee-557e-56d8-cf38bc303dc2):

```
+ node build/script/misc-helper.js handle-pr-push git 695
Caught error TypeError: Cannot read property 'login' of null:
TypeError: Cannot read property 'login' of null
    at /home/gitgitgadget-agent/agent/_work/6/s/build/lib/github-glue.js:266:45
    at Array.map (<anonymous>)
    at GitHubGlue.<anonymous> (/home/gitgitgadget-agent/agent/_work/6/s/build/lib/github-glue.js:256:27)
    at Generator.next (<anonymous>)
    at fulfilled (/home/gitgitgadget-agent/agent/_work/6/s/build/lib/github-glue.js:4:58)
    at processTicksAndRejections (internal/process/task_queues.js:94:5)
```

The [line in question](https://github.com/gitgitgadget/gitgitgadget/blob/ac3ed5a82ef77528631784870d822fcdca927bd3/lib/github-glue.ts#L266):

```typescript
            baseOwner: response.data.base.repo.owner.login,
```

Seems that we have to obtain that information in some other way?

dscho:
> ```ts
> baseOwner: response.data.base.repo.owner.login,
> ```

Seems that `response.data.base.repo.owner` is not a mere container but a full object. Maybe that has anything to do with it...

dscho:
Aww shucks. I managed to overlook that this is not on the _Typescript_ but on the _Javascript_ file. So line 266 in the `.js` actually seems to correspond to this line: https://github.com/gitgitgadget/gitgitgadget/blob/ac3ed5a82ef77528631784870d822fcdca927bd3/lib/github-glue.ts#L327

dscho:
So it's not even a change in the GitHub API. It's just slightly incorrect Typescript information for OctoKit: the `author`/`committer` attribute _needs not be set_! For example, if you make up an author with a funny name and a bogus email and you push it and then ask the GitHub API to show you the `author` object for this commit: there is none!

PR #194, opened 2020-01-17
GitHubGlue.getPRCommits(): do not expect valid login information

dscho:
When users push commits with author or committer information listing an email that GitHub does not know about, the GitHub API apparently simply does not return the `author` or`committer` object when being asked for commits.

This PR teaches GitGitGadget to handle that scenario gracefully.

This fixes https://github.com/gitgitgadget/gitgitgadget/issues/193.

### DIFF
--- a/lib/github-glue.ts
+++ b/lib/github-glue.ts
@@ -318,13 +318,13 @@ export class GitHubGlue {
             result.push({
                 author: {
                     email: cm.commit.author.email,
-                    login: cm.author.login,
+                    login: cm.author ? cm.author.login : "unknown login",
                     name: cm.commit.author.name,
                 },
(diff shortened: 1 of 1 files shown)


sunshineco:
Perhaps apply `s/are apparently/apparently/` to the commit message of patch 1/1.

dscho:
> Perhaps apply `s/are apparently/apparently/` to the commit message of patch 1/1.

Thanks, fixed.

dscho:
>  It would be nice if the type indicated it could be null (changing it would probably break a lot of code).

I actually looked into this, and I did not even find out how the types are generated, I only found out how the routes are translated (but they do not contain information about the return types). 😦

Issue #196, opened 2020-01-22
Adjust the commit message linter to catch also prefixes like `git-p4:`


dscho:
In https://github.com/git/git/pull/698, we let slip through a commit that continues in upper case.

webstech:
Should a prefixed message be any message with no whitespace before the ":"?  A quick check shows a wide variety of characters may be present, including all caps (e.g. CI:).

dscho:
My instinct says we should go for `[A-Za-z][-_/()A-Za-z]*:`.

webstech:
Extending the check also showed a couple more possibilities: `[^-_/.()A-Za-z0-9\[\]{}<>*,]`.  I think there was only one comma so it could probably be removed.  What do you think?

dscho:
> Extending the check also showed a couple more possibilities: `[^-_/.()A-Za-z0-9\[\]{}<>*,]`. I think there was only one comma so it could probably be removed. What do you think?

Makes sense, I think. For my education, what was the oneline with the `,`?

webstech:
> For my education, what was the oneline with the `,`?

Seems it was only one that only had a single `,`.  Others had some of the other characters as well.
0bbc0bc574: {fetch,upload}-pack: sideband v2 fetch response
d76ce4f734: log,diff-tree: add --combined-all-paths option
76a7bc092e: cmd_{read,write}_tree: rename "unused" variable that is used

Do you want to include the comma?  What about `|`?  I could easily use `cmd_(read|write)_tree`.  Are there any characters that should be excluded?  Currently anything not included means the prefix followed by lower case rule is not matched.

dscho:
> Do you want to include the comma?

Based on those example, I would say: yes ;-)

> Are there any characters that should be excluded?

Hmm. Not sure. Maybe you're right, and we should simply go for `^\S+:\s*[a-z]`.

I am just slightly worried that there might be a legitimate thing that matches this but is _not_ a prefix. Something like `C:\WINDOWS is the system directory on Windows` or some such.

But I guess we can always go with the simple regex until the time when it triggers, and then take it from there.

webstech:
Can this be closed?

dscho:
Thanks for the reminder!

PR #197, opened 2020-01-23
Allow any character for commit message prefix lint

webstech:
There are a wide variety of characters used.  Rather than restrict
the characters when there is no documented restriction, allow any
characters.  If there is a problem, it can be handled with a future
change.

This fixes issue #196.


dscho:
@webstech thank you again, so much, for your excellent contributions!

PR #205, opened 2020-02-24
Eslint Prep Work

webstech:
tslint is being deprecated.  There is some tooling to support the conversion to eslint and some typescript specific linting.  This is an initial pass to update the code for the low hanging fruit.  Best  reviewed at the commit level.  There are very few logic changes.

[typescript-eslint](https://github.com/typescript-eslint/typescript-eslint) has more info about converting.  Initial playing with the conversion required turning indent rules off - the errors and warnings were not consistent.  The typescript-eslint team acknowledged in an issue they do not use the indent rules and use prettier instead.

### DIFF
--- a/lib/git.ts
+++ b/lib/git.ts
@@ -121,7 +121,8 @@ export function git(args: string[], options?: IGitOptions | undefined):
  *    the working directory in which to run `git rev-parse`
  * @returns { string | undefined } the full SHA-1, or undefined
  */
-export async function revParse(argument: string, workDir?: string) {
+export async function revParse(argument: string, workDir?: string):
+    Promise<string | undefined> {
     const result = await GitProcess.exec(["rev-parse", "--verify", "-q",
                                           argument],
                                          workDir || ".");
@@ -137,7 +138,7 @@ export async function revParse(argument: string, workDir?: string) {
  * @returns number the number of commits in the commit range
  */
 export async function revListCount(rangeArgs: string | string[],
-                                   workDir: string = "."):
+                                   workDir = "."):
     Promise<number> {
     const gitArgs: string[] = ["rev-list", "--count"];
     if (typeof(rangeArgs) === "string") {
--- a/lib/mail-archive-helper.ts
+++ b/lib/mail-archive-helper.ts
@@ -62,9 +62,8 @@ export class MailArchiveGitHelper {
             return "";
         }
 
-        return "``````````\n" +
-            body + (body.endsWith("\n") ? "" : "\n") +
-            "``````````\n";
+        return "``````````\n"
+            + body + (body.endsWith("\n") ? "" : "\n") + "``````````\n";
     }
 
     protected readonly state: IGitMailingListMirrorState;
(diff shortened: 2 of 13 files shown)


lib/mail-archive-helper.ts
```diff
@@ -62,9 +62,8 @@ export class MailArchiveGitHelper {
             return "";
         }
 
-        return "``````````\n" +
-            body + (body.endsWith("\n") ? "" : "\n") +
-            "``````````\n";
+        return "``````````\n"
+            + body + (body.endsWith("\n") ? "" : "\n") + "``````````\n";
```
dscho: I guess this cannot be converted to a template? ;-)

lib/git.ts
```diff
@@ -137,7 +138,7 @@ export async function revParse(argument: string, workDir?: string) {
  * @returns number the number of commits in the commit range
  */
 export async function revListCount(rangeArgs: string | string[],
-                                   workDir: string = "."):
+                                   workDir = "."):
```
dscho: This change probably wanted to go into "remove inferred types"...

lib/send-mail.ts
```diff
@@ -7,7 +7,7 @@ export interface IParsedMBox {
     cc?: string[];
     date?: string;
     from?: string;
-    headers?: Array<{ key: string; value: string; }>;
+    headers?: Array<{ key: string; value: string }>;
```
dscho: Technically, this change does not fit the commit message, but I do not object strongly to it being in the same commit as the other `;`-related changes.

lib/patch-series.ts
```diff
@@ -86,20 +86,20 @@ export class PatchSeries {
             match = tagMessage.match(/^[\s\S]*?\n\n([\s\S]*)/);
             (match ? match[1] : tagMessage).split("\n").map((line) => {
                 // tslint:disable-next-line:max-line-length
-                match = line.match(/https:\/\/lore\.kernel\.org\/.*\/([^\/]+)/);
```
dscho: Oh right, this is my fault. I am _so_ used to escaping the slash in regular expressions that I escape it even within character ranges... Thanks for cleaning up my mess!

lib/mail-archive-helper.ts
webstech: Maybe next time.
```
const separator = "``````````\n";
return `${separator}${body}${body.endsWith("\n") ? "" : "\n"}${separator}`;
```

webstech:
Thanks for the review.  Please go ahead with the merge.  I will work on my git skills for rolling back, updating and replaying history on my own time.  I had initially done all the changes at once and then broke them up into commits.  Maybe not the best strategy but there were a number of linting messages to get through.  The use of '+' caused numerous indent issues until I turned it off.

lib/mail-archive-helper.ts
dscho: :-)

dscho:
> I had initially done all the changes at once and then broke them up into commits.

That's exactly how I perform similar tasks...

I'll just merge it, the shape of the commits is not _all_ that important ;-)

PR #208, opened 2020-02-27
Fix parsing of mails fetched from lore.kernel.org

dscho:
Due to a mistake of yours truly, the [Azure Pipeline that takes care of synchronizing the replies on the Git mailing list back to their corresponding PRs](https://dev.azure.com/gitgitgadget/git/_build?definitionId=5&_a=summary) did _not_ actually switch to using lore.kernel.org.

Which is a good thing because due to another mistake of yours truly, the parser was not adjusted to reflect that lore.kernel.org uses a different layout than public-inbox.org: instead of adding new mails to the tree, the single file in the tree is overwritten with the latest mail all the time.

This has a couple interesting ramifications, e.g. we need to skip the `-` lines in the diffs (which was not a problem before because there simply were no `-` lines), we need to increase the diff context (which was not a problem because there was no pre-image, i.e. we automatically knew that we saw all of the post-image's lines).

This PR should fix that.

### DIFF
--- a/lib/mail-archive-helper.ts
+++ b/lib/mail-archive-helper.ts
@@ -178,7 +178,7 @@ export class MailArchiveGitHelper {
                     counter = parseInt(match[3], 10);
                     buffer = "";
                 }
-            } else if (counter) {
+            } else if (counter && line.match(/^[ +]/)) {
                 buffer += line.substr(1) + "\n";
                 if (--counter) {
                     return;
--- a/lib/mail-archive-helper.ts
+++ b/lib/mail-archive-helper.ts
@@ -244,7 +244,7 @@ export class MailArchiveGitHelper {
         }
 
         const range = `${this.state.latestRevision}..${head}`;
-        await git(["log", "-p", "--reverse", range],
+        await git(["log", "-p", "-U99999", "--reverse", range],
                   { lineHandler, workDir: this.mailArchiveGitDir });
 
         this.state.latestRevision = head;
--- a/lib/mail-archive-helper.ts
+++ b/lib/mail-archive-helper.ts
@@ -201,6 +201,10 @@ export class MailArchiveGitHelper {
              */
             this.state.latestRevision =
                 "3b38d8d206c64bf3dc873ba8ae9dbd48ed43f612";
+        } else if (this.state.latestRevision ===
+            "205655703b0501ef14e0f0dddf8e57bb726fae97") {
+            this.state.latestRevision =
+                "26674e9a36ae1871f69197798d30f6d3d2af7a56";
         } else if (await revParse(this.state.latestRevision,
                                   this.mailArchiveGitDir) === undefined) {
             const publicInboxGitDir = process.env.PUBLIC_INBOX_DIR;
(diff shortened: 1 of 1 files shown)


dscho:
Hmm. I see that there were replies e.g. to https://lore.kernel.org/git/pull.714.git.git.1582747775.gitgitgadget@gmail.com, but they have not been mirrored to https://github.com/git/git/pull/714. Investigating.

dscho:
Seems to have worked now: https://github.com/git/git/pull/714#discussion_r385076626

lib/mail-archive-helper.ts
```diff
@@ -172,10 +173,13 @@ export class MailArchiveGitHelper {
...
                 const match = line.match(/^@@ -(\d+,)?\d+ \+(\d+,)?(\d+)?/);
                 if (match) {
+                    if (counter) {
+                        console.log(`Oops: unprocessed buffer ${buffer}`);
+                    }
                     counter = parseInt(match[3], 10);
                     buffer = "";
                 }
-            } else if (counter) {
+            } else if (counter && line.match(/^[ +]/)) {
```
webstech: Not sure that `counter` matters anymore.  You are in a patch or not in a patch.  A patch starts with the `@@` line and continues with lines `/^[-+ ]/`.  Using counter will have problems when there are common lines after the last insert and decrementing it for common lines will cut off the email.  See [this code around line 170](https://github.com/webstech/gitgitgadget/commit/8708b800ccbc963e4b91a59381b35dfe0e30d1d2) where the patch state is terminated by a character that is not the patch prefix.  It also handles the last buffer at the end.

dscho: > Using counter will have problems when there are common lines after the last insert and decrementing it for common lines will cut off the email.

That should not be the case. The `+` range in the `@@` line includes the context (or "common") lines. Otherwise the diff is corrupt.

> Not sure that `counter` matters anymore. You are in a patch or not in a patch.

I wanted to make sure that I work on a file by file basis ;-) That's why I parse the `@@` header and then act as soon as the hunk is complete: the expectation is that the hunk encompasses the complete file.

And I would like to keep this code, as it would probably even handle future changes to the layout where multiple mails might be added at once (not that I expect such a change to come, but still, I like the code to be prepared to handle even that).

webstech: > The + range in the @@ line includes the context (or "common") lines.

My mistake.  That makes my changes just a different implementation.

fyi, commit c9379c29a7493780a2f1d39b3f14a12d867d5ecf in the lore repo has two diffs.  The first one deletes all the content and the second one adds the new content.  There are a few of them like that.

dscho: > commit c9379c29a7493780a2f1d39b3f14a12d867d5ecf in the lore repo has two diffs. The first one deletes all the content and the second one adds the new content. There are a few of them like that.

Interesting! I had not realized that. It removes the file `d` and adds the file `m`... 

I guess the reason is fb163402d41dbf62a3b336b3f6f39cb6fe56c9e8 which marks `m` as spam. In the previous layout, those spam mails would be removed from the resulting tree... I guess it is a downside of the new layout that you cannot remove spam mails effectively (because the mails are no longer supposed to live in one huge top-level tree but purely in the Git object database instead).

PR #210, opened 2020-03-08
Update to latest github api (@octokit/rest)

webstech:
The updated api has some renamed functions and variables.

The `set-app-token` command was updated to the new authentication
method.

MailArchiveGitHelper.processMails() no longer tracks the comment id of
replies.  The api does not support referencing replies to replies.

GitHubGlue methods were tested using a manual test to be added in a separate PR.  The new authentication was tested using a private appid and repo.

### DIFF
--- a/lib/github-glue.ts
+++ b/lib/github-glue.ts
@@ -1,4 +1,4 @@
-import octokit = require("@octokit/rest");
+import { Octokit } from "@octokit/rest";
 import { git, gitConfig } from "./git";
 import { GitGitGadget } from "./gitgitgadget";
 
@@ -50,7 +50,7 @@ export interface IGitHubUser {
 
 export class GitHubGlue {
     public workDir?: string;
-    protected client = new octokit();
+    protected client = new Octokit();
     protected authenticated?: string;
 
     public constructor(workDir?: string) {
@@ -102,7 +102,7 @@ export class GitHubGlue {
         await this.ensureAuthenticated(owner);
         const status = await this.client.issues.createComment({
             body: comment,
-            number: nr,
+            issue_number: nr,
             owner,
             repo,
         });
@@ -137,10 +137,10 @@ export class GitHubGlue {
         const status = await this.client.pulls.createComment({
             body: comment,
             commit_id: commit,
-            number: nr,
             owner,
             path,
             position: 1,
+            pull_number: nr,
             repo,
         });
         return {
@@ -164,11 +164,11 @@ export class GitHubGlue {
             GitGitGadget.parsePullRequestURL(pullRequestURL);
         await this.ensureAuthenticated(owner);
 
-        const status = await this.client.pulls.createCommentReply({
+        const status = await this.client.pulls.createReviewCommentReply({
             body: comment,
-            in_reply_to: id,
-            number: nr,
+            comment_id: id,
             owner,
+            pull_number: nr,
             repo,
         });
         return {
@@ -184,8 +184,8 @@ export class GitHubGlue {
 
         await this.ensureAuthenticated(owner);
         const result = await this.client.issues.addLabels({
+            issue_number: prNo,
             labels,
-            number: prNo,
             owner,
             repo,
         });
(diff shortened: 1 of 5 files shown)


lib/github-glue.ts
```diff
@@ -225,7 +225,7 @@ export class GitHubGlue {
             repo: "git",
             state: "open",
         });
-        response.data.map((pr: octokit.PullsListResponseItem) => {
+        response.data.map((pr) => {
```
dscho: This seems to be another ESLint-related change, rather than an Octokit-related one?

lib/github-glue.ts
```diff
@@ -314,7 +314,7 @@ export class GitHubGlue {
             repo: "git",
         });
         const result: IPRCommit[] = [];
-        response.data.map((cm: octokit.PullsListCommitsResponseItem) => {
+        response.data.map((cm) => {
```
dscho: This seems to be another ESLint-related change, rather than an Octokit-related one?

lib/mail-archive-helper.ts
```diff
@@ -138,10 +138,9 @@ export class MailArchiveGitHelper {
                     MailArchiveGitHelper.mbox2markdown(parsed);
 
                 if (issueCommentId) {
-                    const result =  await this.githubGlue
-                        .addPRCommentReply(pullRequestURL, issueCommentId,
-                                           comment);
-                    issueCommentId = result.id;
+                    await this.githubGlue.addPRCommentReply(pullRequestURL,
+                                                            issueCommentId,
+                                                            comment);
```
dscho: This seems to be another ESLint-related change, rather than an Octokit-related one?

webstech:
> I do not understand this comment, though: from how I read the code, this functionality is maintained, and the replies-to-replies-referencing is being maintained, even while switching to a newer API function. Maybe I misunderstand, and you could clarify?

From the [api doc](https://octokit.github.io/rest.js/v17#pulls-create-review-comment-reply): 

> For the comment_id, provide the ID of the review comment you are replying to. This must be the ID of a top-level review comment, not a reply to that comment. **Replies to replies are not supported.**

`issueCommentId = result.id;` was removed from `MailArchiveGitHelper` (old line 144) when adding a comment reply, so `issueCommentId` always refers to the original comment.

lib/github-glue.ts
webstech: `response.data` is already correctly typed so the typing was unnecessary (and just changing 'o' to 'O' did not work 🙄).

lib/github-glue.ts
webstech: See the previous.

lib/mail-archive-helper.ts
webstech: `result` was removed so I chose this reformat (`await this.githubGlue` looked lonely on a line by itself).

lib/github-glue.ts
dscho: Yes, I understood, but I thought that this change was maybe not necessitated by switching to the new Octokit version?

lib/mail-archive-helper.ts
dscho: 👍

dscho:
> > For the comment_id, provide the ID of the review comment you are replying to. This must be the ID of a top-level review comment, not a reply to that comment. **Replies to replies are not supported.**
> 
> `issueCommentId = result.id;` was removed from `MailArchiveGitHelper` (old line 144) when adding a comment reply, so `issueCommentId` always refers to the original comment.

That makes sense to me. Thanks for indulging me!

lib/github-glue.ts
webstech: Changing `octokit` to `Octokit` still got an error so I did the most expeditious change instead of figuring out how to specify the type (it also clarifies the object is not being cast to something else).

dscho:
I will keep an eye out for build failures in https://dev.azure.com/gitgitgadget/git/_build/index?definitionId=5, but do not actually expect any problems.

Thank you @webstech!

dscho:
Okay, it seems to work really well. See e.g. https://dev.azure.com/gitgitgadget/git/_build/results?buildId=32842&view=logs&j=275f1d19-1bd8-5591-b06b-07d489ea915a&t=cf7470c7-7d35-5d46-df77-7e678f528994&l=15 which created https://github.com/gitgitgadget/git/pull/576#issuecomment-596593361.

Oh, and congratulations for opening the 100th PR in this here repository! 🎉

PR #215, opened 2020-03-31
Use libqp to decode quoted-printable strings

webstech:
Multi-byte utf-8 characters were being treated as single byte
characters, with each byte converted to a character.  This
caused quoted-printable strings to be decoded incorrectly.
The libqp package is now used to convert each byte to the
hex value in a buffer.  The buffer is converted to a string.

Also, return in `mbox2markdown()` will now use a constant on
the return statement to avoid repetition.

This corrects a problem in PR #160 as noted by recent comments in issue #158.


dscho:
Thank you so much @webstech !

Issue #220, opened 2020-04-06
Correct handling for Cc field for single-commit submissions


newren:
Quoting a conversation on the git mailing list (from https://lore.kernel.org/git/nycvar.QRO.7.76.6.2004041546290.46@tvgsbejvaqbjf.bet/):

> > > I'm not sure how cc is supposed to work with GGG, but it clearly didn't
> > > here. :)
> >
> > Yeah, I clearly don't either.  I even looked up another submission
> > from Dscho (https://github.com/git/git/pull/728) and attempted to
> > mimic it, but still managed to get it wrong somehow and I don't know
> > how.
> 
> It might be a bug in the way I implemented folding the cover letter into
> single-patch mails. I simply might not pick up the Cc:s there.
> 
> However, I am seriously under water right now and won't be able to work on
> a fix for that. Would you mind opening a ticket at
> https://github.com/gitgitgadget/gitgitgadget/issues/new?
> 
> Thanks,
> Dscho

webstech:
> Quoting a conversation on the git mailing list (from https://lore.kernel.org/git/nycvar.QRO.7.76.6.2004041546290.46@tvgsbejvaqbjf.bet/):
> 
> > > I'm not sure how cc is supposed to work with GGG, but it clearly didn't
> > > here. :)

This came up in [git/git#740](https://github.com/git/git/pull/740#issuecomment-605523086) as well.  The cc is treated as the body when there is no body/description.

> > Yeah, I clearly don't either.  I even looked up another submission
> > from Dscho ([git/git#728](https://github.com/git/git/pull/728)) and attempted to

Markdown format on that one looks like it should be okay.

dscho:
[...]

PR #221, opened 2020-04-07
More single-patch support of PR description

webstech:
This is a three part change for PatchSeries.

1. Allow an empty PR description to be passed in (assuming
   this has been validated elsewhere)
2. Handle `footers` present in the PR description with no
   other text (ie no description)
3. Recognize the Based-on: footer.  The case statement was
    missing `:`.

This fixes https://github.com/gitgitgadget/gitgitgadget/issues/220

### DIFF
--- a/lib/patch-series.ts
+++ b/lib/patch-series.ts
@@ -255,17 +255,14 @@ export class PatchSeries {
             // Just ignore it
         }
 
-        if (!prBody.length) {       // reject empty description
-            throw new Error("A pull request description must be provided");
-        }
-
         const {
             basedOn,
             cc,
             coverLetterBody,
         } = PatchSeries.parsePullRequestBody(prBody);
 
-        const coverLetter = `${prTitle}\n\n${coverLetterBody}`;
+        const coverLetter = `${prTitle}\n${coverLetterBody.length ?
+            `\n${coverLetterBody}` : ""}`;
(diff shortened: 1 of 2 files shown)


lib/patch-series.ts
```diff
@@ -288,7 +285,13 @@ export class PatchSeries {
         let coverLetterBody = prBody.trim();
 
         // parse the footers of the pullRequestDescription
-        const match = prBody.match(/^([^]+)\n\n([^]+)$/);
+        let match = prBody.match(/^([^]+)\n\n([^]+)$/);
+
+        if (!match && !prBody.match(/\n\n/)) {
+            // handle PR descriptions that have no body, just footers
+            match = prBody.match(/^()([A-Za-z]+: [^]+)$/);
```
dscho, line 292: I think we need to allow dashes in footer names, like `Based-On:`. I.e. 
```suggestion
            match = prBody.match(/^()([-A-Za-z]+: [^]+)$/);
```

webstech: Updated - sorry about that.  Added a test for based-on and it was not detected.  Updated the code to detect it.

However, this means based-on may not be well tested in the rest of the code.  I plan to add tests for PatchSeries, which will also test ProjectOptions but may add a separate ProjectOptions test.

dscho: Thank you!

dscho:
Awesome work, as always!

PR #222, opened 2020-04-08
Update nodemailer to 6.4.6

webstech:
A minor code change was needed to get the returned type from
addressParser correct.


dscho:
Thanks!

PR #224, opened 2020-04-15
Add GitHubGlue tests

webstech:
This test requires setup.  It will run successfully if setup has
not been done.  If the test fails, there may be a pull request and a
branch to be deleted on github and the local repo.  The test will
attempt cleanup of a previously failed test.

Setup:
gitgitgadget.githubtest.githubuser must be configured for this test to
identify your GitHub login.
The value must also be configured to identify a test repo to use as
gitgitgadget.<login>.githubrepo.

Additionally, a GitHub personal access token must be set for the
login to succeed.  This can be restricted to the test repo.

For example, these configuration settings are needed (where
`octokity` is your GitHub login):
git config --add gitgitgadget.githubtest.githubuser octokity
git config --add gitgitgadget.octokity.githubrepo ggg-test
git config --add gitgitgadget.octokity.githubtoken feedbeef...

The test repo must exist in github and locally.  It is expected to be
located at the same directory level as this project.

### DIFF
--- a/tests/github-glue.test.ts
+++ b/tests/github-glue.test.ts
@@ -1,0 +1,199 @@
+import { Octokit } from "@octokit/rest";
+import "jest";
+import { git, gitConfig } from "../lib/git";
+import { GitHubGlue } from "../lib/github-glue";
+
+/*
+This test requires setup.  It will run successfully if setup has
+not been done.  If the test fails, there may be a pull request and a
+branch to be deleted on github and the local repo.  The test will
+attempt cleanup of a previously failed test.
+
+Setup:
+gitgitgadget.githubtest.githubuser must be configured for this test to
+identify your GitHub login.
+The value must also be configured to identify a test repo to use as
+gitgitgadget.<login>.githubrepo.
+
+Additionally, a GitHub personal access token must be set for the
+login to succeed.  This can be restricted to the test repo.
+
+For example, these configuration settings are needed (where
+`octokity` is your GitHub login):
+git config --add gitgitgadget.githubtest.githubuser octokity
+git config --add gitgitgadget.octokity.githubrepo ggg-test
+git config --add gitgitgadget.octokity.githubtoken feedbeef...
+
+The test repo must exist in github and locally.  It is expected to be
+located at the same directory level as this project (ie ../).
+*/
+
+class GitHubProxy extends GitHubGlue {
+    public octo: Octokit | any;
+    public constructor(workDir?: string, repo = "git") {
+        super(workDir, repo);
+        this.octo = this.client;
+    }
+
+    public async authenticate(repositoryOwner: string): Promise<void> {
+        await this.ensureAuthenticated(repositoryOwner);
+        this.octo = this.client;
+    }
+}
+
+jest.setTimeout(180000);
+
+let owner: string;
+let repo: string;
+
+beforeAll(async () => {
+    owner = await gitConfig(`gitgitgadget.githubtest.githubuser`) || "";
+    repo = await gitConfig(`gitgitgadget.${owner}.githubrepo`) || "";
+});
+
+test("identify user", async () => {
+    if (owner && repo) {
+        const userName = await gitConfig(`user.name`, `../${repo}`) || "";
+
+        const github = new GitHubProxy(`../${repo}`, repo);
+        await github.authenticate(owner);
+
+        const name = await github.getGitHubUserName(owner);
+        expect(name).toMatch(userName);
+
+        const ghUser = await github.getGitHubUserInfo(owner);
+        expect(ghUser.login).toMatch(owner);
+        expect(ghUser.name).toMatch(userName);
+    }
+});
+
+test("pull requests", async () => {
+    if (owner && repo) {
+        const repoDir = `../${repo}`;
+
+        const github = new GitHubProxy(repoDir, repo);
+        await github.authenticate(owner);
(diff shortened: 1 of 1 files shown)


tests/github-glue.test.ts
```diff
@@ -0,0 +1,204 @@
...
+git config --add gitgitgadget.githubtest.githubuser octokity
+git config --add gitgitgadget.octokity.githubrepo ggg-test
+git config --add gitgitgadget.octokity.githubtoken feedbeef...
+
+The test repo must exist in github and locally.  It is expected to be
+located at the same directory level as this project.
+*/
+
+class GitHubProxy extends GitHubGlue {
+    public octo = new Octokit();
```
dscho, line 32: Should this be initialized? The constructor always overrides it.

Taking a step back, though, I wonder whether we really have to have this class. All we want is to access a private field, right? We could always use the trick `const octo = github["client"];`

tests/github-glue.test.ts
```diff
@@ -0,0 +1,204 @@
...
+        await this.ensureAuthenticated(repositoryOwner);
+        this.octo = this.client;
+    }
+}
+
+jest.setTimeout(180000);
+
+test("identify user", async () => {
+    const owner = await gitConfig(`gitgitgadget.githubtest.githubuser`) || "";
+    const repo = await gitConfig(`gitgitgadget.${owner}.githubrepo`) || "";
```
dscho, line 48: Maybe this could be done as part of a setup phase (see https://jestjs.io/docs/en/setup-teardown.html)?

tests/github-glue.test.ts
```diff
@@ -0,0 +1,204 @@
...
+
+    if (owner && repo) {
+        const userName = await gitConfig(`user.name`, `../${repo}`) || "";
+
+        const github = new GitHubProxy(`../${repo}`, repo);
```
dscho, line 53: So the repository needs to be checked out next to the current worktree? If that is the case, I think we might want to add that to the description of the setup.

tests/github-glue.test.ts
```diff
@@ -0,0 +1,204 @@
...
+
+        try {
+            const name = await github.getGitHubUserName(owner);
+            expect(name).toMatch(userName);
+
+            const ghUser = await github.getGitHubUserInfo(owner);
+            expect(ghUser.login).toMatch(owner);
+            expect(ghUser.name).toMatch(userName);
+        } catch (error) {
+            console.log(`git commmand failed\n${error}`);
```
dscho, line 64: I guess we would want to tell Jest that this is a test failure, right?

tests/github-glue.test.ts
webstech: The initializer was originally there to handle old vs current Octokit.  I had not figured out how to declare type.  Changing to specify type that makes typescript happy.

Sorry if I am missing something.  Attempts to use the private field are not working for me.

tests/github-glue.test.ts
webstech: I can see using a common function to return this info in the test cases but not clear how a setup phase can provide owner and repo vars.  Not joking to say I am still learning jest.

tests/github-glue.test.ts
webstech: Please suggest better wording for lines ~27-28.  When you are nose deep in something, what you think is obvious may not be.  I am open to (and appreciate) suggestions.

tests/github-glue.test.ts
webstech: Yes, that is an excellent idea.  I should have removed test code from earlier incarnations of this code.

tests/github-glue.test.ts
dscho: What I would do is to declare both via `let`, outside the functions, and then set then in the `beforeEach()` function.

tests/github-glue.test.ts
dscho: So this does not work for you?

```js
const github = new GitHub(`../${repo}`, repo);
const octo: Octokit = github["client"];
```

tests/github-glue.test.ts
dscho: Hmm. I fail to find a good wording. "Sibling directory"? "must live in the same directory as the current worktree"? I'm happy with neither.

webstech: Adding `(ie ../)` to help clarify.

tests/github-glue.test.ts
webstech: Changing to set with `beforeAll()`.  They only need to be set once.

tests/github-glue.test.ts
webstech: Okay, it works when I disable `no-string-literal` but that leads to other issues.  The `GitHubProxy` provides two functions.

1. It provides access to the authenticated `Octokit` api token for functions not in `GitHubGlue`.
2. It provides a method to `authenticate` using the protected `ensureAuthenticated()` method with a different owner.

The alternative to using the proxy would be to make `ensureAuthenticated()` public.

The private field could be removed as you suggested but that still leaves the authentication issue.

dscho: Ah, okay, that makes sense!

tests/github-glue.test.ts
dscho: Excellent!

dscho:
@webstech thank you for your patient explanations!

Issue #157, opened 2019-11-22
Make sure that we're not tracking commits of closed PRs


dscho:
The `update-commit-mappings` command in the `Update GitGitGadget's PRs` job takes quite a long time. I suspect that it looks at commits of PRs that were merged long ago.

webstech:
I verified the openPRs list in the options only has the open PRs (<60).  In `CIHelper::updateCommitMappings`, `commitsInPu` has well over 40k entries that will get searched a number of times.  The range-diff for pull #15 (the oldest one still open) took some time (20 sec) and returned over 4k entries to run through a reg-ex.  While the performance of my box may not match the azure server it does appear that some of the actions in `update-commit-mappings` could take some time.

dscho:
Right, and I don't think that we necessarily need to monitor the commits of already-merged branches.

We will still want to map responses on the Git mailing list even to closed PRs, but we can do that relatively cheaply because of the way we look up Message-IDs (we create pseudo blobs for them and then look up their Git notes, if any). That is independent of the ever-growing list of commitsInPu. I think we'll need to find a way to cull that list regularly.

dscho:
Still an issue. See e.g. https://dev.azure.com/gitgitgadget/git/_build/results?buildId=37528&view=logs&j=92a4755a-6173-591e-5c07-8dbe1165e937&t=05dc4a6b-2706-5ea6-77c8-e14f00112193

Probably should stop trying to map commits already in `master`...

dscho:
It seems that with #231 merged, the "update commit mappings" step only takes about 3.5 minutes (see e.g. [here](https://dev.azure.com/gitgitgadget/git/_build/results?buildId=44261&view=logs&j=fd490c07-0b22-5182-fac9-6d67fe1e939b)), down from ~21 minutes (see [the last run without #231 "branch-deployed"](https://dev.azure.com/gitgitgadget/git/_build/results?buildId=43393&view=logs&j=fd490c07-0b22-5182-fac9-6d67fe1e939b)).

So I declare victory on this ticket.

PR #231, opened 2020-05-08
Be smarter about updating commit mappings

dscho:
The `Update commit mappings` step of https://dev.azure.com/gitgitgadget/git/_build?definitionId=2&_a=summary takes a painfully long time: some 20 minutes.

The reason is that it runs a potentially costly `range-diff` for every single open PR on git/git and gitgitgadget/git, and `range-diff` is inherently costly because it compares every diff in the source range to every diff in the target range (and we use `<baseCommit>..pu` as target range, which can contain thousands of commits).

Let's coalesce the base commits and the head commits (by generating throw-away octopus merges with no ref associated with them) so that we can do a single `range-diff`. This will still be expensive, but not as expensive as running several dozens of them.

For technical reasons (a PR might be based on `pu`), that `range-diff` cannot use the same base for the second commit range. We'll be using `maint~100..pu` instead. Technically, this is a change of behavior, but a good one: we really do not need to look at commits that have been merged into `maint` _long_ ago: a previous run would have already seen those commits.

This addresses https://github.com/gitgitgadget/gitgitgadget/issues/157.

### DIFF
--- a/lib/ci-helper.ts
+++ b/lib/ci-helper.ts
@@ -1,7 +1,7 @@
 import * as fs from "fs";
 import * as util from "util";
 import { ILintError, LintCommit } from "./commit-lint";
-import { commitExists, git } from "./git";
+import { commitExists, git, emptyTreeName } from "./git";
 import { GitNotes } from "./git-notes";
 import { GitGitGadget, IGitGitGadgetOptions } from "./gitgitgadget";
 import { GitHubGlue, IGitHubUser, IPRCommit,
@@ -159,6 +159,14 @@ export class CIHelper {
                        { workDir: this.workDir })).split("\n"),
         );
         let result = false;
+        /*
+         * Both `bases` and `heads` accumulate the `-p<commit-hash>` parameters
+         * for the `git commit-tree` command for the two octopus merges. We
+         * need to make sure that no parent is listed twice, as `git
+         * commit-tree` would error out on that.
+         */
+        const bases = new Set<string>();
+        const heads = new Set<string>();
         for (const pullRequestURL of Object.keys(options.openPRs)) {
             const info = await this.getPRMetadata(pullRequestURL);
             if (info === undefined || info.latestTag === undefined ||
@@ -183,16 +191,36 @@ export class CIHelper {
                 meta.commitInGitGit = undefined;
                 result = true;
             }
+            bases.add(`-p${info.baseCommit}`);
+            heads.add(`-p${info.headCommit}`);
+        }
 
+        if (heads.size > 0) {
+            /*
+             * Generate throw-away octopus merges to combine multiple commit
+             * ranges into a single one.
+             */
+            const octopus = async (set: Set<string>): Promise<string> => {
+                const array = Array.from(set);
+                if (array.length === 1) {
+                    return array[0];
+                }
+                return await
+                    git(["commit-tree", ...array, emptyTreeName, "-m", "()"],
+                        { workDir: this.workDir });
+            };
+
+            const range1 = `${await octopus(bases)}..${await octopus(heads)}`;
+            const range2 =
+                "refs/remotes/upstream/maint~100..refs/remotes/upstream/pu";
             const start = Date.now();
             const out = await git(["-c", "core.abbrev=40", "range-diff", "-s",
-                                   info.baseCommit, info.headCommit,
-                                   "refs/remotes/upstream/pu"],
+                                   range1, range2],
                                   { workDir: this.workDir });
             const duration = Date.now() - start;
             if (duration > 2000)
                 console.log(`warning: \`git range-diff ${
-                    info.baseCommit} ${info.headCommit} upstream/pu\` took ${
+                    range1} ${range2}\` took ${
                     duration / 1000} seconds`);
             for (const line of out.split("\n")) {
                 const match =
@@ -200,7 +228,7 @@ export class CIHelper {
                 if (!match) {
                     continue;
                 }
-                const messageID2 = match[1] === info.headCommit ? messageID :
+                const messageID2 =
                     await this.getMessageIdForOriginalCommit(match[1]);
                 if (messageID2 === undefined) {
                     continue;
--- a/lib/ci-helper.ts
+++ b/lib/ci-helper.ts
@@ -111,7 +111,7 @@ export class CIHelper {
                 await this.mail2commit.getGitGitCommitForMessageId(messageID);
         }
         if (!upstreamCommit || upstreamCommit === mailMeta.commitInGitGit) {
-                return false;
+            return false;
         }
         mailMeta.commitInGitGit = upstreamCommit;
         if (!mailMeta.originalCommit) {
--- a/lib/ci-helper.ts
+++ b/lib/ci-helper.ts
@@ -136,6 +136,7 @@ export class CIHelper {
     public async updateCommitMappings(): Promise<boolean> {
         if (!this.gggNotesUpdated) {
             await git(["fetch", "https://github.com/gitgitgadget/git",
+                       "--tags",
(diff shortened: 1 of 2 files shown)


lib/ci-helper.ts
```diff
@@ -183,24 +185,43 @@ export class CIHelper {
...
+             * Generate throw-away octopus merges to combine multiple commit
+             * ranges into a single one.
+             */
+            const octopus = async (set: Set<string>): Promise<string> => {
+                const array = Array.from(set);
+                if (array.length === 1) {
+                    return array[0];
+                }
+                return await git(["commit-tree", ...array, emptyTreeName],
+                                 { workDir: this.workDir, stdin: "()" });
```
webstech, line 203: Just curious - why use stdin instead of -m command line parm?

lib/ci-helper.ts
```diff
@@ -159,6 +159,8 @@ export class CIHelper {
                        { workDir: this.workDir })).split("\n"),
         );
         let result = false;
+        const bases = new Set<string>();
+        const heads = new Set<string>();
```
webstech, line 163: Does `git commit-tree` reject duplicate `-p commit_sha` parameters?  If not, is there a need to create a set that is converted to an array?

lib/ci-helper.ts
dscho: Oh shucks. That `-m` option is [right there](https://git-scm.com/docs/git-commit-tree#Documentation/git-commit-tree.txt--mltmessagegt). I thought I remembered that the `git commit-tree` command did _not_ accept `-m`... Will fix.

lib/ci-helper.ts
dscho: It errors out when it sees those, indeed.

dscho: I added a code comment, see [the interdiff](https://github.com/gitgitgadget/gitgitgadget/compare/c07a3158f20fa90f5418ebce6eeb1d871f61e1f1..baf6d52011de4cd83f2b7176d0971ebc4d1cf3bf).

lib/ci-helper.ts
dscho: Fixed.

dscho:
@webstech do you think it'd be okay to merge this in its current form?

dscho:
FWIW I edited [the build definition](https://dev.azure.com/gitgitgadget/git/_build?definitionId=2&_a=summary) to use this here PR's changes, to see how it fares.

webstech:
> @webstech do you think it'd be okay to merge this in its current form?

I can't disagree.  I was wondering if there is clean-up needed for the commit-tree commands. Was also wondering if this works on a fresh clone with the limits on rev2.  Mostly my lack of knowledge in this area of git and lack of time to run tests.

dscho:
>  I was wondering if there is clean-up needed for the commit-tree commands.

My plan was to leave all this to the regular `git gc --auto` runs triggered by various Git operations.

> Was also wondering if this works on a fresh clone with the limits on rev2.

Do you mean the layout used by lore.kernel.org as opposed to the layout of public-inbox.org? If so, that's orthogonal to what this here PR does: it uses the clone of https://github.com/gitgitgadget/git to do its work. And yes, things should totally work on a fresh clone, especially now that the tags are fetched explicitly (they are needed to make sure that we have all of the commits corresponding to the GitGitGadget PRs).

lib/ci-helper.ts
```diff
@@ -182,18 +191,44 @@ export class CIHelper {
...
+                    range1} ${range2}\` took ${
+                    duration / 1000} seconds`);
             for (const line of out.split("\n")) {
                 const match =
                     line.match(/^[^:]*: *([^ ]*) [!=][^:]*: *([^ ]*)/);
                 if (!match) {
                     continue;
                 }
-                const messageID2 = match[1] === info.headCommit ? messageID :
+                const messageID2: string | undefined =
```
webstech, line 231: The type declaration could be removed (not present on line 178).

webstech:
Trying to understand things.  What is triggered by adding the check completion (`annotateCommit`) to the pull request?

dscho:
> What is triggered by adding the check completion (`annotateCommit`) to the pull request?

This:

image

It's not quite _as_ nice as I would like it (mouse-over with a beautiful link to the upstream commit would have been nicer), but hey, I take what I can get ;-)

lib/ci-helper.ts
dscho: You're right!

dscho:
> I approve but have two possibly new issues unrelated to this change.  If they are real issues, I can open as such.
> 
> 1.  The check is showing as ongoing and not as completed (ie a timer is still running).  Looking at earlier PRs, DCO checks also show this but now show as completed so the api call may not be quite what is required.  Thoughts?

This is unfortunately a known issue that we cannot do anything about: sometimes, the (internal) notifications seem to get lost and the build finishes without updating the PR.

I had the impression that it got better as of recent?

> 2.  The commit in git.git noted in the check message is not also added as a comment in the PR.  It may be useful there.  If you don't know about the check message, the commit can be found by following the merge commit parent but that  is not obvious.  What do you think?

I fear that would add too much chatter, no?

webstech:
> > 1. The check is showing as ongoing and not as completed (ie a timer is still running).  Looking at earlier PRs, DCO checks also show this but now show as completed so the api call may not be quite what is required.  Thoughts?
> 
> This is unfortunately a known issue that we cannot do anything about: sometimes, the (internal) notifications seem to get lost and the build finishes without updating the PR.

Researching this a little more.  The `checks.create()` call sets `completed_at` to the commit time.

```curl https://api.github.com/repos/gitgitgadget/git/check-runs/29019955 -H "Accept: application/vnd.github.antiope-preview+json"```

shows
```
  "status": "completed",
  "conclusion": "success",
  "started_at": "2018-11-05T20:56:12Z",
  "completed_at": "2018-10-11T05:12:45Z",
```
So, is the problem caused by the completed timestamp being before the started timestamp?

> 
> I had the impression that it got better as of recent?
> 
> > 1. The commit in git.git noted in the check message is not also added as a comment in the PR.  It may be useful there.  If you don't know about the check message, the commit can be found by following the merge commit parent but that  is not obvious.  What do you think?
> 
> I fear that would add too much chatter, no?

Could it be added at the time the PR is being closed?

**Comment by dscho** [comment]
> Could it be added at the time the PR is being closed?

Sure, but what does it buy us? There is already a message mentioning the upstream merge commit...

**Comment by dscho** [comment]
> So, is the problem caused by the completed timestamp being before the started timestamp?

Sorry for not answering this earlier: I have no idea. But then, I don't think that we actually set those dates...

**Comment by webstech** [comment]
> Sorry for not answering this earlier: I have no idea. But then, I don't think that we actually set those dates...

We set completed but not started.
```
        const checks = await this.client.checks.create({
            completed_at: completedAt,
```
I will try to run a test on a repo to see if `started_at === completed_at` is allowed.  If not, I will set it slightly earlier so it is always less.  If you would prefer a specific value for started_at, let me know.  Just want the check to show a completed time and not a timer.

webstech:
> I will try to run a test on a repo to see if `started_at === completed_at` is allowed. If not, I will set it slightly earlier so it is always less. If you would prefer a specific value for started_at, let me know. Just want the check to show a completed time and not a timer.

@dscho Confirmed started_at and completed_at can be the same value.  I will update `GitHubGlue`.

dscho:
> > I will try to run a test on a repo to see if `started_at === completed_at` is allowed. If not, I will set it slightly earlier so it is always less. If you would prefer a specific value for started_at, let me know. Just want the check to show a completed time and not a timer.
> 
> @dscho Confirmed started_at and completed_at can be the same value.  I will update `GitHubGlue`.

Great!

PR #241, opened 2020-06-06
Update jest and ts-jest to v26+

webstech:
Jest added improved typescript support that requires more specific imports.

The two commits in the patch series must be used together and not cherry picked.

This resolves the problems found with PR #240.


dscho:
Wonderful, thank you!

PR #242, opened 2020-06-06
GitHubGlue specify started_at for check-run create

webstech:
Checks on GitHub will now appear to be completed.  Previously
the checks showed a 'running elapsed' timer when the completion
timestamp preceded the started timestamp.

The started_at and completed_at values are now the same.


dscho:
Thank you, as always!

PR #248, opened 2020-06-09
Update @types/html-to-text to 5.1.1

webstech:
The types are updated to support the format attribute so a local instance of HtmlToTextOptions is now used.

In `blockquote(),` `indentOptions.wordwrap` is provided in the target for Object.assign to overcome `wordwrap` being optional in the interface (and a couple of possible types).  It is provided in `formatOptions` but the compiler can not know that is the instance that will be passed to `blockquote()`.

### DIFF
--- a/lib/markdown-renderer.ts
+++ b/lib/markdown-renderer.ts
@@ -1,41 +1,33 @@
-import { fromString } from "html-to-text";
+import { fromString, HtmlToTextOptions } from "html-to-text";
 import marked from "marked";
 
-// @types/html-to-text v1.4.31 does not handle the `format` attribute; To work
-// around that, we simply implement a class that has that (we want a different
-// rendering of headings than html-to-text provides by default).
+// Provide our own renderings of headings and blockquotes.
 
-class MyOptions implements HtmlToTextOptions {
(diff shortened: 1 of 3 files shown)


derrickstolee:
@webstech: Could you provide an example of how a PR description or commit message improves with this change? What is the before and after, in the formatted patch? That would help me review this a bit more easily. Thanks!

Even better: describe the new behavior in a test so we can be sure the options work as intended.

webstech:
> @webstech: Could you provide an example of how a PR description or commit message improves with this change? What is the before and after, in the formatted patch? That would help me review this a bit more easily. Thanks!
> 
> Even better: describe the new behavior in a test so we can be sure the options work as intended.

There should be no change.  PR #243 was an automated update that broke gitgitgadget - it required changes to gitgitgadget and the package change was incomplete.  The incomplete package was updated and this PR includes the appropriate gitgitgadget changes.  The older version of `@types/html-to-text` was out of date and `markdown-renderer.ts` had code to work around the stale code.  This patch removes the work around.  The unit tests were not updated because there was no expected change in the output.

Sorry, I should have referenced the PR this (indirectly) supersedes.  Does this help clarify things?

PR #264, opened 2020-06-25
Allow unquoted keys in gitgitgadget.cismtpopts in config

webstech:
Quotes will now be added to gitgitgadget.cismtpopts JSON keys as needed.
Quoting can be difficult to set in automation where env variables may
be used to issue git config commands.

A possible GitHub [action](https://github.com/webstech/gitgitgadget/blob/workflow/.github/workflows/dev-test.yml) may need the cismtopts set as a secret.  The windows cmd environment has problems setting the config.

### DIFF
--- a/lib/send-mail.ts
+++ b/lib/send-mail.ts
@@ -159,7 +159,10 @@ export async function sendMail(mail: IParsedMBox,
     };
 
     if (smtpOptions.smtpOpts) {
-        Object.entries(JSON.parse(smtpOptions.smtpOpts))
+        // Add quoting for JSON.parse
+        const smtpOpts = smtpOptions.smtpOpts
+            .replace(/([ {])([a-zA-Z0-9.]+?) *?:/g,"$1\"$2\":");
+        Object.entries(JSON.parse(smtpOpts))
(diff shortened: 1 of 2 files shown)


dscho:
> A possible GitHub [action](https://github.com/webstech/gitgitgadget/blob/workflow/.github/workflows/dev-test.yml) may need the cismtopts set as a secret. The windows cmd environment has problems setting the config.

Great!

Please note that GitHub seems to call these things "GitHub workflows", and the building blocks of those workflows are "GitHub Actions".

webstech:
> Please note that GitHub seems to call these things "GitHub workflows", and the building blocks of those workflows are "GitHub Actions".

"GitHub Actions" (capitalized) appears to be the product name.  Yes it is about workflows, which have jobs which contain "actions" or tasks/steps.  I was referring to an action.

@dscho would you like the workflow added to gitgitgadget?  It will run on both linux and windows to ensure compatibility.  Should a secret be needed to enable it so only devs who want it get it?

dscho:
I'd love to have that workflow, and I don't think we need to install any secret.

PR #269, opened 2020-06-30
Add GitHub Workflow for Testing of Developer Push

webstech:
Run linux and windows build/test on developer push.

The workflow will run a job on each OS.  Secrets can be passed to set the git config values that some of the tests check for.

### DIFF
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -1,0 +1,79 @@
+# Run tests against developer push
+
+# Runs jobs on linux and windows to maintain compatibility.
+# The main repo is checked out into a sub-directory because an optional
+# second repo may be checked out for testing.
+
+# Optional GitHub secrets
+
+# See ci-helper.test.ts for information on the following
+# GGG_SMTP_USER = 'gitgitgadget.cismtpuser=<email_userid>'
+# GGG_SMTP_PASS = 'gitgitgadget.cismtppass=<email_password>'
+# GGG_SMTP_HOST = 'gitgitgadget.cismtphost=<email_hostname>'
+# GGG_SMTP_OPTS = 'gitgitgadget.cismtpopts=<email_smtp_options>'
+
+# See github-glue.test.ts for information on the following
+# GGG_TOKEN = token to access the test repo
+# GGG_REPOSITORY = name of test repo to update in tests.  Tests will
+# open a PR, create comments, etc.  This can be a private or public
+# repo.
+
+name: pr-test
+
+on:
+  push:
+    branches-ignore: [ master, main ]
+
+jobs:
+  build-test:
+    name: Node v${{ matrix.node-version }} on ${{ matrix.os }}
+
+    strategy:
+      fail-fast: false
+      matrix:
+        node-version: [14.x]
+        os: [ubuntu-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
+
+    # job level env vars used for a couple of steps
+    env:
+      # see github-glue.test.ts for info on setting these
+      GGG_TOKEN: ${{ secrets.GGG_TOKEN }}
+      GGG_REPOSITORY: ${{ secrets.GGG_REPOSITORY }}
+
+    steps:
+    # Check out repo under sub-dir so other repo can be checked out
+    - uses: actions/checkout@v2
+      with:
+        path: gitgitgadget
+
+    # Check out github-glue.test.ts repo if configured for it
+    - uses: actions/checkout@v2
+      with:
+        repository: ${{ github.repository_owner }}/${{ env.GGG_REPOSITORY }}
+        token: ${{ env.GGG_TOKEN }}
+        path: ${{ env.GGG_REPOSITORY }}
+      if: env.GGG_TOKEN
+
+    - uses: actions/setup-node@v1.4.2
+      with:
+        node-version: '${{ matrix.node-version }}' # optional, default is 10.x
+
+    - name: Install packages
+      run: npm ci
+      working-directory: gitgitgadget
+
+    - name: Set git test repo config
+      shell: bash
+      run: echo "::set-env name=GGG_REPO::'gitgitgadget.githubtest.githubuser=$GITHUB_REPOSITORY_OWNER' 'gitgitgadget.$GITHUB_REPOSITORY_OWNER.githubrepo=$GGG_REPOSITORY' 'gitgitgadget.$GITHUB_REPOSITORY_OWNER.githubtoken=$GGG_TOKEN'"
+      if: env.GGG_TOKEN
+
+    - name: Run build
+      run: npm run build
+      working-directory: gitgitgadget
+
+    - name: Run tests
+      env:
+        GIT_CONFIG_PARAMETERS: ${{ secrets.GGG_SMTP_USER }} ${{ secrets.GGG_SMTP_PASS }} ${{ secrets.GGG_SMTP_HOST }} ${{ secrets.GGG_SMTP_OPTS }} ${{ env.GGG_REPO }}
+      run: npm run test
+      working-directory: gitgitgadget


.github/workflows/dev-test.yml
```diff
@@ -0,0 +1,140 @@
...
+      run: |
+        git config --add gitgitgadget.cismtpuser %SMTPUSER%
+        git config --add gitgitgadget.cismtphost %SMTPHOST%
+        git config --add gitgitgadget.cismtpopts "%SMTPOPTS%"
+        git config --add gitgitgadget.cismtppass %SMTPPASS%
+      if: env.SMTPUSER
+      working-directory: gitgitgadget
+
+    - name: Set git test repo config
+      shell: cmd
```
dscho, line 126: How about using `bash` here, and then turning it into a [matrix workflow](https://www.edwardthomson.com/blog/github_actions_6_fail_fast_matrix_workflows.html)?

.github/workflows/dev-test.yml
```diff
@@ -0,0 +1,140 @@
+# Run tests against developer push
+
+# Runs jobs on linux and windows to maintain compatibility.
+# The main repo is checked out into a sub-directory because an optional
+# second repo may be checked out for testing.
+
+name: dev-test
+
+on:
+  push:
```
dscho, line 10: Wouldn't we want this to run also on PRs?

.github/workflows/dev-test.yml
```diff
@@ -0,0 +1,140 @@
...
+    branches-ignore: [ master, main ]
```
dscho, line 11: Why ignore `main`?

.github/workflows/dev-test.yml
```diff
@@ -0,0 +1,140 @@
...
+      with:
+        path: gitgitgadget
+
+    # Check out github-glue.test.ts repo if configured for it
+    - uses: actions/checkout@v2
+      with:
+        repository: ${{ github.repository_owner }}/${{ env.GGG_REPOSITORY }}
+        token: ${{ env.GGG_TOKEN }}
+        path: ${{ env.GGG_REPOSITORY }}
+      if: env.GGG_TOKEN
```
dscho, line 35: Hmm. I'd rather move that repository to the `gitgitgadget` org, then make this step contingent on `${{ github.repository_owner }} == 'gitgitgadget'`  and clone it always, using a shell step if necessary.

.github/workflows/dev-test.yml
```diff
@@ -0,0 +1,84 @@
...
+
+    - uses: actions/setup-node@v1.4.2
+      with:
+        node-version: '14.x' # optional, default is 10.x
```
dscho, line 46:
```suggestion
        node-version: '${{ matrix.node-version }}' # optional, default is 10.x
```

.github/workflows/dev-test.yml
dscho: Oh! I think I understand now. This is a fork of `gitgitgadget/git` for testing purposes?

.github/workflows/dev-test.yml
```diff
@@ -0,0 +1,84 @@
...
+        SMTPPASS: ${{ secrets.SMTPPASS }}
+        SMTPHOST: ${{ secrets.SMTPHOST }}
+        SMTPOPTS: ${{ secrets.SMTPOPTS }}
+      run: |
+        git config --add gitgitgadget.cismtpuser $SMTPUSER
+        git config --add gitgitgadget.cismtphost $SMTPHOST
+        git config --add gitgitgadget.cismtpopts "$SMTPOPTS"
+        git config --add gitgitgadget.cismtppass $SMTPPASS
+      if: env.SMTPUSER
+      working-directory: gitgitgadget
```
dscho, line 66: I wonder whether it might make more sense to combine these into a value ready for `GIT_CONFIG_PARAMETERS` (which is the environment variable that backs the `git -c <var>=<value>` feature: try for yourself in a Bash: `GIT_CONFIG_PARAMETERS="'chris.handle=webster'" git config chris.handle`). It could then be used _just_ in the "Run tests" job below, in the `env` section:

```yaml
    env:
        GIT_CONFIG_PARAMETERS: ${{ secrets.GGG_SMTP_CONFIG_PARAMETERS }}
```

The secret would then look somewhat like this:

```
'gitgitgadget.cismtpuser=hello-world' 'gitgitgadget.cismtphost=mail.example.com' 'gitgitgadget.cismtppass=xoxo'
```

What do you think?

.github/workflows/dev-test.yml
webstech: Sure.  Not sure why I didn't.

.github/workflows/dev-test.yml
webstech: I was initially setting up for ensuring developer pushes got tested on both windows and linux (I am still a windows cmd user but the azure pipelines are all linux).  If you want to switch the PR check build to here that would be great.  In that case, it should probably be renamed.  I am open to suggestions (ci-test.yml?).

.github/workflows/dev-test.yml
webstech: My workflow, which may not be correct, is to rebase my main branch from upstream after PRs have been merged.  Since we (I) don't develop on main, it seemed like a waste to run the workflow when the main was pushed to github.  I am open to change or suggestions to take another look at some doc if this is not the way to do things. 🙄

.github/workflows/dev-test.yml
webstech: It is a repo for testing purposes.  It does not have to be a fork.  I have [webstech/gitout](https://github.com/webstech/gitout) set up for this.  The github-glue.test.ts uses it to open a PR, create comments, etc.  Since it can be a private repo, a PAT is needed.  A new github repo could be created/deleted but using an existing one was easier and faster.  gitgitgadget would need one for CI tests if we don't add the create ability.

.github/workflows/dev-test.yml
webstech: Good catch.

.github/workflows/dev-test.yml
webstech: It works but updating secrets is a replace and not an edit so updating the password would be a real pain.  How about:

```
    env:
        GIT_CONFIG_PARAMETERS: ${{ secrets.GGG_SMTP_USER secrets.GGG_SMTP_PASS secrets.GGG_SMTP_HOST secrets.GGG_SMTP_OPTS }}
```
Then each secret would be in the format you you showed.  This would not prevent you from stuffing everything into one secret and not setting the other three.

Maybe the `Set git test repo config` step can just set a var to use here as well.
```
echo "::set-env name=GGG_REPO::'gitgitgadget.githubtest.githubuser=$GITHUB_REPOSITORY_OWNER' 'gitgitgadget.$GITHUB_REPOSITORY_OWNER.githubrepo=$GGG_REPOSITORY' 'gitgitgadget.$GITHUB_REPOSITORY_OWNER.githubtoken=$GGG_TOKEN'"
```
This avoids running `git config` but still requires a step.  Your thoughts?

.github/workflows/dev-test.yml
dscho: That's good! Maybe we can add a YAML comment at the top to explain this?

.github/workflows/dev-test.yml
dscho: Okay. I guess we already have the merge commit tested when merging PRs, so we do not _strictly_ need a CI build.

.github/workflows/dev-test.yml
dscho: `pr-test.yml`?

.github/workflows/dev-test.yml
dscho: Between all the options we have on the table so far, I prefer the "secret `GIT_CONFIG_PARAMETERS` and your current solution as a close runner-up. Since the latter does not require any additional work from you, let's just keep it as-is.

.github/workflows/dev-test.yml
webstech: > That's good! Maybe we can add a YAML comment at the top to explain this?

Renamed and comments added.

.github/workflows/pr-test.yml
```diff
@@ -0,0 +1,79 @@
...
+# The main repo is checked out into a sub-directory because an optional
+# second repo may be checked out for testing.
+
+# Optional GitHub secrets
+
+# See ci-helper.test.ts for information on the following
+# SMTP_USER = 'gitgitgadget.cismtpuser=<email_userid>'
+# SMTP_PASS = 'gitgitgadget.cismtpuser=<email_password>'
+# SMTP_HOST = 'gitgitgadget.cismtpuser=<email_hostname>'
+# SMTP_OPTS = 'gitgitgadget.cismtpuser=<email_smtp_options>'
```
dscho, line 13: Two questions:
1. should we prefix those with `GGG_`, too?
2. Does it really make sense to require the secrets to contain all that non-secret stuff? I am starting to regret suggesting the `GIT_CONFIG_PARAMETERS` route...

webstech: > 1. should we prefix those with `GGG_`, too?

I was going to suggest renaming the others but I suppose keeping them in a namespace is better to avoid conflicts if future vars are added by GitHub.

> 2. Does it really make sense to require the secrets to contain all that non-secret stuff? 

They are mostly secrets.  The `SMTP_OPTS` and `GGG_REPOSITORY` may not need to be.  These values are unique to each person who wants the tests run.

> I am starting to regret suggesting the `GIT_CONFIG_PARAMETERS` route...

Would you prefer I revert to the two git config steps?

webstech:
It appears git config/options do not allow leading whitespace in `GIT_CONFIG_PARAMETERS`.  While that may be desirable, a PR will be opened to pass in a dummy config parm.  Jobs with no secrets are [failing](https://github.com/gitgitgadget/gitgitgadget/pull/298/checks?check_run_id=992283517).

@dscho Any interest in having git handle leading whitespace in the env var?

dscho:
> @dscho Any interest in having git handle leading whitespace in the env var?

The `GIT_CONFIG_PARAMETERS` variable is actually intended to be internal to Git, so what we do here is basically a hack and we have to play by Git's rules, i.e. _not_ have leading whitespace.

In other words, I think we will have to convert https://github.com/gitgitgadget/gitgitgadget/blob/e36bd5e0674bb1902eb55ce154e0db4b80456306/.github/workflows/pr-test.yml#L77 so that it looks more like https://github.com/gitgitgadget/gitgitgadget/blob/e36bd5e0674bb1902eb55ce154e0db4b80456306/.github/workflows/pr-test.yml#L68

webstech:
> we have to play by Git's rules, i.e. _not_ have leading whitespace.

I opened pr #302 so a leading value is always present in `GIT_CONFIG_PARAMETERS`.  Otherwise there is still the problem of having a space between the two sets - needed if the first set is valued and not needed if the first set is not valued.  I'm not seeing a simple way to handle it (not saying there isn't one - it is just not obvious to me).

Issue #274, opened 2020-07-15
Include fetch instructions in the "Submitted as pull.<num>.git.*" message


phil-blain:
It would be nice if when the bot answers to a `/submit` command with the link to the submitted version on the list,
instructions for how to fetch that version locally would also be included (similar to the `Fetch-It-Via` footer of the cover letter in the mailing list submission).

I was thinking something along the lines of 
> To fetch this version into `FETCH_HEAD`:
>
>     git fetch https://github.com/gitgitgadget/git pr-<num>/<github_username>/<branch-name>-v<x>
>
> To fetch this version to a local tag:
>
>     git fetch https://github.com/gitgitgadget/git tag pr-<num>/<github_username>/<branch-name>-v<x> --no-tags
>

---

I wish the second command could be simply `git fetch <ggg url> tag <tag>`, but either the [`git fetch tag <tag>`](https://git-scm.com/docs/git-fetch#Documentation/git-fetch.txt-ltrefspecgt) or the [`-no-tags`](https://git-scm.com/docs/git-fetch#Documentation/git-fetch.txt---no-tags) documentation is wrong, [or the doc is correct but the code is buggy...](https://lore.kernel.org/git/12F0A2A4-CC71-4452-B8B8-F540339F8519@gmail.com/T/#u)

dscho:
> I wish the second command could be simply `git fetch <ggg url> tag <tag>`

I figure you meant only `<tag>` not `tag <tag>`. And yes, according to @peff's feedback, that works, right?

@phil-blain If you want to give it a try, you will probably want to touch [these lines](https://github.com/gitgitgadget/gitgitgadget/blob/e8ec25f755cc0b770ba956e2932f45b4d0fe0e03/lib/ci-helper.ts#L605-L607).

phil-blain:
Yes it works correctly, the documentation for `--no-tags` could need a little clarification. I really meant `tag <tag>`, i.e. not have to also add `--no-tags`. (I guess `git fetch <tag>` would work only if a refspec is already configured for tags, which is not the case "out-of-the-box"...)

Anyway, thanks for the pointer.

webstech:
> @phil-blain If you want to give it a try, you will probably want to touch [these lines](https://github.com/gitgitgadget/gitgitgadget/blob/e8ec25f755cc0b770ba956e2932f45b4d0fe0e03/lib/ci-helper.ts#L605-L607).

[...]

PR #295, opened 2020-08-12
Report The Patch Tag on Submit Command

webstech:
Provide easy to copy git fetch instructions for fetching the
submitted version of the patches. The comment will now include:

To fetch this version into `FETCH_HEAD`:
```
git fetch https://github.com/gitgitgadget/git <ggg-tag-name>
```

To fetch this version to local tag <ggg-tag-name>:
```
git fetch --no-tags https://github.com/gitgitgadget/git tag <ggg-tag-name>
```
This addresses issue #274.

### DIFF
--- a/lib/ci-helper.ts
+++ b/lib/ci-helper.ts
@@ -601,9 +601,19 @@ export class CIHelper {
                         `\n\nWARNING: ${comment.author
                         } has no public email address set on GitHub` : "";
 
-                    const coverMid = await gitGitGadget.submit(pr, userInfo);
+                    const metadata = await gitGitGadget.submit(pr, userInfo);
+                    const uri = "https://github.com/gitgitgadget/git";
+                    const code = "\n```";
                     await addComment(`Submitted as [${
-                        coverMid}](https://lore.kernel.org/git/${coverMid})${
+                            metadata?.coverLetterMessageId
+                            }](https://lore.kernel.org/git/${
+                            metadata?.coverLetterMessageId
+                            })\n\nTo fetch this version into \`FETCH_HEAD\`:${
+                            code}\ngit fetch ${uri} ${metadata?.latestTag}${code
+                            }\n\nTo fetch this version to local tag \`${
+                            metadata?.latestTag}\`:${
+                            code}\ngit fetch --no-tags ${
+                            uri} tag ${metadata?.latestTag}${code}${
                             extraComment}`);
                 }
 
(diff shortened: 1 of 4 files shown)


dscho:
> `git fetch --no-tags https://github.com/gitgitgadget/git tag <ggg-tag-name>`

Is that bare `tag` really intended? Did you maybe mean `<ggg-tag-name>:<ggg-tag-name>` instead?

webstech:
> > `git fetch --no-tags https://github.com/gitgitgadget/git tag <ggg-tag-name>`
> 
> Is that bare `tag` really intended? Did you maybe mean `<ggg-tag-name>:<ggg-tag-name>` instead?

@phil-blain Do you have a comment on this?  

I thought the two syntax were somewhat equivalent.  Is `tag` a newer or older syntax?  If older, then I would be inclined to switch to `<tag>:<tag>` - the most current and prevalent format should be suggested.

dscho:
> I thought the two syntax were somewhat equivalent. Is `tag` a newer or older syntax?

I would expect `git fetch <url> tag abc` to try to fetch two refs, one named `tag` and one named `abc`. Is that not so?

webstech:
> I would expect `git fetch <url> tag abc` to try to fetch two refs, one named `tag` and one named `abc`. Is that not so?

Am I misunderstanding the [doc](https://github.com/gitgitgadget/git/blob/878e727637ec5815ccb3301eb994a54df95b21b8/Documentation/pull-fetch-param.txt#L33-L34)?

```
`tag <tag>` means the same as `refs/tags/<tag>:refs/tags/<tag>`;
it requests fetching everything up to the given tag.
```

Looks like the syntax has been around for awhile but if the usage is not very common, it can be changed to `tag:tag` as you suggested.

dscho:
> Am I misunderstanding the [doc](https://github.com/gitgitgadget/git/blob/878e727637ec5815ccb3301eb994a54df95b21b8/Documentation/pull-fetch-param.txt#L33-L34)?
> 
> ```
> `tag <tag>` means the same as `refs/tags/<tag>:refs/tags/<tag>`;
> it requests fetching everything up to the given tag.
> ```

Whoa! I learned something new today... 

> Looks like the syntax has been around for awhile

Actually, it has been around for _quite a while_: 5fca669f1933ca3ef5f93ee2bf8871183bc2a7c2

> but if the usage is not very common, it can be changed to `tag:tag` as you suggested.

Nah, I think you're right, since that syntax is available, we should use it.

phil-blain:
@webstech thanks for putting this in. I was offline last week. I indeed myself learned of the `tag <tag>` syntax just before I opened issue #274.

Issue #228, opened 2020-05-05
Name and address on the From: header of e-mail coming from GGG is misleading


gitster:
It seems that GGG places <gitgitgadget@gmail.com> as the sender address and "Commit Author Name via GitGitGadget" as the human-readable name for the sender.

This has a few drawbacks.

 - Because the real commit author name is never with "via GitGitGadget", this arrangement forces all patches to have in-body From: that records the real author name. This redundancy does not add any value to reviewers, as we humans know how to strip out "via GitGitGadget" anyway.

 - When sending out a multi-patch series with multiple authors, each patch is sent with "From: " header taken from the commit author name. There is *no* clue, other than in the cover letter, that tells who created a pull request to trigger GGG,

Can we fix these by doing a few simple changes, perhaps?

 - Record the user who did the pull request on the "From:" header of cover letter and patches in the series.
 - Keep adding in-body From: as necessary to record the commit author name when it is different from the user who created the pull request (whose name and address is recorded on the "From:" header of the e-mail).
 - Use "Sender:" to record the fact that the message is sent from <gitgitgadget@gmail.com> instead.  Add the same on "Cc:" if needed to ensure that responses will be seen by GGG.

See https://lore.kernel.org/git/pull.622.git.1588347029.gitgitgadget@gmail.com/ for an example of a topic that has multiple authors.

Thanks.

dscho:
Are we sure that such a mail would not be filtered out by vger and/or subscribers of the Git mailing list? That is, and was, my biggest concern when I came up with the `via` strategy.

gitster:
I manually added "Sender: (redacted)@pobox.com" to send out https://lore.kernel.org/git/xmqqpnbg6gfv.fsf_-_@gitster.c.googlers.com/raw and it seems to have reached lore.kernel.org/git/ archive.

gitster:
Any progress on this?

dscho:
> Any progress on this?

Sorry, no, I am swamped and short on time :-(

webstech:
There is an edge case where the author of the pull request does not have a public email on github and is not the author or contributor to any of the patches.  The sender email would be unknown.  In that case, how should the `From:` be done?  Should the existing `via GitGitGadget` format be used (with ggg email) or just show the name with no email address?  I think I prefer the latter but could it be a problem for some mail agents?  There would still be the `Sender:` header that some agents show as  `From:`.

[...]

PR #296, opened 2020-08-12
Clarify `From:` and add `Sender:` Email fields

webstech:
For GGG, record the user who did the pull request on the "From:" header of cover letter and patches in the series.  Do not add the user to the `Cc:` list.

Only add in-body From: as necessary to record the commit author name when it is different from the user who created the pull request.  Do not re-add the commit author name if already in the `Cc:` list.

Use "Sender:" to record the fact that the message is sent from gitgitgadget@gmail.com.

This addresses issue #228. 

This initial PR includes comments to be removed (they may or may not be helpful).

Are more unit tests needed to verify the handling of `cc` and in body `from` lines?

### DIFF
--- a/lib/patch-series.ts
+++ b/lib/patch-series.ts
@@ -207,11 +207,6 @@ export class PatchSeries {
                                                wrapCoverLetterAt,
                                                indentCoverLetter);
 
-        // if known, add submitter to email chain
-        if (senderEmail) {
-            cc.push(`${senderName} <${senderEmail}>`);
-        }
-
         if (basedOn && !await revParse(basedOn, workDir)) {
             throw new Error(`Cannot find base branch ${basedOn}`);
         }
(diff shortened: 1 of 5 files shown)


dscho:
For the record: sadly, I had to revert this via #301, as all the `From:` headers listed GitGitGadget's email address in the end, _not_ the intended commit author's or the PR submitter's. See https://github.com/gitgitgadget/gitgitgadget/pull/301#issuecomment-674787307 for some analysis from my side.

webstech:
> I have to admit that I could not _completely_ wrap my head around the part touching `patch-series.ts` but the tests verify that it does what it is supposed to do, and I'm willing to just give it a go.

If you agree to require an email for /submit to work, that code can be simpler.  The changes avoided double adding to the cc (already in the cc) and more handling of the possible \n in the encoded name.

I will research some more.  Another possibility is finding another smtp server.

I would like to get some of this implemented even if the sender part can not be done.

dscho:
> I will research some more.

Thank you!

> Another possibility is finding another smtp server.

Indeed. I do not know how to find one, though: the most common two options are GMail and Outlook, and neither allow this (Outlook does not even allow the "ABC via ..." thing, it _always_ rewrites the `From:` header to match the records).

PR #297, opened 2020-08-16
Add a Cc: for replies on the Git mailing list

webstech:
When processing email replies to the Git mailing list the cc: footer will be updated to include people who were not already on the list.  gitster@pobox.com will be excluded from the list.

A comment is added to indicate the cc: list has been updated.

**Refactor mbox Parsing in handle-new-mails**

Function parseMBox was called from both mail-archive-helper processMails() and send-mail parseMBoxMessageIDAndReferences(). processMails() will now call parseMBoxMessageIDAndReferences() with a parsed mail.

**Add GitHubGlue::updatePR() to Update Title or Body**

Tests are also included.

### DIFF
--- a/lib/github-glue.ts
+++ b/lib/github-glue.ts
@@ -92,6 +92,33 @@ export class GitHubGlue {
     }
 
     /**
+     * Add a cc to a Pull Request
+     *
+     * @param {string} pullRequestURL the Pull Request to comment on
+     * @param {string} cc to add
+     * @returns the comment ID and the URL to the comment
+     */
+    public async addPRCc(pullRequestURL: string, cc: string):
+        Promise<void> {
+        const id = cc.match(/(<.*>)/);
+
+        if (!id || id[1].match(/gitster@pobox\.com/)) {
+            return;
+        }
+
+        const url = GitGitGadget.parsePullRequestURL(pullRequestURL);
+        const pr = await this.getPRInfo(url[0], url[2]);
+        const ccNow = pr.body.match(/\n\n(cc:.*)/is);
+
+        if (!ccNow || !ccNow[1].match(id[1])) {
+            await this.updatePR(url[0], url[2], `${pr.body}${
+                ccNow ? "" : "\n"}\ncc: ${cc}`);
+            await this.addPRComment(pullRequestURL, `User \`${
+                                    cc}\` has been added to the cc: list.`);
+        }
+    }
+
+    /**
      * Add a Pull Request comment
      *
      * @param {string} pullRequestURL the Pull Request to comment on
--- a/lib/github-glue.ts
+++ b/lib/github-glue.ts
@@ -180,6 +180,30 @@ export class GitHubGlue {
         };
     }
 
+    /**
+     * Update a Pull Request body or title
+     *
+     * @param {string} pullRequestURL the Pull Request to comment on
+     * @param {string} body the updated body
+     * @param {string} title the updated title
+     * @returns the PR number
+     */
+    public async updatePR(owner: string, prNumber: number,
+                          body?: string | undefined, title?: string):
+        Promise<number> {
+
+        await this.ensureAuthenticated(owner);
+        const result = await this.client.pulls.update({
+            "body": body || undefined,
+            owner,
+            pull_number: prNumber,
+            repo: this.repo,
+            "title": title || undefined,
+        });
+
+        return result.data.id;
+    }
+
     public async setPRLabels(pullRequestURL: string, labels: string[]):
         Promise<string[]> {
         const [owner, repo, prNo] =
--- a/lib/mail-archive-helper.ts
+++ b/lib/mail-archive-helper.ts
@@ -157,6 +157,9 @@ export class MailArchiveGitHelper {
                         .addPRComment(pullRequestURL, comment);
                 }
 
+                await this.githubGlue.addPRCc(pullRequestURL,
+                                              parsedMbox.from || "");
+
                 await this.gggNotes.set(parsed.messageID, {
                     issueCommentId,
                     messageID: parsed.messageID,
--- a/lib/mail-archive-helper.ts
+++ b/lib/mail-archive-helper.ts
@@ -88,16 +88,21 @@ export class MailArchiveGitHelper {
             return keys.has(MailArchiveGitHelper.hashKey(messageID));
         };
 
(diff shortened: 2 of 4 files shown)


lib/send-mail.ts
```diff
@@ -117,7 +113,7 @@ export async function parseMBoxMessageIDAndReferences(mbox: string):
      */
     const msgIdRegex =
         /^\s*<([^>]+)>(\s*|,)(\([^")]*("[^"]*")?\)\s*|\([^)]*\)$)?(<.*)?$/;
-    for (const header of parsed.headers) {
+    for (const header of parsed.headers ?? []) {
```
dscho, line 116: I guess `??` means that `[]` is substituted only if `parsed.headers` is `undefined`, while a single `?` would also substitute if `parsed.headers` is already the empty array?

lib/github-glue.ts
```diff
@@ -91,6 +91,32 @@ export class GitHubGlue {
...
     }
 
+    /**
+     * Add a cc to a Pull Request
+     *
+     * @param {string} pullRequestURL the Pull Request to comment on
+     * @param {string} cc to add
+     * @returns the comment ID and the URL to the comment
+     */
+    public async addPRcc(pullRequestURL: string, cc: string):
```
dscho, line 101: Technically, I would write this as `Cc` (i.e. with an upper-case `C` to indicate that this is a new CamelCased "word"). Not sure whether you want to change it? I'd be fine with leaving it as-is, but I figured I'd mention that I stumbled over it.

lib/github-glue.ts
```diff
@@ -91,6 +91,32 @@ export class GitHubGlue {
...
+        Promise<void> {
+        const id = cc.match(/(<.*>)/);
+
+        if (id) {                   // filter out gitster@pobox.com?
```
dscho, line 105: Good idea. @gitster mentioned a couple times that he wants to be Cc:ed only when a patch series is considered "ready" (which is kind of a fuzzy topic, as I am sure that most people who open PRs on GitHub think that their patches are already ready 😀).

lib/github-glue.ts
```diff
@@ -91,6 +91,32 @@ export class GitHubGlue {
...
+            const url = GitGitGadget.parsePullRequestURL(pullRequestURL);
+            const pr = await this.getPRInfo(url[0], url[2]);
+            const ccNow = pr.body.match(/\n\n([Cc][Cc]:.*)/s);
```
dscho, line 108: Does the `i` attribute work? It seems to work over here:

```sh
$ node -e 'console.log("Cc:".match(/cc/i));'
[ 'Cc', index: 0, input: 'Cc:', groups: undefined ]
```

lib/github-glue.ts
dscho: Oh, and I am a fan of early returns. That way, no brain power needs to be expended on the question "but what do we do in this function if `!id`?". I.e. something like this instead:

```js
if (!id) {
    return;
}
```

lib/mail-archive-helper.ts
```diff
@@ -157,6 +157,9 @@ export class MailArchiveGitHelper {
                         .addPRComment(pullRequestURL, comment);
                 }
 
+                await this.githubGlue.addPRcc(pullRequestURL,
+                                              parsedMbox.from || "");
```
dscho, line 161: Maybe we should guard this more like this:

```js
        if (parsedMbox.from) {
                await this.githubGlue.addPRcc(pullRequestURL,
                                              parsedMbox.from);
        }
```

But then, we return without doing anything from `addPRcc()` if no email address was in that value, so I'd be fine with leaving this as-is.

dscho (approved):
I left a couple of suggestions, but would not mind if you wanted to merge as-is.

lib/send-mail.ts
webstech: A single `?` sends the compiler into a conditional operator search state that errors out with `error TS1005: ':' expected.`.  The `??` gets past this.  Either I found an answer (SO?) or I just added `?` until the compiler was happy, knowing the only current caller of the code errors out if the headers are not found (as did the old code in this function).  The ts syntax is what it is and maybe using a local to reference the possibly not present array is cleaner.

lib/github-glue.ts
webstech: Updating.

lib/github-glue.ts
webstech: I cannot explain why I did not use it.  Updating.

lib/github-glue.ts
webstech: > Good idea. @gitster mentioned a couple times that he wants to be Cc:ed only when a patch series is considered "ready" (which is kind of a fuzzy topic, as I am sure that most people who open PRs on GitHub think that their patches are already ready 😀).

That comment/question was kind of a conversation starter I forgot I was starting.  The question really is, since `git@vger.kernel.org` is usually in the `to:` list, should replies from `gitster@pobox.com` not be added to the cc list?  Is he getting double emails?

webstech: > Oh, and I am a fan of early returns. 

In this case, there seems little need with the function size.  In general we probably disagree on this (probably based on experience).  Early returns cause the compiler to generate destructors for any objects constructed earlier on the stack.  As an assembler coder that was very visible as code bloat.  On the other hand, I have seen some ugly techniques employed to enforce single entry/exit.

lib/mail-archive-helper.ts
webstech: The chance of an email without a `from:` seems pretty low so will leave the code as is.

lib/github-glue.ts
dscho: > > Oh, and I am a fan of early returns.
> 
> In this case, there seems little need with the function size.

I did have to look for the end of the block and determine whether anything else happens there (not currently). My idea here is not about the efficiency of the generated code, but rather the readability for _me_ ;-) In general, I am happy whenever I can avoid levels of indentation, and an early return over here would definitely help me next time I read that code.

dscho (commented):
Excluding @gitster specifically, and doing the early return, are still things I would like to see. I like the rest of this PR.

webstech:
> Excluding @gitster specifically, and doing the early return, are still things I would like to see. I like the rest of this PR.

Done and a test was added to verify the filtering on gitster.

dscho (approved):
Very cool, thanks!

PR #305, opened 2020-08-25
Handle missing secrets in github push workflow

webstech:
Git does not allow leading spaces in the `GIT_CONFIG_PARAMETERS` environment variable.  This corrects that for when no secrets are defined for the workflow.  Trailing spaces are allowed.

Additionally, the workflow will now get an error if only some of the secrets in a group (there are two) are defined.

The format of the smtp related secrets are now just the basic values and not the internal formats.

### DIFF
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -7,10 +7,10 @@
 # Optional GitHub secrets
 
 # See ci-helper.test.ts for information on the following
-# GGG_SMTP_USER = 'gitgitgadget.cismtpuser=<email_userid>'
-# GGG_SMTP_PASS = 'gitgitgadget.cismtppass=<email_password>'
-# GGG_SMTP_HOST = 'gitgitgadget.cismtphost=<email_hostname>'
-# GGG_SMTP_OPTS = 'gitgitgadget.cismtpopts=<email_smtp_options>'
+# GGG_SMTP_USER = <email_userid>
+# GGG_SMTP_PASS = <email_password>
+# GGG_SMTP_HOST = <email_hostname>
+# GGG_SMTP_OPTS = <email_smtp_options> This is a json object.
 
 # See github-glue.test.ts for information on the following
 # GGG_TOKEN = token to access the test repo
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-test.yml
```diff
@@ -37,16 +37,50 @@ jobs:
...
+      env:
+        GGG_SMTP_OPTS: ${{ secrets.GGG_SMTP_OPTS }}
+      shell: bash
+      run: |
+        if [ -n "$GGG_SMTP_USER" ] && [ -n "$GGG_SMTP_PASS" ] && [ -n "$GGG_SMTP_HOST" ];
+        then
+          m="'gitgitgadget.cismtpuser=$GGG_SMTP_USER' 'gitgitgadget.cismtppass=$GGG_SMTP_PASS' 'gitgitgadget.cismtphost=$GGG_SMTP_HOST'";
+          if [ -n "$GGG_SMTP_OPTS" ];
+          then
+            m+=" 'gitgitgadget.cismtpopts=$GGG_SMTP_OPTS'"
```
dscho, line 74: `+=` is a Bash-ism, but since we explicitly ask for Bash, that's okay.

.github/workflows/pr-test.yml
```diff
@@ -37,16 +37,50 @@ jobs:
...
 
+    - name: Set git test repo config
+      shell: bash
+      run: |
+        if [ -n "$GGG_TOKEN" ] && [ -n "$GGG_REPOSITORY" ];
+        then
+          echo "::set-env name=GGG_REPO::'gitgitgadget.githubtest.githubuser=$GITHUB_REPOSITORY_OWNER' 'gitgitgadget.$GITHUB_REPOSITORY_OWNER.githubrepo=$GGG_REPOSITORY' 'gitgitgadget.$GITHUB_REPOSITORY_OWNER.githubtoken=$GGG_TOKEN'"
+        else
+          echo "::error::Both GGG_TOKEN and GGG_REPOSITORY secrets must be set"
+          exit 1
```
dscho, line 60: I guess it does not hurt much to have an `exit 1` here, but I would have expected the `echo ::error::...` to mark this task as failed, and consequently all subsequent tasks to be skipped anyway.

dscho (approved):
While I would have preferred the more succinct style (`test -z "$<name>" || ...`), my preference is not worth holding up this PR.

dscho:
Thanks!

webstech:
> While I would have preferred the more succinct style (`test -z "$<name>" || ...`), my preference is not worth holding up this PR.

Thanks for the suggestion.  Like perl (and the numerous (semi) obsolete scripting languages still rattling around in my head) there is more than one way to do things.  There is a wide variety of styles in the scripts in git.  Will look around for some recommendations on current best practices next time.

dscho:
> Will look around for some recommendations on current best practices next time.

Oh, there are none... We make them up on the go, and since you're the most active developer of GitGitGadget now, you get to make the call 😁

Issue #153, opened 2019-11-20
Teach GitGitGadget to guess the full name and email address of new contributors


dscho:
Since new contributors usually contribute their own code, we can make use of the REST API to [get a single commit](https://developer.github.com/v3/repos/commits/#get-a-single-commit) that has enough information to guess the full name and email address of a contributor.

[Example](https://api.github.com/repos/dscho/git/commits/add-i-in-c-status-and-help) produces this:

```json
{
  "sha": "ccc6fb45580e1c07b143d1a55d9518c13ac3c032",
  "node_id": "MDY6Q29tbWl0MjMyNjA0NzpjY2M2ZmI0NTU4MGUxYzA3YjE0M2QxYTU1ZDk1MThjMTNhYzNjMDMy",
  "commit": {
    "author": {
      "name": "Slavica Đukić",
      "email": "slawica92@hotmail.com",
      "date": "2019-02-14T10:41:46Z"
    },
    "committer": {
      "name": "Johannes Schindelin",
      "email": "johannes.schindelin@gmx.de",
      "date": "2019-11-14T14:35:15Z"
    },
[...]
  "author": {
    "login": "slavicaDj",
    "id": 26170660,
    "node_id": "MDQ6VXNlcjI2MTcwNjYw",
    "avatar_url": "https://avatars0.githubusercontent.com/u/26170660?v=4",
    "gravatar_id": "",
    "url": "https://api.github.com/users/slavicaDj",
    "html_url": "https://github.com/slavicaDj",
    "followers_url": "https://api.github.com/users/slavicaDj/followers",
    "following_url": "https://api.github.com/users/slavicaDj/following{/other_user}",
    "gists_url": "https://api.github.com/users/slavicaDj/gists{/gist_id}",
    "starred_url": "https://api.github.com/users/slavicaDj/starred{/owner}{/repo}",
    "subscriptions_url": "https://api.github.com/users/slavicaDj/subscriptions",
    "organizations_url": "https://api.github.com/users/slavicaDj/orgs",
    "repos_url": "https://api.github.com/users/slavicaDj/repos",
    "events_url": "https://api.github.com/users/slavicaDj/events{/privacy}",
    "received_events_url": "https://api.github.com/users/slavicaDj/received_events",
    "type": "User",
    "site_admin": false
  },
  "committer": {
    "login": "dscho",
    "id": 127790,
    "node_id": "MDQ6VXNlcjEyNzc5MA==",
[...]

PR #306, opened 2020-08-27
Add /cc command

webstech:
The value passed may be one or more email addresses.  The format is the same as for the cc: footer in the PR body.

### DIFF
--- a/azure-function/index.js
+++ b/azure-function/index.js
@@ -139,7 +139,7 @@ module.exports = async (context, req) => {
             }
 
             /* Only trigger the Pipeline for valid commands */
-            if (!comment.body || !comment.body.match(/^\/(submit|preview|allow|disallow|test)\b/)) {
+            if (!comment.body || !comment.body.match(/^\/(submit|preview|allow|disallow|test|cc)\b/)) {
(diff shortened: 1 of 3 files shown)


dscho (changes requested):
Very nice! Just one thing, we need to adjust the Azure Function, too: https://github.com/gitgitgadget/gitgitgadget/blob/b3ef98529be5bbd61ec4fb9ead26919316eec019/azure-function/index.js#L142

webstech:
> Very nice! Just one thing, we need to adjust the Azure Function, too:

You had mentioned it and I still managed to miss it.  

Should the /cc command  be documented somewhere?

dscho (approved):

dscho:
Thank you!

dscho:
> Should the /cc command be documented somewhere?

Maybe ;-)

While we're thinking about `/cc`, I would also _love_ to have the option to, say, `/cc @webstech` (i.e. auto-map GitHub handles to email addresses, which might require having an internal map for users without public email addresses, or even https://github.com/gitgitgadget/gitgitgadget/issues/153 to be resolved)...

Issue #310, opened 2020-09-02
Automatic CC feature duplicates email addresses


phil-blain:
In https://github.com/gitgitgadget/git/pull/714, both Peff and Brian replied twice to my patch, and the new GGG automatic CC feature that adds a "cc:" footer to the PR description added their email twice.

Also, probably the PR author themself need not be added to the CC list if they answer to their own series.

@dscho @webstech

dscho:
> the new GGG automatic CC feature that adds a "cc:" footer to the PR description added their email twice.

Hmm. It should avoid that, of course... And I thought that https://github.com/gitgitgadget/gitgitgadget/pull/297/files#diff-be774741832b606f27bf194081322cedR111-R113 made it so? @phil-blain if you have time to investigate, that would be nice...

> Also, probably the PR author themself need not be added to the CC list if they answer to their own series.

Right. However, I do not really know that we have an easy way to look that up? Your GitHub profile does not list a public email, for example, so I don't really know how GitGitGadget would know that a certain email address refers to you...

webstech:
One of the edits of the PR description converted all `\n` to `\r\n` so the original `cc:` was not found.  This caused the second set of `cc:` to be added.

phil-blain:
I think that's GitHub following web standard conventions to use `\r\n` in text fields...

webstech:
Sorry about that.  I had got used to GGG converting them to `\n` internally.

webstech:
> > Also, probably the PR author themself need not be added to the CC list if they answer to their own series.
> 
> Right. However, I do not really know that we have an easy way to look that up? Your GitHub profile does not list a public email, for example, so I don't really know how GitGitGadget would know that a certain email address refers to you...

PR #323 corrects this in the case where the GitHub profile has a matching public email.

dscho:
Shouldn't we already start with this internal mapping of GitHub handles to full email information? I am thinking of extending the allowed users list from a mere array to a map (or _is_ it already a map? *clicketyclick* [nope, it's still an array](https://github.com/gitgitgadget/gitgitgadget/blob/345128bb9a7e9d7c87b4a03f2290c98aa9b88c33/lib/gitgitgadget.ts#L13)).

dscho:
I guess this has been addressed 😉

PR #311, opened 2020-09-02
Change to use crlf for new lines when adding cc

webstech:
GitHub PR body uses `\r\n` for newline.  That convention will now be followed when updating the `cc:` list.

Fixes the duplicate `cc:` problem in issue #310.


dscho (approved):
Awesome, thanks for the fast work!

PR #323, opened 2020-09-25
Do not add PR owner to CC list

webstech:
A previous change added email reply authors to the cc list.  The owner of the PR will no longer be added to the list because they are already added internally.

Users with no public email on GitHub will be added to the list.


dscho (approved):
Excellent!

PR #339, opened 2020-10-27
Convert to eslint from tslint

webstech:
Many generic types are now more specific.

The linting will now be based on eslint and prettier.  Both with typescript extensions.

There are some ignore rules (file and line level) to be cleaned up.  One file is being ignored at the moment.  Issues will be opened after this is approved to ensure they are not missed.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,1 @@
+**/script/mail-patch-series.ts
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,169 @@
+module.exports = {
+    "env": {
+        "es6": true,
+        node: true,
+    },
+    "extends": [
+        "plugin:@typescript-eslint/recommended",
+        "plugin:@typescript-eslint/recommended-requiring-type-checking",
+        "prettier",
+        "prettier/@typescript-eslint"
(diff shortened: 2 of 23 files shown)


dscho:
Wow, this is a lot... I'll try to find some time today to review this properly.

dscho:
@webstech I tried my hand at a couple of alternative changes (and at breaking up the lot of small changes into lots of commits). See #341 for details.

Unfortunately, the diff relative to your `eslint` branch is rather big, not only because of the `package-lock.json` changes (which seem to always disagree heavily between machines) but also because I addressed some spelling fixes while at it (for me, VS Code's cSpell would otherwise report problems, and I really wanted to bring the problem count down to zero).

What do you think?

webstech:
> breaking up the lot of small changes into lots of commits

Still wrapping my head around the process where all commits must be accepted.  I would have broken it up but still in a mindset that all the parts must be one.

> `package-lock.json` changes (which seem to always disagree heavily between machines) 

That is a diff I no longer look at.  Might be more consistent if I did a clean install.

> What do you think?

I have left comments.  I will close this pull request.  Thanks.  I have learned things in your changes.

PR #342, opened 2020-10-31
Resolve eslint unbound method error

webstech:
Removing the need to specify this override:
`eslint-disable-next-line @typescript-eslint/unbound-method`

The code will now use `jest.spyOn()`, which provides a cleaner implementation of swapping out the `console.log()` function.


dscho (approved):
Good find, that `spyOn()` thing! Very nice.

PR #349, opened 2020-11-07
CVE-2020-15228 fix to set-env and other workflow updates

webstech:
Workflows will now use environment files to set variables.  See [CVE-2020-15228](https://github.com/advisories/GHSA-mfwh-5m23-j46w) for background information.

Updated the `branches-ignore` to match current documentation.

Change from a specific release to the more generic version which should get the latest code.

This gets rid of the warning:
`The 'add-path' command is deprecated and will be disabled soon.`

### DIFF
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -22,7 +22,9 @@ name: pr-test
 
 on:
   push:
-    branches-ignore: [ master, main ]
+    branches-ignore:
+      - master
+      - main
 
 jobs:
   build-test:
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -91,7 +91,7 @@ jobs:
         path: ${{ env.GGG_REPOSITORY }}
       if: env.GGG_TOKEN
 
-    - uses: actions/setup-node@v1.4.2
+    - uses: actions/setup-node@v1
       with:
         node-version: '${{ matrix.node-version }}' # optional, default is 10.x
 
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -54,7 +54,7 @@ jobs:
       run: |
         if [ -n "$GGG_TOKEN" ] && [ -n "$GGG_REPOSITORY" ];
         then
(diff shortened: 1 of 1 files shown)


dscho (dismissed):

dscho:
Hmm. I don't have a clue why the `pr-test` workflow didn't run... do you, @webstech?

webstech:
> Hmm. I don't have a clue why the `pr-test` workflow didn't run... do you, @webstech?

I do not.  Then again, it wasn't intended to run, since the Azure pipeline will run.  I updated the workflow to match the current doc.  It no longer shows the `[master, main]` syntax (assuming I originally got it from the doc and not some random blog).  Even if the syntax was incorrect, I am not sure how a `push` event was triggered.

After pushing the update, the workflow ran on my repo fine (as it did yesterday when I determined not all the tests were working).  The check status in the PR did not change so I can't tell if it failed again or did not run - the pipeline did run.

webstech:
Added another commit to get rid of a warning annotation but still no change in the checks status.

.github/workflows/pr-test.yml
```diff
@@ -22,7 +22,9 @@ name: pr-test
 
 on:
   push:
-    branches-ignore: [ master, main ]
+    branches-ignore:
+      - master
+      - main
```
dscho, line 27: Wait, why does this even run in other PRs? Because those PRs are opened by dependabot and the topic branches are pushed directly into gitgitgadget/gitgitgadget!

So it is my mistake: I should not have made those checks mandatory for PRs...

dscho (approved):

.github/workflows/pr-test.yml
webstech: > So it is my mistake: I should not have made those checks mandatory for PRs...

I do not understand what you are saying.  This workflow is for push requests.  How could it be done for PRs?

dscho: Indeed, it is for pushes only! So it _won't work_ to make this a required check to pass for PRs. Yet that's exactly what I had done.

It _happened_ to work for the dependabot PRs because dependabot pushes directly into gitgitgadget/gitgitgadget. Therefore, there was a push, and there was a check that passed.

But it did not work for your PR because you pushed to webstech/gitgitgadget, and the check from over there is not seen by the PR checks in gitgitgadget/gitgitgadget.

I fixed that by not requiring those checks anymore.

webstech: > I fixed that by not requiring those checks anymore.

This is the part I was missing.  A search showed me where to find the required checks settings.  I was looking for rules in all the wrong places (or something like that).

PR #361, opened 2020-11-18
CI: verify emails have been received

webstech:
If configured, an email imap session will verify the sent test emails have been received.  This requires, in addition to existing environment variables, gitgitgadget.CIimapHost to be set.

For GitHub workflows, a new secret GGG_IMAP_HOST must be set.

This feature waits one second after sending the email to check for it.  If this is found to be insufficient (tests failing for not found), retry logic can be added.

This has been tested with the ethereal email service.

For local testing, if the `Error: self signed certificate in certificate chain` message is issued, set environment variable `NODE_TLS_REJECT_UNAUTHORIZED=0`.

### DIFF
--- a/tests/ci-helper.test.ts
+++ b/tests/ci-helper.test.ts
@@ -6,7 +6,10 @@ import {
      GitHubGlue, IGitHubUser, IPRComment, IPRCommit, IPullRequestInfo,
 } from "../lib/github-glue";
 import { IMailMetadata } from "../lib/mail-metadata";
+import { connect, ImapSimple } from "imap-simple";
 import { testCreateRepo, TestRepo } from "./test-lib";
+import { promisify } from "util";
+const sleep = promisify(setTimeout);
 
 jest.setTimeout(180000);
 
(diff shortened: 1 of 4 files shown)


tests/ci-helper.test.ts
```diff
@@ -120,7 +125,7 @@ async function setupRepos(instance: string):
     ]);
 
     const { smtpUser, smtpHost, smtpPass, smtpOpts } =
```
dscho, line 127: Doesn't this need to be changed to this?

```js
     const { smtpUser, smtpHost, , smtpPass, smtpOpts } =
```

dscho: Ah, I guess not, because `{}` actually means a map, not a tuple.

tests/ci-helper.test.ts
```diff
@@ -158,6 +163,75 @@ async function setupRepos(instance: string):
...
+
+    return connection;
+}
+
+/**
+ * Terminate the imap mail server connection.
+ *
+ * @param connection
+ */
+async function imapDisConnect(connection: ImapSimple): Promise <void> {
```
dscho, line 194:
```suggestion
async function imapDisconnect(connection: ImapSimple): Promise <void> {
```

dscho (approved):
Looks very good!

I have just that one suggestion: `s/DisConnect/Disconnect/`.

tests/ci-helper.test.ts
webstech: Sorry, not what I would have done from scratch.  Changed.

dscho (approved):

dscho:
Thank you!!!

PR #362, opened 2020-11-19
Improve checking when adding to PR cc: list

webstech:
The checking for existing entries in the cc: list will now:

- allow a non-empty line (whitespace) before the footers
- ignore case for both cc: entries and PR author email
- support a body with only a footer
- check only email and not names in cc: entries

The code assumes an empty line (may have whitespace) separates the body from the footers and no empty lines are in the footers.

Tests were added to validate the new checks.

This resolves a problem seen in [776](https://github.com/gitgitgadget/git/pull/776) where users were copied multiple times and [783](https://github.com/gitgitgadget/git/pull/783) where the PR author was copied.

Additionally this change will simplify GitHubGlue add CC testing:

Mock functions are now used for the tests.  This removes the need to update the actual PR on GitHub, which is significantly faster.

### DIFF
--- a/lib/github-glue.ts
+++ b/lib/github-glue.ts
@@ -1,3 +1,4 @@
+import addressparser = require("nodemailer/lib/addressparser");
 import { Octokit } from "@octokit/rest";
 import { git, gitConfig } from "./git";
 import { GitGitGadget } from "./gitgitgadget";
@@ -100,21 +101,52 @@ export class GitHubGlue {
      */
     public async addPRCc(pullRequestURL: string, cc: string):
         Promise<void> {
-        const id = cc.match(/(<.*>)/);
+        const id = cc.match(/<(.*)>/);
 
-        if (!id || id[1].match(/gitster@pobox\.com/)) {
+        if (!id || id[1] === "gitster@pobox.com") {
             return;
         }
 
+        const ccLower = id[1].toLowerCase();
         const url = GitGitGadget.parsePullRequestURL(pullRequestURL);
         const pr = await this.getPRInfo(url[0], url[2]);
-        const ccNow = pr.body.match(/\r?\n\r?\n(cc:.*)/is);
+        const trimBody = pr.body.trimRight();
+        const bodyEnd = trimBody.split(/\r?\n\s*?\r?\n/).pop() as string;
 
-        if (!ccNow || !ccNow[1].match(id[1])) {
+        let found = false;
+        let footerSeparator = "\r\n";
+
+        if (bodyEnd.match(/:/)) try {
+            bodyEnd.split(/\r?\n/).reverse().forEach(line => {
+                const match = line.match(/^([a-z][-a-z0-9]+):\s*(.*)$/i);
+
+                if (!match) {       // stop if not a footer
+                    throw new Error("No Footer");
+                }
+
+                footerSeparator = ""; // body already has footers
+                if (!found && match[1].toLowerCase() === "cc") try {
+                    addressparser(match[2], {flatten: true}).forEach(email => {
+                        if (ccLower === email.address.toLowerCase()) {
+                            found = true;
+                            throw new Error("Found");
+                        }
+                    });
+                } catch {
+                    // quick exit for cc matched (comment to quiet linter)
+                }
+            });
+        } catch {
+            found = false;          // ensure it was not a cc: false positive
+            footerSeparator = "\r\n"; // reset
+        }
+
+        if (!found) {
             const user = await this.getGitHubUserInfo(pr.author);
-            if (!user.email || !user.email.match(id[1])) {
+
+            if (!user.email || ccLower !== user.email.toLowerCase()) {
                 await this.updatePR(url[0], url[2], `${pr.body}${
-                    ccNow ? "" : "\r\n"}\r\ncc: ${cc}`);
+                    footerSeparator}\r\ncc: ${cc}`);
                 await this.addPRComment(pullRequestURL, `User \`${
                                         cc}\` has been added to the cc: list.`);
             }
--- a/tests/github-glue.test.ts
+++ b/tests/github-glue.test.ts
@@ -174,23 +174,97 @@ test("pull requests", async () => {
         const prCc2 = "Not Real <RealNot@saturn.cosmos>";
         const prCcGitster = "Git Real <gitster@pobox.com>"; // filtered out
         const ghUser = await github.getGitHubUserInfo(owner);
+        const ccRegex = /<ReallyNot@saturn\.cosmos>/g;
 
+        // Test with no linefeed
+        const prBodyCc1 = `${prInfo.body}`;
+        await github.updatePR(owner, prData.number, prBodyCc1);
         await github.addPRCc(prData.html_url, prCc);
         await github.addPRCc(prData.html_url, prCc);
+        const prCcInfo1 = await github.getPRInfo(owner, prData.number);
+        const ccFound1 = prCcInfo1.body.match(ccRegex);
+        expect(ccFound1?.length).toEqual(1);
+
+        // Test with linefeeds present
+        const prBodyCc2 = `${prInfo.body}\r\n\r\nGlue`;
+        await github.updatePR(owner, prData.number, prBodyCc2);
+        await github.addPRCc(prData.html_url, prCc);
+        await github.addPRCc(prData.html_url, prCc);
+        await github.addPRCc(prData.html_url, prCc.toLowerCase());
         await github.addPRCc(prData.html_url, prCc2);
         await github.addPRCc(prData.html_url, prCcGitster);
         if (ghUser.email) {
             await github.addPRCc(prData.html_url, `PR Owner <${ghUser.email}>`);
+            await github.addPRCc(prData.html_url,
+                `PR Owner <${ghUser.email.toUpperCase()}>`);
         }
-        const prCcInfo = await github.getPRInfo(owner, prData.number);
-        expect(prCcInfo.body).toMatch(prCc);
-        const ccFound = prCcInfo.body.match(prCc);
-        expect(ccFound?.length).toEqual(1);
-        expect(prCcInfo.body).toMatch(prCc2);
-        expect(prCcInfo.body).not.toMatch(prCcGitster);
+        const prCcInfo2 = await github.getPRInfo(owner, prData.number);
+        const ccFound2 = prCcInfo1.body.match(ccRegex);
+        expect(ccFound2?.length).toEqual(1);
+        expect(prCcInfo2.body).toMatch(/RealNot@saturn\.cosmos/);
+        expect(prCcInfo2.body).not.toMatch(/gitster@pobox\.com/);
         if (ghUser.email) {
-            expect(prCcInfo.body).not.toMatch(ghUser.email);
+            expect(prCcInfo2.body).not.
+                toMatch(new RegExp(ghUser.email, "i"));
         }
+
+        // Test with linefeeds and unknown footers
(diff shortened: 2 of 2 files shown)


lib/github-glue.ts
```diff
@@ -108,7 +108,7 @@ export class GitHubGlue {
 
         const url = GitGitGadget.parsePullRequestURL(pullRequestURL);
         const pr = await this.getPRInfo(url[0], url[2]);
-        const ccNow = pr.body.match(/\r?\n\r?\n(cc:.*)/is);
+        const ccNow = pr.body.match(/\r?\n\s*\r?\n(cc:.*)/is);
```
dscho, line 111: Wouldn't this also match something like

```
In this patch series, we ensure that `git send-email` adds a
Cc: header whenever encountering even unknown footers,
as long as they are in the format "Some Name
<some@email.com>"
```

?

webstech: I was kind of hoping to avoid fully parsing the PR body.  In the referenced git PR, [this](https://github.com/gitgitgadget/gitgitgadget/blob/841ad8c6d483e0244fe0cd03c73778775c8e2c61/lib/patch-series.ts#L288) only works because there is a second paragraph in the body, causing much of the body to be treated as a footer (but added back).  `parsePullRequestBody` would also get hit by a line starting with `cc:` like your sample (not sure if the `Cc:` is really at the start of a line).

This change does not change the problem of unknown footers containing a matching email, causing the `cc:` to **not** be added.  I am contemplating a fuller fix so please just ignore for now.  I did notice a couple of matches that need to be made case insensitive.

lib/github-glue.ts
```diff
@@ -108,13 +108,34 @@ export class GitHubGlue {
...
+                    found = true;
+                    footerSeparator = "";
+                    break;
+                } else if (footer.match(/^\s*$/)) { // stop at first empty line
+                    break;
+                } else if (footer.match(/^[-A-Za-z]+:/)) { // body has footers
+                    footerSeparator = "";
+                }
+            }
+        }
```
dscho, line 131: I learned that Javascript often performs better, and is more readable, using a map/reduce approach. Something like this:

```js
let found = false, footersSeparator = "";
const footers = pr.body.split(/\r?\n\s*?\r?\n/).pop();
if (footers) try {
    footersSeparator = "\r\n";
    footers.split(/\r?\n/).reverse().forEach((footer) => {
        const match = footer.match(/^([a-z][-a-z0-9]*):\s*(.*)$/i);
        if (!match)
            throw new Error("No footers block");
        found ||= match[0].toLowerCase() === 'cc' &&
            (match[1].trimRight() === id[1] || match[1].search(`<${id[1]>`) >= 0);
    });
} catch {
    found = false;
    footersSeparator = "";
}

if (!found) {
    const user = await this.getGitHubUserInfo(pr.author);
    if (!user.email || user.email !== id[1]) {
        await this.updatePR(url[0], url[2], `${pr.body}${
            footersSeparator}\r\ncc: ${cc}`);
        await this.addPRComment(pullRequestURL, `User \`${
            cc}\` has been added to the cc: list.`);
    }
}
```

lib/github-glue.ts
```diff
@@ -108,13 +108,34 @@ export class GitHubGlue {
 
         const url = GitGitGadget.parsePullRequestURL(pullRequestURL);
         const pr = await this.getPRInfo(url[0], url[2]);
-        const ccNow = pr.body.match(/\r?\n\r?\n(cc:.*)/is);
+        const index = pr.body.search(/\r?\n\s*?\r?\n/);
```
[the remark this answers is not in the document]
webstech: Sure. The first still worked (loop stopped at an empty line) but this code is simpler.

lib/github-glue.ts
```diff
@@ -108,13 +108,34 @@ export class GitHubGlue {
...
+
+        let found = false;
+        let footerSeparator = "\r\n";
+        if (index > -1) {
+            const regExId = RegExp( `^cc:.*${id[1]}`, "i");
```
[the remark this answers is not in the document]
webstech: Switching to use `addressParser` to avoid other pitfalls such as `"me@example.com" <you@example.com>`.

lib/github-glue.ts
webstech: Old habits die hard - aversion to `try/catch` (in cpp) due to early performance issues.  Changing to use chaining as suggested - it was not used because I was avoiding try/catch.

Updated code will now complain if the last block in the body happens to contain what looks like footer lines.  For example:

```
some text

last block
todo: fix something
```

A test was added for this.

webstech: btw, your sample code has `footerSeparator` use reversed (for reference when reviewing code).  Also, checking on user.email is now case insensitive after seeing you added to the `cc:` in your own PR. 😀

lib/github-glue.ts
```diff
@@ -100,21 +101,52 @@ export class GitHubGlue {
      */
     public async addPRCc(pullRequestURL: string, cc: string):
         Promise<void> {
-        const id = cc.match(/(<.*>)/);
+        const id = cc.match(/<(.*)>/);
 
         if (!id || id[1].match(/gitster@pobox\.com/)) {
```
dscho, line 106: I guess at this stage it would make sense to change this to `id[1] === "gitster@pobox.com"`?

lib/github-glue.ts
```diff
@@ -100,21 +101,52 @@ export class GitHubGlue {
...
             return;
         }
 
+        const ccLower = id[1].toLowerCase();
         const url = GitGitGadget.parsePullRequestURL(pullRequestURL);
         const pr = await this.getPRInfo(url[0], url[2]);
-        const ccNow = pr.body.match(/\r?\n\r?\n(cc:.*)/is);
+        const bodyEnd = pr.body.split(/\r?\n\s*?\r?\n/).pop();
```
dscho, line 113: How certain are we that there is always an empty line? (I.e. is it possible for `pr.body` to contain just a single `Cc:` line?)

lib/github-glue.ts
```diff
@@ -100,21 +101,52 @@ export class GitHubGlue {
...
+                            found = true;
+                            throw new Error("Found");
+                        }
+                    });
+                }
+            });
+        } catch (e) {
+            const error = e as Error;
+            if (error.message === noFooters && footerSeparator === "") {
+                throw new Error("Footer contains unknown text");
```
dscho, line 140: I'd rather heed Postel's Law here. If, for some reason, the body ends in a paragraph whose last line starts with a word followed by a colon (which does _not_ need to be a footer!), we want to treat this as a missing footer. For example,

```
In this patch series, we do XYZ. If you wonder
why: Because ABC.
```

does _not_ have any footers.

lib/github-glue.ts
```diff
@@ -100,21 +101,52 @@ export class GitHubGlue {
...
+                if (!match) {       // stop if not a footer
+                    throw new Error(noFooters);
+                }
+
+                footerSeparator = ""; // body already has footers
+                if (match[1].toLowerCase() === "cc") {
+                    addressparser(match[2], {flatten: true}).forEach(email => {
+                        if (ccLower === email.address.toLowerCase()) {
+                            found = true;
+                            throw new Error("Found");
```
dscho, line 132: While it is quite unlikely for a last paragraph containing a line of the form `Cc: <exactly the address we're looking for>`, I would feel much better if we verified that we are really looking at footers, by verifying that every line begins with `<key>:`.

tests/github-glue.test.ts
```diff
@@ -175,23 +175,66 @@ test("pull requests", async () => {
...
 
+        // Test with no linefeed
+        const prBodyCc1 = `${prInfo.body}`;
+        await github.updatePR(owner, prData.number, prBodyCc1);
         await github.addPRCc(prData.html_url, prCc);
         await github.addPRCc(prData.html_url, prCc);
+
+        const prCcInfo1 = await github.getPRInfo(owner, prData.number);
+        expect(prCcInfo1.body).toMatch(prCc);
+        const ccFound1 = prCcInfo1.body.match(new RegExp( prCc, "g"));
```
dscho, line 186: There is an extra space before `prCc,`.

tests/github-glue.test.ts
```diff
@@ -175,23 +175,66 @@ test("pull requests", async () => {
...
+        const prBodyCC4 = `${prInfo.body}\r\n \t\r\nbb: ${prCc}`;
+        await github.updatePR(owner, prData.number, prBodyCC4);
+        await github.addPRCc(prData.html_url, prCc);
+        await github.addPRCc(prData.html_url, prCc);
+        const prCcInfo4 = await github.getPRInfo(owner, prData.number);
+        const ccFound4 = prCcInfo4.body.match(new RegExp( prCc, "g"));
+        expect(ccFound4?.length).toEqual(2);
+
+        // Test to catch last block in body sort of a footer
+        const prBodyCC5 = `${prInfo.body}\r\n\r\nfoo\r\nbb: bar`;
```
dscho, line 233: I might be missing something, but I don't think we look at ``prBodyN = `cc: ${prCc}`;`` nor at ``prBodyNPlus1 = `This is not a footer even if it contains\ncC: ${prCc}`;``.

dscho (commented):
Sorry to make it so difficult, but I think this still needs to be tweaked a bit more.

lib/github-glue.ts
webstech: Agreed.

webstech: Sure.  I had intended to do it and forgot in the morning.

lib/github-glue.ts
webstech: Split will always return a non-empty array, even if the `pr.body` is empty.  If there is no empty line, split will return one entry in the array with `pr.body`.

lib/github-glue.ts
webstech: Okay.  Was not sure which way to go.  Was thinking the error might catch someone who failed to add the empty line but now thinking this is not the place for that.

tests/github-glue.test.ts
webstech: Adding more tests.  Separating the `cc` related tests, which do not need actual github updates, to use jest mock functions. They run much faster.

webstech: Should be covered now.  Hopefully I have not missed more test cases.

The tests were redone to use mocks for the github calls since that is not the purpose of these tests.

lib/github-glue.ts
```diff
@@ -100,21 +101,52 @@ export class GitHubGlue {
...
+        if (bodyEnd.match(/:/)) try {
+            bodyEnd.split(/\r?\n/).reverse().forEach(line => {
+                const match = line.match(/^([a-z][-a-z0-9]+):\s*(.*)$/i);
+
+                if (!match) {       // stop if not a footer
+                    throw new Error("No Footer");
+                }
+
+                footerSeparator = ""; // body already has footers
+                if (match[1].toLowerCase() === "cc") try {
```
dscho, line 128: <suggestion>
                 if (!found && match[1].toLowerCase() === "cc") try {
</suggestion>

dscho (dismissed):
Looks pretty good! I like how sweet the unit test looks now (although I have to admit that I only skimmed it, trusting that it is correct).

lib/github-glue.ts
webstech: Pushed it up for one last round.

dscho (approved):

dscho:
Thank you!

webstech:
Thank you for your patience.  Should have been less painful for you.

dscho:
Oh, but it wasn't painful at all! Thank you for taking such a lot of GitGitGadget maintenance off of my shoulders :heart:

PR #363, opened 2020-11-19
tests: only connect via IMAP when so configured

dscho:
This should prevent the CI build from failing (see e.g. [here](https://dev.azure.com/gitgitgadget/git/_build/results?buildId=65324&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=5caf77c8-9b10-50ef-b5c7-ca89c63e1c86)).


webstech (approved):
Thanks.  I had opted for configuration over code but maybe I should not have.

PR #392, opened 2020-12-22
Test isolation

webstech:
This patch series will change the GitHubGlue tests to use unique branch names and pull request titles.  This will get rid of the collisions that are happening when running workflows on multiple platforms at the same time.  Clean up for failed tests is also included.

- tests: isolate GitHubGlue test artifacts
Branch names and pull requests will now include the `platform` to
isolate tests running concurrently.  This could be GitHub, Azure
pipelines or locally on windows/wsl.  Initial cleanup of pull
requests and branches will only be done for local tests.  Other
locations will use timestamped branch names and pull requests.

- tools: new script to clean up GitHubGlue test artifacts
delete-test-branch.ts process the command line and calls
delete-branches.
delete-branches.ts retrieves a list of branches from GitHub
and deletes the ones older than the specified time.  The deletion
of the refs will result in any open pull requests on the branch
being closed.  This routine is also intended to be used from a
GitHub workflow.

- ci: add GitHubGlue test cleanup to workflow
Add a step to the workflow to delete old branches and pull
requests left from failed GitHubGlue test runs.

- tools: add cleanbranch command to package.json
Allow easy running of cleanbranch to delete test branches.
Sample command:
npm run cleanbranch -- -o mygithubuser -r gitout -h 4

**NOTE**
The GitHub GraphQL api is used here.  It was originally used so that a search for PRs could be done:
```
{
  search(
    query: "ggg repo:webstech/gitout in:title is:pr is:open created:<=2020-12-18T20:00:00.249Z", type: ISSUE, last: 100) {
    edges {
      node {
        ... on PullRequest {
          id
          url
          title
          createdAt
          updatedAt
        }      }    }  }}
```
The api provides an easy way to deletes the refs representing branches, which closes the PR (removing the need for the search).

### DIFF
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -114,3 +114,12 @@ jobs:
         GIT_CONFIG_PARAMETERS: ${{ env.GGG_SMTP }}${{ env.GGG_REPO }}
       run: npm run test
       working-directory: gitgitgadget
+
+    - name: Clean up two day old branches and PRs
+      uses: actions/github-script@v3
+      id: clean-up
+      with:
+        script: |
+            const exports = require(`${process.env.GITHUB_WORKSPACE}/gitgitgadget/build/lib/delete-ci-test-branches.js`)
+            await exports.deleteBranches(github, process.env.GITHUB_REPOSITORY_OWNER, process.env.GGG_REPOSITORY);
+      if: env.GGG_REPOSITORY
--- a/tests/github-glue.test.ts
+++ b/tests/github-glue.test.ts
@@ -75,42 +75,57 @@ test("pull requests", async () => {
         await github.authenticate(owner);
         const content = Buffer.from("test data").toString("base64");
 
-        const branch = "test-branch";
-        const branchRef = `refs/heads/${branch}`;
-        const title = "Test pulls integration";
+        const branchBase = `ggg-test-branch-${process.platform}`;
+        const titleBase = `ggg Test pulls integration-${process.platform}`;
 
         const oldPrs = await github.getOpenPRs(owner);
-        let pullRequestURL = "";
+        let suffix = "";
 
         // Clean up in case a previous test failed
+        // NOTE: Runs on GitHub and Azure pipelines use a timestamped
+        // branch/PR request that gets cleaned up separately.
 
-        oldPrs.map(pr => {
-            if (pr.title === title) { // need to clean up?
-                pullRequestURL = pr.pullRequestURL;
+        if (!process.env.GITHUB_WORKFLOW &&
+            !process.env.hasOwnProperty("system.definitionId")) {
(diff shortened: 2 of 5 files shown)


.github/workflows/pr-test.yml
```diff
@@ -114,3 +114,12 @@ jobs:
         GIT_CONFIG_PARAMETERS: ${{ env.GGG_SMTP }}${{ env.GGG_REPO }}
       run: npm run test
       working-directory: gitgitgadget
+
+    - name: Clean up two day old branches and PRs
+      uses: actions/github-script@v3
+      id: clean-up
+      with:
+        script: |
+            const exports = require(`${process.env.GITHUB_WORKSPACE}/gitgitgadget/build/lib/delete-branches.js`)
```
dscho, line 123: Maybe we can find a name that conveys that the code is only ever expected to be run during the integration tests?

tests/github-glue.test.ts
```diff
@@ -75,43 +75,58 @@ test("pull requests", async () => {
         await github.authenticate(owner);
         const content = Buffer.from("test data").toString("base64");
 
-        const branch = "test-branch";
-        const branchRef = `refs/heads/${branch}`;
-        const title = "Test pulls integration";
+        const branchBase = `ggg-test-branch-${process.platform}`;
+        const titleBase = `ggg Test pulls integration-${process.platform}`;
```
dscho, line 79: Maybe we could also include the `id` of the run? That would have two benefits:

- two concurrent runs can't clash anymore
- it makes it easier to see *which* run left branches behind that then needed to be cleaned up later.

webstech: Branches/PRs on Github and Azure will have names/titles like:
```
ggg-test-branch-win32_2020-12-18T23_56_40_422Z
```
While it does not ensure there won't be a conflict the risk is pretty low.  Branches/PRs should only be left behind if the run failed.  Will the raw log from failing runs provide the precision to determine the branch/PR?

Guess I am asking if this can be deferred until we see there is a problem?  There would be different vars to use based on GitHub workflow vs Azure pipeline (or the value could be set in the yaml).

.github/workflows/pr-test.yml
webstech: `delete-ci-test-branches`?  Sorry, my imagination is not running very wild.  The command line driver is not well named either.  It started as an inline script in the yaml but being able to run it from the command line or in the workflow made a lot more sense (especially for testing).

Wide open to naming suggestions.

dscho: That name makes sense to me. Thanks!

tests/github-glue.test.ts
dscho: Okay, let's take it as-is, then.

dscho (approved):

PR #393, opened 2020-12-23
dependabot: increase pull request limit to 10

webstech:
There are frequently more than five updates each week so an increase should help to get @types and base packages updated in one set of pull requests.


dscho (approved):
Thank you!

PR #403, opened 2021-01-03
build(deps): bump @octokit/rest 18.0.9 to 18.0.12

webstech:
- `build(deps): bump @octokit/rest 18.0.9 to 18.0.12`
This change updates the types/schema.  GitHubGlue will now throw errors if the response is missing some of the nullable objects.

- `CI(GitHub): add tests to validate GitHub response`
A new set of tests replaces the octokit endpoint with a wrapper to allow any response to be returned, allowing GitHubGlue checks to be fully and easily tested.
There are two sets of tests.  The first is using the octokit schemas that requires extensive boilerplate data.  The second is schemas limited to the data GitHubGlue is interested in.

- `CI(GitHub): remove tests with octokit schema`
The tests with the octokit schema were added for demonstration and committed for possible future reference.
They are fragile in that any schema changes by GitHub could break the tests, when the changes are of no interest to GitHubGlue.

The package-lock.json includes some extra changes from running `npm update` to ensure all the octokit related packages were current.


webstech:
Two additional points.

1. The octokit schema was not done just for the sake of doing work.  The original idea was to replace the specific endpoints (eg `client.issues.getComment`) and that required using the octokit schema.  Using the wrapper was much simpler and less brittle but I did not think the possible use of the full schema should be lost.
2. The wrapper tests are pretty much unit tests but demonstrate what could be done for other tests to move the mock of GitHubGlue down a layer, letting all the GitGitGadget code get exercised.

dscho:
Happy new year, @webstech ! Wow, this looks like a ton of work. I hope to get to review it tomorrow.

webstech:
> Happy new year, @webstech ! Wow, this looks like a ton of work. I hope to get to review it tomorrow.

Happy new year to you too!  Take note there is a code bump in the third commit that goes away.  If you would prefer I do this differently (describe in the wiki?), let me know.

dscho (approved):
Looks great! Thank you once again for all your hard work.

PR #410, opened 2021-01-07
Stop CC-ing Junio by default

avar:
I don't know if this patch makes sense, and I don't use GitGitGadget
myself, just for use In-Reply-To
https://lore.kernel.org/git/nycvar.QRO.7.76.6.2101071517260.2213@tvgsbejvaqbjf.bet/

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,10 @@ update the description to include information about changes performed relative
 to the latest patch series iteration, and then another `/submit` will ask
 GitGitGadget to send a new iteration of the patch series.
 
+If the patch is meant for consideration and inclusion by the
+maintainer of Git make sure to put `cc: gitster@pobox.com` in the
+description before doing `/submit`.
+
 All relevant information, such as the current iteration of the patch series,
 the Message-ID of the sent mails, etc is stored in the Git notes in
 `refs/notes/gitgitgadget`.
--- a/lib/project-options.ts
+++ b/lib/project-options.ts
@@ -42,7 +42,6 @@ export class ProjectOptions {
         if (await commitExists("e83c5163316f89bfbde", workDir)) {
             // Git
             to = "--to=git@vger.kernel.org";
-            cc.push("Junio C Hamano <gitster@pobox.com>");
             upstreamBranch = "upstream/pu";
             if (await git(["rev-list", branchName + ".." + upstreamBranch],
                 { workDir })) {


avar:
I found out this was done in b2221f4 (Stop Cc:ing the Git maintainer, 2020-01-08) and I'm a victim of Git's defaults + the s/master/main/ change in this repo :(

dscho:
While I have your attention, can you think of a non-cumbersome way to Cc: Junio when it is appropriate?

webstech:
> While I have your attention, can you think of a non-cumbersome way to Cc: Junio when it is appropriate?

Is adding a cc: to the PR description too difficult?  The `/cc` command would allow it if the filtering on `gitster` was moved up a layer to the `MailArchiveGitHelper`.

dscho:
> > While I have your attention, can you think of a non-cumbersome way to Cc: Junio when it is appropriate?
> 
> Is adding a cc: to the PR description too difficult? The `/cc` command would allow it if the filtering on `gitster` was moved up a layer to the `MailArchiveGitHelper`.

The problem is not the mechanics of it. The problem is the under-announced cognitive tax we put on contributors.

I invented GitGitGadget to allow contributors to focus on the code, on their changes, on things that actually matter to them. That does not include the nitty-gritty of the review process of the Git mailing list, or how to set up `git send-email`.

There seems to be the expectation on the Git mailing list that "letting in new people" would just lower the bar to get tons of low-quality contributions, but when I talk to my colleagues, I see that this high bar keeps quite capable people out, preventing the Git project from attracting new talent who do not want to learn e.g. that they have to watch out for twice-a-week mails on the Git mailing list to learn about the status of their contribution.

In other words, while I see that many regulars on the Git mailing list are quite happy with the fact that many contributors are "kept out" by these rules and expectations, I am quite sad about that because I maintain that this mostly keeps out high-quality contributions from developers who may only want to fix that one itch they have (and are totally disinterested in learning how to navigate the Git mailing list's unspoken policies to do that).

So no, I don't think that the problem is solved by saying "just Cc: the maintainer when this is ready". When is it ready? Who is the maintainer? What level of reviewing is enough? How is a one-time contributor supposed to know? Why do we even expect them to know?

webstech:
I misunderstood the context of `when it is appropriate?`.  Thank you for clarifying the issue.  I have been on the confused side of this.

The process can be confusing because you start on GitHub, submit changes, have some email exchanges, go back to GitHub and possibly update the patch, submit changes and then what?  

GitGitGadget could help in a couple of ways.  These are my initial thoughts on this.

1. If a PR appears to be quiet for a certain period (following a submit of course but what other criteria could be considered?), post a comment about whether the PR is ready for moving forward.  The comment would suggest what to do next.
2. Provide a new command to send an email to the maintainer requesting the patch series move forward.  Any extra text in the command would be added beyond the template value. The requestor would not need to know the maintainer info.  The email could be a fake reply to the last posted email on the PR, if found.

Someone familiar with the process would use the command when they felt it was appropriate.  Newbies would get the comment reminders about doing so.

> When is it ready? What level of reviewing is enough? How is a one-time contributor supposed to know? Why do we even expect them to know?

I dropped one question from the list.  The other questions are tough, basically requiring a judgement call.  If the patch is sent to the maintainer to move forward, has responsibility been transferred?  Isn't that what happens now?  

Code for commenting on quiet PRs could learn other tricks (eg the author is not responding).

dscho:
> GitGitGadget could help in a couple of ways. These are my initial thoughts on this.
> 
> 1. If a PR appears to be quiet for a certain period (following a submit of course but what other criteria could be considered?), post a comment about whether the PR is ready for moving forward.  The comment would suggest what to do next.
> 
> 2. Provide a new command to send an email to the maintainer requesting the patch series move forward.  Any extra text in the command would be added beyond the template value. The requestor would not need to know the maintainer info.  The email could be a fake reply to the last posted email on the PR, if found.

I fear that really is the best we can do for now.

webstech:
> I fear that really is the best we can do for now.

Other projects have relate-able issues - I think definitely-typed re-pings maintainers for code reviews after a certain time.

For the first bit (scanning for quiet PRs), any preference of a GitHub workflow vs Azure pipeline?  Should we be preferring workflows now? (do they run on Azure anyway?)

dscho:
>  Should we be preferring workflows now?

I think so.

> (do they run on Azure anyway?)

Yes, they even run in similarly set-up agent pools, but it is my understanding that workflows get all the developer attention these days.

PR #411, opened 2021-01-08
Remove obsolete function and its tests

ugultopu:


ugultopu:
@dscho There are no more references to `getGitHubUserName` in the codebase. Hence, this PR removes that function and its tests. Please feel free to close this PR without merging if you feel like leaving that function.

tests/github-glue.test.ts
```diff
@@ -63,9 +63,6 @@ test("identify user", async () => {
         const github = new GitHubProxy(`../${repo}`, repo);
         await github.authenticate(owner);
 
-        const name = await github.getGitHubUserName(owner);
-        expect(name).toMatch(userName);
-
         const ghUser = await github.getGitHubUserInfo(owner);
         expect(ghUser.login).toMatch(owner);
         expect(ghUser.name).toMatch(userName);
```
webstech, line 68: @dscho Should this test allow `ghUser.name` to be empty?  ie, should GGG contributors be allowed to have no name or should this be a reminder that a name is needed?  Not related to this change and no reason to hold it up but it comes to mind after the previous change.

Only testers with a test repo will actually run this test so it may not show up anyway. @ugultopu Did you set up a test repo when testing?  The basic instructions are at the start of `github-glue.test.ts`.  Also interested in feedback if the instructions are not clear.

ugultopu: No, I didn't set up a test repo, just realized that `getGitHubUserName` wasn't referenced (anymore?) in the TypeScript files and hence, removed it. So `getGitHubUserName` is still being used somewhere?

webstech: > `getGitHubUserName` wasn't referenced (anymore?)

The last reference may have been removed in PR #110, which would make this something I should have removed.

webstech (approved):
Thanks!

tests/github-glue.test.ts
dscho: > should GGG contributors be allowed to have no name

A `/submit` will fail if no name was provided.

PR #412, opened 2021-01-08
/allow: gracefully handle absence of GitHub user name again

dscho:
We became a bit too strict; for details, see

https://github.com/git/git/pull/943#issuecomment-756815087

/cc @ugultopu

### DIFF
--- a/lib/github-glue.ts
+++ b/lib/github-glue.ts
@@ -464,15 +464,11 @@ export class GitHubGlue {
             username: login,
         });
 
-        if (!response.data.name) {
-            throw new Error(`Unable to get name for ${login} - ${
-                response.data.toString()}`);
-        }
 
         return {
             email: response.data.email,
             login: response.data.login,
-            name: response.data.name,
+            name: response.data.name || "",
             type: response.data.type,
         };
     }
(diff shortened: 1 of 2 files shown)


lib/github-glue.ts
```diff
@@ -464,10 +464,6 @@ export class GitHubGlue {
             username: login,
         });
 
-        if (!response.data.name) {
```
webstech, line 467: Sorry.  Does the return need to be `name: response.data.name || "",`?  The `name` changed from string to string|null.  Returning an empty string vs changing the interface and handling other upstream changes (ie `lib/gitgitgadget.ts:284:44`) seems simpler.

webstech: Hmmm.  The upstream error is on using `userInfo.name`.  That should probably be changed to `pr.author` if null user names are allowed.

dscho: Null user names are rejected because a name is required on whose behalf GitGitGadget sends the mails (or at least the cover letter). However, we do support `/allow`ing before a name is set (but `/submit`ing will still fail, with the error message "Error: Could not determine full name of [name]", see [here](https://github.com/git/git/pull/943#issuecomment-756867961)).

> Does the return need to be `name: response.data.name || "",`?

Good point, I made it so.

webstech: > Hmmm. The upstream error is on using `userInfo.name`. That should probably be changed to `pr.author` if null user names are allowed. Also means the interface could be changed to allow null.

Just noting for future clarity this comment is incorrect since the two values are not equivalent.

webstech (approved):
Thanks for fixing this.

PR #422, opened 2021-01-18
ESLint: include also Typescript files in lib/

dscho:
I missed this when reviewing a0a392750003502071dc02794087dfd7440bdafc.


webstech (approved):
Thanks.  I had missed that options in the extending file override the extended file.  Wonder why it was not obviously missing.

Issue #414, opened 2021-01-17
Assist users with next steps...


webstech:
Users are not always inclined to study the doc.  Having submitted a change it is not always obvious what to do next if there are some email exchanges and possibly new /submitted changes.  This can cause changes to linger that would otherwise be ready to be integrated.  It would be helpful to notify users an action may be required on their part to move a patch forward.  Automated posting of a comment on 'quiet' PRs with suggestions could be useful.  When a patch is ready to move forward after some exchanges and updates, the Git maintainer needs to be notified.  A command to do this would be helpful (ie /mergeready, /ready2merge, /mergable, ...?).

PR #410 has the initial discussion of this issue.  Something like [actions/stale](https://github.com/actions/stale/blob/main/action.yml) does some of the work.  There may be other workflows providing more features.

webstech:
More mind dumping on this...

Some other projects use labels to help track the status of PRs.  It could be useful with git as well.

- new users would have a `new user` label to identify contributors who may need extra assistance
- a `submitted` label would be added with the `/submit' command
- an `accepted` label would be added when an integrated type email is seen
- a `pending` label and comment would be added if the PR is submitted but not pending or accepted for `n` days.  This would be removed when adding the `accepted` label or other state changes such as a new `/submit`
- possibly only support a single state (which may not match reality) of submitted/accepted/pending - ie remove other labels to keep a simple view on GitHub (although a dashboard combining the status of git/git and gitgitgitgadget/git could be useful)
- the use of labels would make automation easier since they would provide easy state determination

dscho:
I wonder whether we should do https://github.com/gitgitgadget/gitgitgadget/issues/152 instead, and then warn the user if the topic has either not been picked up for over a week, or has "fallen off the train" (i.e. has been in "What's cooking" but no longer is, despite the PR still being open). Particular attention could be paid to tell the user if their topic has been moved to the "Stalled" section.

dscho:
I _think_ this might be a good start for that idea:

```typescript
/*
 * This class is designed to parse the "What's cooking" mails sent to
 * the Git mailing list about twice a week by the Git maintainer.
 */

export class SousChef {
    protected readonly mbox: string;
    protected readonly messageID: string | undefined;
    protected readonly branches: {
        [branchName: string]: {
            merged: string | undefined;
            sectionName: string;
            text: string;
        }
    };

    public constructor(mbox: string) {
        this.mbox = mbox;
        this.branches = {};

[...]

PR #423, opened 2021-01-20
Add "new users" label to GitGitGadget PRs of first-time users

dscho:
This was suggested in https://github.com/gitgitgadget/gitgitgadget/issues/414#issuecomment-762022194 (among other good suggestions).

### DIFF
--- a/tests/ci-helper.test.ts
+++ b/tests/ci-helper.test.ts
@@ -42,6 +42,7 @@ class TestCIHelper extends CIHelper {
     public ghGlue: GitHubGlue;      // not readonly reference
     public addPRCommentCalls: string[][]; // reference mock.calls
     public updatePRCalls: string[][]; // reference mock.calls
+    public addPRLabelsCalls: any[][]; // reference mock.calls
 
     public constructor(workDir?: string, debug = false, gggDir = ".") {
         super(workDir, debug, gggDir);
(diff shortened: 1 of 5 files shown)


derrickstolee (approved):

tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     /* Basic Options */
-    "target": "es6",                          /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017','ES2018' or 'ESNEXT'. */
+    "target": "es2018",                       /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017','ES2018' or 'ESNEXT'. */
```
webstech, line 4: Should lib be updated in synch with target?  Current recommendations for node 12 are [in this repo](https://github.com/tsconfig/bases/tree/master/bases).  fyi, have been running with latest node without issues so we may want to consider updating the node version in package.json.

dscho: > 
> 
> Should lib be updated in synch with target? Current recommendations for node 12 are [in this repo](https://github.com/tsconfig/bases/tree/master/bases). fyi, have been running with latest node without issues so we may want to consider updating the node version in package.json.

Makes sense! Pushed.

tests/ci-helper.test.ts
```diff
@@ -61,6 +62,12 @@ class TestCIHelper extends CIHelper {
         this.ghGlue.updatePR = updatePR;
         this.updatePRCalls = updatePR.mock.calls;
 
+        const addPRLabels = jest.fn( async (_:string, labels: string[]):
```
webstech, line 65: I guess there is no linting rule to require a space here `_:string`.  I will let you decide if you want to update it.

webstech (dismissed):
Thanks for getting this done.  I will let you decide on the space before merging.

tests/ci-helper.test.ts
dscho: Good point. Fixed.

webstech (approved):
Thanks.

Issue #402, opened 2021-01-01
Automatically close bogus PRs


dscho:
As mentioned in https://github.com/git/git/pull/937#issuecomment-752087607, we'd like to close PRs that are clearly bogus, such as PRs by new contributors trying to merge `next` into `maint` or anything similarly undesirable.

Some ideas for heuristics:

- the base branch isn't in the contributor's fork
- the patches aren't authored by the (not yet /allow'ed) contributor
- a single patch adds new `.c` or `.h` or even `.py` file(s), without adjusting anything else
- ...

tussi003:
> Jak již bylo zmíněno v [git / git # 937 (komentář)](https://github.com/git/git/pull/937#issuecomment-752087607) , rádi bychom, aby poblíž PRS, které jsou jasně falešné, jako jsou VAS nových přispěvatelů, kteří se snaží sloučit `next`do `maint`nebo něco podobně nežádoucí.
> 
> Několik nápadů pro heuristiku:
> 
> * základní větev není ve vidlici přispěvatele
> * opravy nejsou autorem (zatím / nepovoleným) přispěvatelem
> * jedna oprava přidává nové `.c`nebo `.h`dokonce `.py`soubory, aniž by upravovala cokoli jiného
> * ...

webstech:
Update: Ready to open PR

Playing with this.  This comment is tracking what is and isn't working.  Running a test which analyzes the git/git open PRs and many of the closed bogus PRs.  Reject means close the PR and warn means add a label to the PR.  This is a WIP.

**Open PRs**
1. Three have changes to a single source and could be rejected

**Closed PRs**
1.  Reject files with a space in the name
2. Reject files with extensions longer than 8 characters
3. Reject updates to certain filenames (or adding similarly named files) (eg README, SECURITy.md)
4.  Warn for files with no type
5. Warn for files with small additions and no deletions (see open prs above)
6. Warn for files with zero additions and multiple deletions (possibly deleted file)

Current testing results:
Checking 47 Open PRs
--------------------------------
==> PR 924 warned: Label `validate` added for this reason:

File `builtin/init-db.c` expecting more than one change.

--------------------------------
==> PR 907 warned: Label `validate` added for this reason:

File `sha1dc/sha1.c` expecting more than one change.

--------------------------------
==> PR 859 warned: Label `validate` added for this reason:

File `userdiff.c` expecting more than one change.

--------------------------------
==> PR 857 warned: Label `validate` added for this reason:

[...]

PR #441, opened 2021-02-07
Autoclose/flag questionable PRs from new contributors

webstech:
This relates to issue #402.  New PRs by 'unallowed' users will have extra checks.  If the contents appear to be questionable, a label will be added to the PR to request a check be done.  Anyone (authorized?) reviewing the PR can remove the check.  Some PRs will be closed based on history of bad PR requests.

Commits in the PR;

1. Make GitHubGlue::closePR comment generic
The caller will now control the narrative on why a PR is closed.

2. tests: remove redundant new user check
The test did not need to have a new user to be a valid test.

3. Add enhanced linting for new contributors
For first time contributors, extra checks will be done to validate the PR.  A label will be added to flag the PR for further review in many instances.  The label can then be removed by any reviewer.  The intent is to NOT close PRs that may be valid.  Some PRs will be closed based on the content.

Changing some of the tests to flagged PRs instead of closed may be the preferred initial implementation.

### DIFF
--- a/lib/ci-helper.ts
+++ b/lib/ci-helper.ts
@@ -1,5 +1,8 @@
 import * as fs from "fs";
+import * as path from "path";
 import * as util from "util";
+import type { PullRequestChangedFile, Repository
+    } from "@octokit/graphql-schema";
 import addressparser = require("nodemailer/lib/addressparser");
 import { ILintError, LintCommit } from "./commit-lint";
 import { commitExists, git, emptyTreeName } from "./git";
@@ -14,6 +17,12 @@ import { IPatchSeriesMetadata } from "./patch-series-metadata";
 
 const readFile = util.promisify(fs.readFile);
 type CommentFunction = (comment: string) => Promise<void>;
+
+type labelInfo = {
+    label: string;
+    message: string;
+}
+type setLabelInfo = (label: string, message: string) => labelInfo;
 
 /*
  * This class offers functions to support the operations we want to perform from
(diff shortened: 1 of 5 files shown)


lib/ci-helper.ts
```diff
@@ -763,6 +772,211 @@ export class CIHelper {
...
+                pr.baseOwner,
+                queryRef
+            );
+
+            if (!refInfo || !refInfo.repository.ref) {
+                throw new Error(`Base branch ${prInfo.baseRefName
+                    } not found in ${pr.author}/${pr.baseRepo}.`);
+            }
+        } catch (e) {
+            if ((e as labelInfo).label === undefined ) {
```
dscho, line 898: Does that mean that the absence of a label should tell us that the PR should be closed? Maybe an explicit `.closePR: boolean` would be better?

lib/ci-helper.ts
```diff
@@ -763,6 +772,211 @@ export class CIHelper {
...
+
+            const refInfo = await this.github.graphql<queryResponse>(
+                pr.baseOwner,
+                queryRef
+            );
+
+            if (!refInfo || !refInfo.repository.ref) {
+                throw new Error(`Base branch ${prInfo.baseRefName
+                    } not found in ${pr.author}/${pr.baseRepo}.`);
+            }
```
dscho, line 896: I guess this check should come before any linting, as a deleted fork is pretty fatal: that's not a contribution, that's a "throw over the wall".

lib/ci-helper.ts
```diff
@@ -763,6 +772,211 @@ export class CIHelper {
...
+                }
+            }`;
+
+            const refInfo = await this.github.graphql<queryResponse>(
+                pr.baseOwner,
+                queryRef
+            );
+
+            if (!refInfo || !refInfo.repository.ref) {
+                throw new Error(`Base branch ${prInfo.baseRefName
```
dscho, line 894: Is that the `baseRefName`? There should be a `refName`, too, right?

lib/ci-helper.ts
```diff
@@ -763,6 +772,211 @@ export class CIHelper {
...
+    }
+
+    /**
+     * Perform some validation tests on a file in a pull request.
+     *
+     * @param changedFile file that is part of a commit in a pull request
+     * @param files number of files that are changed in the pull request
+     */
+    public checkNewUserFile(changedFile: PullRequestChangedFile,
+        files: number, setLabel: setLabelInfo):
```
dscho, line 922: Let's call it `fileCount` rather than `files`.

lib/ci-helper.ts
```diff
@@ -763,6 +772,211 @@ export class CIHelper {
...
+        const filenameParts = path.parse(changedFile.path);
+
+        // one file with only deletions?
+        if (files === 1 &&
+            !changedFile.additions && changedFile.deletions) {
+            throw setLabel("validate", `File \`${changedFile.path
+                }\` expecting more than one change.`)
+        }
+
+        if (filenameParts.name.match(/.+\./)) {
```
dscho, line 933: Wouldn't that catch `git.c`? It contains a dot that is not the first character.

lib/ci-helper.ts
```diff
@@ -763,6 +772,211 @@ export class CIHelper {
...
+            throw new Error(`File \`${changedFile.path
+                }\` not allowed with multi-part name`);
+        }
+
+        if (filenameParts.name.match(/ /)) {
```
dscho, line 938: I think we'll need to be more precise here:

```shell-session
$ git ls-files *\ *
t/t4135/add-with backslash.diff
t/t4135/add-with quote.diff
t/t4135/add-with spaces.diff
t/t4135/add-with tab.diff
t/t4135/diff-with backslash.diff
t/t4135/diff-with quote.diff
t/t4135/diff-with spaces.diff
t/t4135/diff-with tab.diff
t/t4135/git-with backslash.diff
t/t4135/git-with quote.diff
t/t4135/git-with spaces.diff
t/t4135/git-with tab.diff
```

So maybe something like `/^(?!t\/t[0-9]{4}\/).* /`, i.e. with a negative look-ahead that would force any path starting with `t/tNNNN/` to be excluded from the check?

lib/ci-helper.ts
```diff
@@ -763,6 +772,211 @@ export class CIHelper {
...
+            throw new Error(`File \`${changedFile.path
+                }\` not allowed with space in name`);
+        }
+
+        const badNames = ["SECURITY.md", "README"];
```
dscho, line 943: Can we tighten this `SECURITY.md` test? As I mentioned elsewhere, I would _love_ for us to have a `SECURITY.md`, it's just a lot of work to put in the relevant information. Looking at an example from https://github.com/git/git/pulls?q=is%3Apr+security.md+is%3Aclosed, I see that this would be a good needle:

> Use this section to tell people about

If the PR contains a `SECURITY.md` file containing these words, it is _so_ clearly not a quality contribution that we can close it straight away.

lib/ci-helper.ts
```diff
@@ -763,6 +772,211 @@ export class CIHelper {
...
+                        }\` not updated.`);
+                }
+            }
+
+            if (files === 1 &&
+                filenameParts.ext.match(/^\.([ch]|py|yml)$/)) {
+                // what if they rewrote some logic?
+                if (!changedFile.additions || !changedFile.deletions) {
+                    throw setLabel("validate", `File \`${changedFile.path
+                        }\` expecting more than one change.`)
```
dscho, line 969: Maybe it would be sufficient to check for the addition of one empty line at the end of the file. IIRC that was the case in quite a couple bogus contributions.

dscho (commented):
Thank you for working on this!

webstech:
> Thank you for working on this!

Thank you for the fast and detailed review.

lib/ci-helper.ts
```diff
@@ -1,5 +1,8 @@
 import * as fs from "fs";
+import * as path from "path";
 import * as util from "util";
+import type { PullRequestChangedFile, Repository
+    } from "@octokit/graphql-schema";
```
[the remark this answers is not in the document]
webstech: Do we need to hide the graphql api?  It is not the same as the GitHub rest api where there is  more ORM type management.  Graphql requests are quite data specific so anything coming back from a `GitHubGlue` wrapper would look very much like what is returned now.  There are some tools to generate a type from the graphql query (similar to the hand-coded types in the `ci-helper.test.ts` test cases).  They use the query and the schema to create type files.  Would you want to integrate one of those tools into gitgitgadget?

lib/ci-helper.ts
```diff
@@ -763,6 +772,211 @@ export class CIHelper {
...
+            const commits = prInfo.commits.nodes;
+
+            if (commits) {
+                commits.map(node => {
+                    const login = node?.commit.committer?.user?.login ??
+                        node?.commit.author?.user?.login;
+
+                    // login may not be present if email unknown to github
+                    if (login && pr.author !== login) {
+                        throw new Error("Request author is not commit author.");
```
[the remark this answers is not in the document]
webstech: I decided (right or wrong) to flag for review or close at the first failed check.  If it is flagged/labeled, someone will be reviewing and can make a determination so not all possible problems need to be listed.  It is usually pretty obvious to a human if it is a valid PR or not.

lib/ci-helper.ts
webstech: The code is throwing an `Error` for PRs to be closed and a `labelInfo` for PRs to be flagged.  The check for `label` is really a type check (is there another way to do it?).  If you want, a single type can be used that has `label?`, `closePR` and `message`.

lib/ci-helper.ts
webstech: It is last to avoid a second trip to the server.  It was not happening in many of the tests against existing closed bogus PRs because they failed for other reasons.  If they don't have the branch there is a good chance they will fail the author check.

lib/ci-helper.ts
webstech: This is just the name (ie git) part of the base.

lib/ci-helper.ts
webstech: Okay.

lib/ci-helper.ts
webstech: Was hoping to avoid checking file content.  Some of the bogus PRs had files with names containing `SECURITY.md`.  Since there is not a `SECURITY.md` at the moment, it seems unlikely that a first time contributor would be providing a complete one and this check only adds a label to the PR.

lib/ci-helper.ts
webstech: Depends upon whether we want to start reading files or running diffs.

lib/ci-helper.ts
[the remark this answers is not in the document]
webstech: Another possibility is to move the new user check functions into a separate file/class.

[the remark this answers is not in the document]
dscho: I'd rather keep this simple. To be honest, I would even have preferred to implement our own commit parser and read the commit history from the local repository (into which the commits would have to be fetched)...

But since you already have the GraphQL-based implementation, maybe just implement the minimal interface that we need? Then we can return an array of that interface type from a method in `GitHubGlue`. At least that's my current (admittedly hand-waving) thinking.

lib/ci-helper.ts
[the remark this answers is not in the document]
dscho: > It is usually pretty obvious to a human if it is a valid PR or not.

Right, but the goal of this PR is to make feedback more immediate, and less personal than a human interaction would be.

lib/ci-helper.ts
dscho: I think that would be better. The current version looks a little bit as if it tries too hard to imitate Java's `instanceof` operator (or C++'s RTTI), which Javascript does not have.

lib/ci-helper.ts
dscho: Since this affects new users only, I think it'd not be too bad to do this check first. After all, if the user already deleted their fork, the PR can be closed without any further effort spent: In this case, the contributor is clearly not interested in collaborating, so why should the Git project be.

lib/ci-helper.ts
dscho: Then I'm puzzled what this is supposed to do (and why the `+` is needed: `/.\./` should be idempotent).

lib/ci-helper.ts
dscho: >  Adding check for `dir` starting with `t/`.

More like: Adding a check that the path _does not_ start with `t/t\d\d\d\d/` ;-) (because in the support directories of the tests, the space is legitimate in file names).

lib/ci-helper.ts
dscho: As I said a couple of times, I think it would make sense for Git to have a `SECURITY.md` because it is shown prominently in GitHub's [`Security` tab](https://github.com/git/git/security/policy) (if there _is_ a `SECURITY.md`). But it will take quite a bit more effort to craft a good one, the contributions we got so far were really inadequate: they essentially left the template alone that was supposed to be adapted to whatever project the contribution was supposed to target.

> Was hoping to avoid checking file content.

Ah. Maybe a good indication that we do need the commits locally, and that we do need a real parser so that we can perform all the checks that we are interested in?

lib/ci-helper.ts
dscho: I'd really be in favor of reading the output of `git log -p <base-rev>..<tip-rev>`. That way, we can easily determine whether an empty line was added at the end of a file.

webstech:
Going to close this.  It is not the preferred implementation.  Researching a git based solution.

PR #459, opened 2021-02-21
Clarify "prefix" for subjects

gitster:
It is not wrong per-se to say that our commit titles begin with "<area>:" prefix, but "prefix like tests and commit" is too weak to explain to newbies what these "prefix" words are trying to convey. It's not like a commit that includes tests get "test:" prefix while others get "commit:" prefix, but there is no way for new people to know that.

Instead, hint that the prefix is to state which subsystem the change is about to guide them to form a better commit title.


dscho (approved):
Nice!

PR #458, opened 2021-02-16
"Read" the "What's cooking" mails and map the comments to the PRs

dscho:
One of the most confusing aspects for new contributors seems to be the fact that they do not get any status updates, but that they are expected to get those from the regularly-sent "What's cooking" mails.

Let's help, by parsing those, and echoing the feedback in the corresponding open PRs GitGitGadget knows about.

This addresses https://github.com/gitgitgadget/gitgitgadget/issues/152.

PR #487, opened 2021-03-08
Do handle the "What's cooking" mails correctly

dscho:
In reacting to the review of https://github.com/gitgitgadget/gitgitgadget/pull/458, I managed to break the central part of it. D'oh.

### DIFF
--- a/lib/mail-archive-helper.ts
+++ b/lib/mail-archive-helper.ts
@@ -118,7 +118,7 @@ export class MailArchiveGitHelper {
                 throw new Error(`Could not parse Message-ID of ${mbox}`);
             }
             const whatsCookingBaseURL = "https://lore.kernel.org/git/";
-            for (const branchName of Object.keys(sousChef.branches)) {
+            for (const branchName of sousChef.branches.keys()) {
                 const pullRequestURL = branchNameMap.get(branchName);
                 if (pullRequestURL) {
                     const branchBaseURL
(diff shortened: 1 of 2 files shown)


derrickstolee (approved):

lib/mail-archive-helper.ts
```diff
@@ -117,8 +117,9 @@ export class MailArchiveGitHelper {
             if (!sousChef.messageID) {
                 throw new Error(`Could not parse Message-ID of ${mbox}`);
             }
+            console.log(`Handling "${sousChef.subject}"`);
```
webstech, line 120: Just noting that in the unlikely event `subject` is not found the log will say:

```
  console.log
    Handling "undefined"
```

lib/mail-archive-helper.ts
```diff
@@ -117,8 +117,9 @@ export class MailArchiveGitHelper {
...
             const whatsCookingBaseURL = "https://lore.kernel.org/git/";
-            for (const branchName of Object.keys(sousChef.branches)) {
+            for (const branchName of sousChef.branches.keys()) {
```
webstech, line 122: Sorry I missed this (I had updated in my local test branch).  May need to look at creating tests for `MailArchiveGitHelper`.

dscho: That would be nice. I am not quite comfortable yet adding these tests...

dscho:
So... https://lore.kernel.org/git/xmqqft1542wu.fsf@gitster.c.googlers.com/ has been sent on Monday, and it looks as if this PR did its job (see e.g. https://github.com/gitgitgadget/git/pull/889#issuecomment-793188741)!

Issue #488, opened 2021-03-09
Relax lowerCaseAfterPrefix commit message check to permit ALLCAPS after prefix ("foo: BAR")


ahunt:
I was recently trying to send a patch with the following title for review:

>  init-db: UNLEAK template_dir when converting to absolute path

However GitGitGadget blocked it with the following message:

> Prefixed commit message must be in lower case: init-db: UNLEAK template_dir when converting to absolute path
( https://github.com/gitgitgadget/git/pull/899#issuecomment-792915386 )

This appears to be because of the `lowerCaseAfterPrefix()` check:
https://github.com/gitgitgadget/gitgitgadget/blob/17c499650b56a6ca4aad5094a5b6611f4c5d9fa2/lib/commit-lint.ts#L93

I was wondering whether it's worth making the check less strict:

- AFAIUI "foo: Bar..." is what we want to prevent - `lowerCaseAfterPrefix()`  correctly blocks this example.
- But something like "foo: BAR..." where BAR is a function/macro/other code identifier seems valid?

There are existing commits with the "foo: BAR" format*, so it seems like it would be OK to let these through? (That said I've also found a lot of commits violating the intended rule - so clearly this check is useful.) I'm happy to prepare a PR to change this, but I wanted to file an isssue to discuss it first. My initial proposal is to update the Regex to add `[a-z]` at the end, i.e to match `/^\S+?:\s*?([A-Z][a-z])/`.

\* verified using `git log --format="%s %as" | egrep '^\S+?:\s*([A-Z]{2})'`.

dscho:
Maybe `[a-z ]`, to catch things like `init: A new option`?

ahunt:
Yeah that seems better. I'll prepare a PR.

(I realised that change would block things like `init: I am doing X`, however I can't imagine any valid commit messages following that format, AND git unsurprisingly has no existing commits with such a format.)

ahunt:
(But of course, `init: I am doing X` is already blocked by the existing regex, so that's unaffected.)

dscho:
Fixed via https://github.com/gitgitgadget/gitgitgadget/pull/489! Thanks, @ahunt!

PR #489, opened 2021-03-14
Relax lowerCaseAfterPrefix to permit ALLCAPS after prefix

ahunt:
We do want to block things like:
init: A bad example
init: Another bad example

However this should be acceptable:
init: UNLEAK foo

Therefore we tweak the regex to permit ALLCAPS after the prefix.

See also discussion in #488 .


dscho (approved):
Very nice!

PR #497, opened 2021-03-15
ESLint: quote globs to avoid expanding to invalid patterns

ahunt:
When I try to run 'npm test' I see the following error:

  > gitgitgadget@1.0.0 lint /home/ahunt/git/gitgitgadget
  > eslint -c .eslintrc.js --ext .ts,.js {lib,script,tests}/**/*.{ts,tsx,js}

  Oops! Something went wrong! :(

  ESLint: 7.21.0

  No files matching the pattern "lib/**/*.tsx" were found.
  Please check for typing mistakes in the pattern.

This is because my shell appears to be performing brace expansion before
passing the globs to eslint. Quoting the globs avoids this, and lets me
run 'npm test' locally.

An alternative is to use --no-error-on-unmatched-pattern - however
that's noisier than just quoting the globs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
   "scripts": {
     "build": "tsc",
     "cleanbranch": "node ./build/script/delete-test-branches.js",
-    "lint": "eslint -c .eslintrc.js --ext .ts,.js {lib,script,tests}/**/*.{ts,tsx,js}",
+    "lint": "eslint -c .eslintrc.js --ext .ts,.js '{lib,script,tests}/**/*.{ts,tsx,js}'",
     "start": "node server.js",
     "test": "npm run lint && jest --env=node",
     "test:watch": "jest --watch --notify --notifyMode=change --coverage",


ahunt:
What I don't understand is why no one else who is contributing to gitgitgadget has ever hit this. (It seems to be a common enough issue elsewhere though.)

I was able to reproduce both on Linux and on Mac, and in both cases I'm using bash. Perhaps everyone else is using a different shell though?

webstech:
> What I don't understand is why no one else who is contributing to gitgitgadget has ever hit this. (It seems to be a common enough issue elsewhere though.)
> 
> I was able to reproduce both on Linux and on Mac, and in both cases I'm using bash. Perhaps everyone else is using a different shell though?

The github actions run using bash and are not having this issue.  Do you have `script-shell` set in your npm configuration? hmmm, running on Ubuntu WSL, the globbing is not getting an error or building a full list:

```
echo {lib,script,tests}/**/*.{ts,tsx,js}
lib/**/*.ts lib/**/*.tsx lib/**/*.js script/**/*.ts script/**/*.tsx script/**/*.js tests/**/*.ts tests/**/*.tsx tests/**/*.js
```

Might be something environmental?  Not that I disagree with the change but should we just remove `.tsx` from the list?

sunshineco:
> Might be something environmental? Not that I disagree with the change but should we just remove `.tsx` from the list?

This error is easily explained if @ahunt doesn't have any `.tsx` files in his tree, and also explains why it works correctly when he quotes the pattern, letting `eslint` expand it rather than the shell.

I'm not familiar with this code. Are the `.tsx` files generated? If so, perhaps he's missing the generation step, thus explaining why he's seeing this error.

webstech:
> This error is easily explained if @ahunt doesn't have any `.tsx` files in his tree, and also explains why it works correctly when he quotes the pattern, letting `eslint` expand it rather than the shell.
> 
> I'm not familiar with this code. Are the `.tsx` files generated? If so, perhaps he's missing the generation step, thus explaining why he's seeing this error.

There are no `.tsx` files (and none generated).  Removing `.tsx` from the glob pattern will resolve the problem but not explain it.  My _very simple_ test suggests globbing will create an array of patterns but not any actual names for this particular input.  The man page indicates there should be no errors for unmatched patterns so I am not clear why @ahunt is getting an error.

webstech:
> An alternative is to use --no-error-on-unmatched-pattern - however
> that's noisier than just quoting the globs.

I am confused.  If the problem occurs before eslint runs, how will this help?

webstech:
> I am confused. If the problem occurs before eslint runs, how will this help?

Sorry.  This sounded like a shell globbing issue but a search suggests it might be an eslint issue. See [SO](https://stackoverflow.com/questions/54543063/how-can-i-suppress-the-no-files-matching-the-pattern-message-in-eslint).

Sorry for the noise but the error message was not very helpful (not your fault - not what would be expected from eslint).  Can you run with `echo` instead of `eslint` for the `lint` command to see what is being passed by the shell?

sunshineco:
Sorry for confusion. Without reading the error message closely enough, it sounded like a shell globbing issue, but indeed it appears to be reported by `eslint`. If there are indeed no `.tsx` files anywhere, then removing `*.tsx` from the pattern seems like a good idea in addition to quoting the pattern to ensure it gets passed literally to `eslint` without the shell doing any sort of brace or wildcard expansion on it.

It's still curious, though, why this works in CI but not for @ahunt.

ahunt:
Further investigation as suggested:

1. Using echo instead of eslint prints all permutations:
`lib/**/*.ts lib/**/*.tsx lib/**/*.js script/**/*.ts script/**/*.tsx script/**/*.js tests/**/*.ts tests/**/*.tsx tests/**/*.js`
2. Next I tried using sh as npm's shell using `npm config set script-shell /bin/sh`
Same result as in 1.
3. Then I tried dash: `npm config set script-shell /bin/dash`
`{lib,script,tests}/**/*.{ts,tsx,js}`
4. csh: `npm config set script-shell /bin/csh`
`echo: No match.`
5. ksh: `npm config set script-shell /bin/ksh`
Same result as in 1.
6. zsh: `npm config set script-shell /bin/zsh`
`zsh:1: no matches found: lib/**/*.tsx`

I don't know how to figure out what is being used in CI though - it looks like Github actions used to use bash back in 2019:
https://github.com/actions/runner/blob/main/docs/adrs/0277-run-action-shell-options.md

But then I realised maybe it's my bash version (I'm guessing npm was using bash by default): I was using bash 4.4.23(1)-release. I've upgraded to 5.0.18(1)-release and don't see any difference in behaviour.

dscho (approved):
This change seems not to break anything, and it fixes things for @ahunt. Let's merge it.

dscho:
@ahunt it might have something to do with Ubuntu using `dash` as default shell, I think.

Though I have to admit that I also develop using Ubuntu locally (inside WSL), and haven't encountered the problem.

webstech:
> 1. Using echo instead of eslint prints all permutations:
>    `lib/**/*.ts lib/**/*.tsx lib/**/*.js script/**/*.ts script/**/*.tsx script/**/*.js tests/**/*.ts tests/**/*.tsx tests/**/*.js`
> 2. Next I tried using sh as npm's shell using `npm config set script-shell /bin/sh`
>    Same result as in 1.
> 3. Then I tried dash: `npm config set script-shell /bin/dash`
>    `{lib,script,tests}/**/*.{ts,tsx,js}`

The `dash` shell does not support `braceexpand`, as shown by your results in 3.  Based on 2, it appears your `/bin/sh` is pointing at bash.  It looks like the default shell for macs is now `zsh`.  `man sh` should tell you what is being used.

Always nice to know what was causing the problem.  Something to keep in mind for the future.

ahunt:
Thanks for merging! And yes those observations explain it - I took another look and noticed:
- npm defaults to using the system default shell for running "scripts", including eslint: https://docs.npmjs.com/cli/v7/commands/npm-run-script#script-shell . I think the relevant code lives [here](https://github.com/npm/cli/blob/e6b6f2c47350d472e2cdefff652e52bf99f50d97/node_modules/%40npmcli/run-script/lib/make-spawn-args.js#L29). (And even if npm is invoked from bash, it will still use /bin/sh to invoke eslint.)
- As already discussed upthread, on Ubuntu, /bin/sh == dash, and dash doesn't perform the brace expansion -> Ubuntu or WSL users won't see this issue. 
- On my system (openSuse), /bin/sh is a symlink to bash -> I hit this issue.
- On my mac, /bin/sh is actually the bash binary (this machine is still on Catalina - it seems like the switch to zsh happens with Big Sur)?

dscho:
Aargh! This breaks all the "PR" tests.

@webstech you added `pr-test.yml`, but it is not actually testing PRs. Instead, it is testing on `push`, and only if the branch is _neither_ `master` nor `maint`.

The idea was to prevent redundant workflows from running: when a user pushes to their fork, the workflow is triggered there, and when opening a Pull Request, it would be triggered in the target repository, too.

However, this idea only holds water if the contributor did not disable Actions in their fork.

And @ahunt disabled Actions, so there was no test failure because there was no test in the first place! https://github.com/ahunt/gitgitgadget/actions

ahunt:
> And @ahunt disabled Actions, so there was no test failure because there was no test in the first place! https://github.com/ahunt/gitgitgadget/actions

Sorry about that 🤦 . I don't remember disabling actions... perhaps they require being explicitly enabled first (I've done that now)? I had also been assuming that my PR would be automatically checked (but failed to actually verify that that happened).

Thanks for fixing the breakage. Now that I have actions enabled I can hopefully produce _and_ verify a better fix for the brace expansion isssues (unless you're feeling too burned by my first attempt, in which case I can just override the eslint config in local development).

dscho:
@ahunt none of this is your fault. It was my responsibility as a reviewer, not yours as a contributor, to verify that the tests run on Windows, still.

webstech:
> @ahunt none of this is your fault. It was my responsibility as a reviewer, not yours as a contributor, to verify that the tests run on Windows, still.

I did run on Windows without problems.  I had seen the ESLint doc mentioned double quotes and wanted to verify there was no problem.  Sorry I failed at that.  Not sure if npm is using `cmd` or `powershell` in the `actions` environment.  Locally I am using `cmd`.

dscho:
AFAIR `npm` uses `CMD` on Windows, for historical reasons.

PR #507, opened 2021-03-22
Fix the ESLint configuration

dscho:
After https://github.com/gitgitgadget/gitgitgadget/pull/497 was merged, there are test failures [left and right](https://github.com/gitgitgadget/gitgitgadget/actions):

The fault is mine: I relied on the PR test workflow to figure out whether the Pull Request would break anything.

What I failed to notice was that the PR test workflow was not even run, it did not even run as a CI build on `main` after merging.

Let's fix the breakage, and then err on the side of caution and rather risk running the workflow one time too many than one time too few.


.github/workflows/build-and-test.yml
```diff
@@ -19,13 +19,13 @@
...
 # repo.
 
-name: pr-test
+name: Build & run tests
 
 on:
   push:
     branches-ignore:
-      - master
-      - main
```
webstech, line 28: This will now run every time someone (like me) pushes a rebased main to github.

webstech (approved):
May need to watch for problems when these jobs run in parallel.  There have been some failures sending emails.

webstech:
> There have been some failures sending emails.

I see the email secrets are not set.  I do not remember if I cleared them because of the problems.

dscho:
> 
> 
> > There have been some failures sending emails.
> 
> I see the email secrets are not set. I do not remember if I cleared them because of the problems.

I _think_ they're only set in the Azure Pipeline.

PR #515, opened 2021-03-29
ci: _really_ only handle Dependabot via `pull_request`

dscho:
Currently, Dependabot PRs trigger each build _twice_: one time because of the PR, the other one because of the push. We wanted to prevent the latter specifically, but apparently I got the syntax wrong, see https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-ignoring-branches-and-tags.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -24,7 +24,7 @@ name: Build & run tests
 on:
   push:
     branches-ignore:
-      - dependabot/*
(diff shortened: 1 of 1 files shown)


webstech (approved):
Sorry I missed they were multi-level branch names.

dscho:
> 
> 
> Sorry I missed they were multi-level branch names.

Me, too!

PR #517, opened 2021-03-29
Octokit - convert to new `rest` endpoints

webstech:
GitHub REST API endpoints are moving to the `rest` object in Octokit.  This change supports that move.


webstech:
This is in response to PR #513.

dscho (approved):
Wonderful!

PR #522, opened 2021-04-09
Touch up status updates

dscho:
I noticed that there are a couple status updates that look uglier than necessary. See e.g. https://github.com/gitgitgadget/git/pull/923#issuecomment-816281852.

Let's avoid mentioning an "empty update", and let's include the section name in the status updates.

While at it, mock a method in the `CIHelper` tests that was previously forgotten.

### DIFF
--- a/lib/mail-archive-helper.ts
+++ b/lib/mail-archive-helper.ts
@@ -128,14 +128,26 @@ export class MailArchiveGitHelper {
                     const pre = info?.text
                         .replace(/&/g, "&amp;")
                         .replace(/</g, "&lt;").replace(/>/g, "&gt;");
-                    const comment = `There was a [status update](${
-                        whatsCookingBaseURL}${
-                        sousChef.messageID}) in the "${
-                        info?.sectionName}" section about the branch [\`${
-                        branchName}\`](${
-                        branchBaseURL}${
-                        branchName}) on the Git mailing list:\n\n<pre>\n${
-                        pre}\n</pre>`;
+                    let comment;
+                    if (!pre || pre.trim() === "") {
+                        comment = `The branch [\`${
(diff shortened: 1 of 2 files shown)


tests/ci-helper.test.ts
```diff
@@ -888,6 +904,7 @@ test("handle push/comment too many commits fails", async () => {
     ci.setGHGetPRInfo(prInfo);
     ci.setGHGetPRComment(comment);
     ci.setGHGetGitHubUserInfo(user);
+    ci.setGHGetPRCommits(commits);
```
webstech, line 907: It wasn't really forgotten.  It was more that it did not need to be mocked knowing the PR had no merges to affect the test results.  I agree it should have been mocked in retrospect.

lib/mail-archive-helper.ts
```diff
@@ -128,14 +128,26 @@ export class MailArchiveGitHelper {
...
-                    const comment = `There was a [status update](${
-                        whatsCookingBaseURL}${
-                        sousChef.messageID}) in the "${
-                        info?.sectionName}" section about the branch [\`${
-                        branchName}\`](${
-                        branchBaseURL}${
-                        branchName}) on the Git mailing list:\n\n<pre>\n${
-                        pre}\n</pre>`;
+                    let comment;
+                    if (!pre || pre.trim() === "") {
```
webstech, line 132: Did you want to add a test for this comment being generated?

webstech (approved):
If you don't want to add a test I will merge.  Leaving open for you to decide.

lib/mail-archive-helper.ts
dscho: Not really... :-)

PR #548, opened 2021-05-05
GitHubGlue::getGitHubUserInfo check reply status

webstech:
The octokit api was updated to allow a `202` status on the request.  This does not make sense but it causes an eslint error.  This change  adds a response status check so eslint is able to determine the type of `response.data`.  This change may be removed in the future if the 202 status is removed.


dscho (approved):
Excellent!

PR #608, opened 2021-06-22
Migrate to node v16+ and npm v7+

webstech:
npm v7 introduces a new version of the lockfile.  It is not compatible with older versions of npm.  This change updates the lockfile and managed workflows to the new version.  Azure jobs will need to be updated to use the newer versions of node/npm.

The scheduled and CI jobs will fail with this PR.    They will need to be updated to set the expected  versions.

On a side note, running install against an empty `node_modules` saved a bunch of disk space.


dscho (approved):

dscho:
Vaguely related, so: not _completely_ an aside:

I wonder whether there is any good way to migrate the Azure Pipelines that are not tracked in gitgitgadget/gitgitgadget to GitHub Actions (and tracking them, of course).

The most obvious challenge are the secrets, I guess: we need to ensure that they stay, well, secret.

One new tool in the toolbox we might want to use are environment secrets, where we can restrict in the GitHub UI which workflows are allowed to access them.

Since the introduction of the `workflow_dispatch` event, it is possible to port the `* PR Handler` definitions, too, which GitGitGadget needs to call via its backend (namely, the Azure Function) because we need to be able to pass parameters and we need to trigger on events originating from a _different_ repository (which seems not to be possible with "plain" GitHub Actions so far).

Other than that, I don't think that there is any blocker.

I guess I should write this up in a proper ticket, eh?

What do you think, @webstech ?

dscho:
> Since the introduction of the `workflow_dispatch` event, it is possible to port the `* PR Handler` definitions, too, which GitGitGadget needs to call via its backend (namely, the Azure Function) because we need to be able to pass parameters and we need to trigger on events originating from a _different_ repository (which seems not to be possible with "plain" GitHub Actions so far).

Ah, dangit! _There_ is a thinko. The [`GitGitGadget PR Handler`](https://dev.azure.com/gitgitgadget/git/_build/index?definitionId=3&_a=completed) pipeline is connected to the `gitgitgadget/git` repository, so that it is run automatically on `push` events, and so that its runs are automatically associated with the respective commit/ref on GitHub (for example, you see an "in progress" check in the PR while GitGitGadget is running).

This would _not_ be possible using a GitHub workflow _committed to `gitgitgadget/gitgitgadget`_ at the moment, as the Actions would be associated with _that_ repository and would therefore _never_ show up in the PRs' UI. At least I do not see any way to do that _currently_. Maybe a future improvement in GitHub Actions will allow us to do that. There's hoping.

So let's shelve this idea for now.

dscho:
I wrote this up in https://github.com/gitgitgadget/gitgitgadget/issues/609, anyway, just to have a proper record (and a proper location to discuss this further).

dscho:
I updated the `GitGitGadget PR Handler`, `GitGitGadget PR Handler (git)`, `GitGitGadget PR Handler (dscho)`, `Mirror Git List to GitGitGadget's PRs` and `Update GitGitGadget's PRs` Azure Pipelines to use Node v16.x instead of v12.x.

🤞 that it will all keep working ;-)

dscho:
> 🤞 that it will all keep working ;-)

To force GitGitGadget's hand, I opened a PR (incidentally paying down some of Git for Windows' technical debt) and the build is running: https://dev.azure.com/gitgitgadget/git/_build/results?buildId=95921&view=logs&j=fd490c07-0b22-5182-fac9-6d67fe1e939b

dscho:
The build succeeded, so I'll just go ahead and assume that everything is all right ;-)

PR #616, opened 2021-06-28
ci: auto-approve Dependabot PRs

dscho:
We trust Dependabot and therefore it is not _really_ necessary to ask for a thorough review by a human being. Therefore, let's just auto-approve Dependabot's Pull Requests.

Example run demonstrating that this works: https://github.com/dscho/gitgitgadget/pull/211#pullrequestreview-694436263

### DIFF
--- a/.github/workflows/auto-approve-dependabot-prs.yml
+++ b/.github/workflows/auto-approve-dependabot-prs.yml
@@ -1,0 +1,19 @@
+name: Auto-approve Dependabot Pull Requests
+
+on: [pull_request_target]
+
+jobs:
+  auto-approve:
+    if: github.actor == 'dependabot[bot]' || github.actor == 'dependabot-preview[bot]'
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/github-script@v4
+      with:
+        github-token: ${{ secrets.GITHUB_TOKEN }}
+        script: |
+          await github.pulls.createReview({
+            owner: context.repo.owner,
+            repo: context.repo.repo,
+            pull_number: context.payload.pull_request.number,
+            event: "APPROVE",
+          });


dscho:
@webstech what do you think, will this make merging that regular slur of Dependabot PRs more joyful?

dscho:
For the record, I got the idea from [here](https://github.com/GuillaumeFalourd/setup-windows10-sdk-action/commit/d91bc956fa6f23d328ba1f9f2a78d2eb478f0853), but wanted to avoid depending on [a third-party Action](https://github.com/hmarr/auto-approve-action/blob/v2/src/approve.ts).

derrickstolee (approved):

webstech:
> @webstech what do you think, will this make merging that regular slur of Dependabot PRs more joyful?

Have been thinking about this myself.  The rebasing is the worst part.

1. Should this be restricted to running only when the target repo owner is gitgitgadget?  Don't really need the approvals being done for the Dependabot PRs on my repo.
2. It will be interesting to see if this happens fast enough to avoid the cycle of rebases.
3. What do you think about a scheduled job that actually merges Dependabot PRs that have been approved for an hour or some fixed amount of time (in case there is a rebasing cycle)?

dscho:
> The rebasing is the worst part.

True.

It _does_ annoy me, though, that I cannot easily merge those PRs on my phone (it is quite the hassle to go to the review part and approve).

>     3\. What do you think about a scheduled job that actually merges Dependabot PRs that have been approved for an hour or some fixed amount of time (in case there is a rebasing cycle)?

I've been thinking about this, too. We _could_ actually let this here workflow merge _immediately_ (if mergeable and the checks pass). The only concern I have there is that we should probably make sure that the change is legit, e.g. verifying that the diff really shows `package.json` to be modified only in one line that mentions the same package as the Pull Request, and that `npm ci` ends up with an unmodified `package-lock.json`. Maybe this is too paranoid? But then, maybe not...

.github/workflows/auto-approve-dependabot-prs.yml
```diff
@@ -0,0 +1,19 @@
+name: Auto-approve Dependabot Pull Requests
+
+on: [pull_request_target]
```
webstech, line 3: Is there any way to have this run after the check suite has run successfully on a pull request?  Maybe a combo of `on` and `if`?  Want to avoid a false sense of approval when a changes has failing CI tests.  I am interested in seeing the effects of this change.  Not sure there will be any.  Could end up with PRs that are approved but need merging/rebasing.

webstech:
Let's implement this change and see how it goes when a batch of update PRs get opened.  Manual merging means nothing bad can happen.

.github/workflows/auto-approve-dependabot-prs.yml
dscho: We _could_ add a `github-script` step that simply polls the current PR's checks, and that succeeds only once all those checks have succeeded.

But that would mean a bit more hassle if a PR failed for bogus reasons (such as network problems, something that happens only rarely, but it does happen)...

webstech:
This workflow was disabled when PR #645 failed due to no access.  The controls do not allow PR updates while the repo is locked from updates.  Anyway, today (happy new year), I approved and merged the dependabot PRs using the `GitHub CLI`.  It was simple.

However, looking around it appears there is a secure(?) [auto approve/merge](https://github.com/dependabot/fetch-metadata) workflow suggested by the fine folks at dependabot.  Will look some more but for the moment, the CLI makes the process easier than the web interface.

dscho:
I _think_ we can fix this in a much better way, by using `permissions: contents: write` (see [here](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#permissions)). I just had no chance to get to it yet.

dscho:
Oh, FWW I wrote a node.js script to merge Dependabot PRs, as I wanted to merge only those PRs whose checks succeeded (and did not find an easy way with `gh`). The actual script I use contains way more than that, so I tried to extract the essentials here:

```js
let octokit
let graphqlWithAuth
const getOctokit = () => {
  if (octokit) return octokit

  const ghAuthStatus = ChildProcess.spawnSync('gh', ['auth', 'status', '--show-token'])
  if (ghAuthStatus.error) throw ghAuthStatus.error
  const auth = ghAuthStatus.stderr.toString().match(/Token: (\S+)/)[1]

  octokit = new Octokit({
    auth
  })
  return octokit
}

const dependabotRepos = {
  'git-for-windows/get-azure-pipelines-artifact': {},
  'git-for-windows/setup-git-for-windows-sdk': {},
  'gitgitgadget/gitgitgadget': {
    needApproval: true
  }
}
for (const repo in dependabotRepos) {
  const match = repo.match(/(.*)\/(.*)/)
  if (match) {
    dependabotRepos[repo].owner = match[1]
    dependabotRepos[repo].repo = match[2]
  }
}

const handleDependabot = async () => {
  getOctokit()

  const dependabotPRsToBeMerged = []
  for (const repo in dependabotRepos) {
    const info = dependabotRepos[repo]
    if (typeof info !== 'object') continue
    for (const pull of (await octokit.rest.pulls.list(info)).data) {
      if (pull.user.login === 'dependabot[bot]') dependabotPRsToBeMerged.push(pull)
      else console.log(`Skipping PR from ${pull.user.login}: ${pull.html_url}`)
    }
  }

  if (dependabotPRsToBeMerged.length === 0) {
    console.log('No Dependabot PR to be merged')
    return
  }

  for (const pull of dependabotPRsToBeMerged) {
    // There's no REST API, just this awkward GraphQL thing, to get to the status check roll-up of a commit
    const response = await graphqlWithAuth(`query($owner: String!, $repo: String!, $pull_number: Int!) {
      repository(owner: $owner, name:$repo) {
        pullRequest(number:$pull_number) {
          commits(last: 1) {
            nodes {
              commit {
                statusCheckRollup {
                  state
                }
              }
            }
          }
        }
      }
    }`, {
      owner: pull.base.user.login,
      pull_number: pull.number,
      repo: pull.base.repo.name
    })
    const status = (() => {
      try {
        return response.repository.pullRequest.commits.nodes[0].commit.statusCheckRollup.state
      } catch (e) {
        return '(NONE)'
      }
    })()
    if (status === 'SUCCESS') {
      if (pull.needApproval) {
        console.log(`Approving ${pull.base.repo.full_name}#${pull.number} (${pull.title})`)
        console.log(check(await octokit.rest.pulls.createReview({
          event: 'APPROVE',
          owner: pull.base.user.login,
          pull_number: pull.number,
          repo: pull.base.repo.name
        })).message)
      }

      console.log(`Merging ${pull.base.repo.full_name}#${pull.number} (${pull.title})`)
      try {
        const merge = check(await octokit.rest.pulls.merge({
          owner: pull.base.user.login,
          pull_number: pull.number,
          repo: pull.base.repo.name
        }))
        console.log(merge.message)
      } catch (e) {
        console.log(`Merge failed: ${e}\n${pull.html_url}`)
      }
    } else {
      console.log(`NOT merging ${pull.base.repo.full_name}#${pull.number}: ${status}`)
    }
  }
}

(async () => {
  await handleDependabot()
)().catch(console.log)
```

Maybe we should turn that script into an actual GitHub workflow and replace the previous auto-approver with it?

dscho:
> ```js
>   for (const repo in dependabotRepos) {
>     const info = dependabotRepos[repo]
>     if (typeof info !== 'object') continue
> ```

FWIW I would no longer write it this way, but instead:

```js
for (const info of Object.values(dependabotRepos)) {
```

PR #619, opened 2021-06-30
Do not prevent the built-in FSMonitor patch series from being sent

dscho:
Currently, https://github.com/gitgitgadget/git/pull/923 is blocked because it has more commits than the (hard-coded) limit. Let's lift this limit for this PR specifically, and set an example how the same can be done in the future for other PRs.


jeffhostetler (approved):
Nice!  Thanks!

derrickstolee (approved):

PR #655, opened 2021-07-29
Allow overriding the automatically detected tip commit in git.git

dscho:
This allows us to manually correct incorrectly identified tip commits in git.git, which can lead to incorrect, prematire closing of PRs. This happened in https://github.com/gitgitgadget/git/pull/999, for example.

### DIFF
--- a/script/misc-helper.ts
+++ b/script/misc-helper.ts
@@ -169,6 +169,21 @@ async function getCIHelper(): Promise<CIHelper> {
         const gitGitCommit = commander.args[2];
 
         await ci.setUpstreamCommit(originalCommit, gitGitCommit);
+    } else if (command === "set-tip-commit-in-git.git") {
+        if (commander.args.length !== 3) {
+            process.stderr.write(`${command}: needs 2 parameters:${
+                "\n"}PR URL and tip commit in git.git`);
+            process.exit(1);
+        }
+        const pullRequestURL = commander.args[1];
+        const gitGitCommit = commander.args[2];
+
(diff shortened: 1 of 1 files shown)


derrickstolee (dismissed):
So, we would be able to run this command:

```
/set-tip-command-in-git.git <oid>
```

and that would create the correct tracking? Is this unrelated to the choice of branch name (e.g. `ds/add-with-sparse-index`)?

dscho:
> So, we would be able to run this command:
> 
> ```
> /set-tip-command-in-git.git <oid>
> ```
> 
> and that would create the correct tracking? Is this unrelated to the choice of branch name (e.g. `ds/add-with-sparse-index`)?

No, this is not hooked up via PR comments yet. I guess we could, but then I'd like to wait with implementing that until we need it more often.

script/misc-helper.ts
```diff
@@ -169,6 +169,21 @@ async function getCIHelper(): Promise<CIHelper> {
...
+            process.exit(1);
+        }
+        const pullRequestURL = commander.args[1];
+        const gitGitCommit = commander.args[2];
+
+        const data = await ci.getPRMetadata(pullRequestURL);
+        if (!data) {
+            throw new Error(`No metadata for ${pullRequestURL}`);
+        }
+	data.tipCommitInGitGit = gitGitCommit;
```
webstech, line 185: Is there a tab/indentation issue with this line?

dscho: Sure enough! Fixed.

webstech (approved):

PR #664, opened 2021-08-03
CIHelper::identifyMergeCommit - update refs

webstech:
The method can be called directly so it needs to update the refs.

Running the command 
```
ts-node script/misc-helper.ts identify-merge-commit maint 4589bca829a2ace58bc98876cccd7dbd2e89f732 -g ..\some_new_git_repo
```

where `some_new_git_repo` does not exist or is an outdated github fork will fail with errors like:

```
Caught error Error: git rev-list --ancestry-path --parents 4589bca829a2ace58bc98876cccd7dbd2e89f732..upstream/maint failed: 128,
fatal: ambiguous argument '4589bca829a2ace58bc98876cccd7dbd2e89f732..upstream/maint': unknown revision or path not in the working tree.
```

The error can happen with an outdated github repo but will always happen when a new repo is specified on the command line (a clone is done).

The fix is the simplest fix.  Alternatives are to make `CIHelper::maybeUpdateMail2CommitMap` public or to provide a wrapper function to do the update before calling identifyMergeCommit.

### DIFF
--- a/lib/ci-helper.ts
+++ b/lib/ci-helper.ts
@@ -457,6 +457,8 @@ export class CIHelper {
     public async identifyMergeCommit(upstreamBranch: string,
                                      integratedCommit: string):
         Promise<string | undefined> {
+        await this.maybeUpdateMail2CommitMap();
+
         const revs =
             await git(["rev-list", "--ancestry-path", "--parents",
                        `${integratedCommit}..upstream/${upstreamBranch}`],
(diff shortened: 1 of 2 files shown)


dscho (commented):
Looks as if the tests fail now... maybe the function needs to be mocked, now that it is called more often?

webstech:
@dscho I apologize for the CI tests failing.  Fell in to the trap of 'this change won't hurt anything'.  Will check to see if `insteadOf` or some other config change will work for the tests.

dscho:
Oh, no worries, I would have done the exact same.

webstech:
> Looks as if the tests fail now... maybe the function needs to be mocked, now that it is called more often?

Fixed now thanks to you providing a flag that made it easy.

dscho (approved):

dscho:
Thank you so much!

PR #676, opened 2021-08-22
Resolve node package tar issue with updated version

webstech:
Update minor release for security update.

Tests ran okay.

Unrelated (since all supported versions were affected) but using package dugite is running a 
couple of releases behind current tar release.


dscho (approved):
Looks good!

PR #696, opened 2021-09-09
Fix type error where `encodeWord` is undefined

webstech:
Fix an error that started showing up with typescript 4.4.2.  The exports are now imported as an object to make `this` properly available.


dscho (approved):
Thank you very much!

PR #734, opened 2021-10-12
Add Types to Commander Options

webstech:
Make the types known before upcoming linter changes start flagging them.

Also, correct `delete-ci-test-branches.ts` to properly set a default timeout if none is specified with `--dry-run` present.


dscho:
As always: thank you!

PR #762, opened 2021-11-10
test: use internal SMTP server

webstech:
Use package [`test-smtp-server`](https://github.com/webstech/test-smtp-server#readme) to run an internal SMTP server.  This will remove the dependency on a third party and may be a little faster.

This is an attempt to resolve the `ConnectionTimeoutError: connection timed out. timeout = 3000 ms` errors.  These errors occur randomly and break tests.  Recently, a popular fake SMTP server had a full day outage.

There is no external configuration for this.  The server will always run as part of the test.

The changes in `ci-helper.test.ts` in the actual tests are undenting the code after removing the check to do the test.  The changes up top are mostly removal of the imap related code.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,7 +13,6 @@
         "commander": "^8.3.0",
         "dugite": "^1.104.0",
         "html-to-text": "^8.1.0",
-        "imap-simple": "^5.1.0",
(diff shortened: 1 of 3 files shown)


dscho (approved):
This is awesome! No internet access required, neither any credentials, to verify that sending mails works.

PR #764, opened 2021-11-11
Update `marked` and `@types/marked` to v4.0.0+

webstech:
Source is updated to handle breaking changes.

marked had a second update to handle a problem that showed up with jest but not regular builds.


dscho (approved):
Thank you so much!

PR #777, opened 2021-11-23
Fix `From:` header parsing

dscho:
GitGitGadget had [a problem](https://github.com/gitgitgadget/git/pull/1076#issuecomment-976087421) parsing the `From:` header in https://github.com/gitgitgadget/git/pull/1076. The cause was an incorrect regular expression, which this PR fixes.


mjcheetham (approved):

webstech:
Good catch.

Trying to figure out why the duplicate `cc:` entries were added to the PR description (with a newline?).   There seems to be some encoding changes in the emails.  If you have already accounted for this please say so to save me some time.

dscho:
> If you have already accounted for this please say so to save me some time.

I haven't even noticed, and I am running out of time today before I can work on GitGitGadget :-(

webstech:
The end result is a failure in `PatchSeries::parsePullRequestBody` with a [submit](
https://lore.kernel.org/git/pull.1076.v7.git.1637298298.gitgitgadget@gmail.com/) not including all the `cc:`.  Want to make sure you are aware of this while I research a solution.  Manual edits of the PR description may introduce empty lines that need to be ignored/trimmed when parsing for the footer.

Both `PatchSeries::parsePullRequestBody`  and `GithubGlue::addPRCc` may need changes.

Have not found a way to easily see the raw edits of changes to the PR description that would make it easier to replicate the problem.

dscho:
@webstech thank you for the update, I have not had a chance to focus on this because there have been issues I needed to address before I could release Git for Windows v2.34.1 (which is still not published).

PR #778, opened 2021-11-26
Capture PR footer from last linefeed pair

webstech:
User edited PR descriptions may have an odd number of linefeeds separating the body from the footer.  The footer is now extracted from the last linefeed pair.

Additionally, the trimmed PR body will be used for updating to ensure there is no single trailing linefeed.

This corrects the problem observed in a recent [PR](https://github.com/gitgitgadget/gitgitgadget/pull/777#issuecomment-978889386).

### DIFF
--- a/lib/github-glue.ts
+++ b/lib/github-glue.ts
@@ -111,13 +111,18 @@ export class GitHubGlue {
         const url = GitGitGadget.parsePullRequestURL(pullRequestURL);
         const pr = await this.getPRInfo(url[0], url[2]);
         const trimBody = pr.body.trimRight();
-        const bodyEnd = trimBody.split(/\r?\n\s*?\r?\n/).pop() as string;
+        let footer = trimBody.match(/^[^]+\r?\n\s*?\r?\n([^]+)$/);
+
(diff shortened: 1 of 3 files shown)


lib/github-glue.ts
```diff
@@ -111,13 +111,18 @@ export class GitHubGlue {
         const url = GitGitGadget.parsePullRequestURL(pullRequestURL);
         const pr = await this.getPRInfo(url[0], url[2]);
         const trimBody = pr.body.trimRight();
-        const bodyEnd = trimBody.split(/\r?\n\s*?\r?\n/).pop() as string;
+        let footer = trimBody.match(/^[^]+\r\n\s*?\r\n([^]+)$/);
```
dscho, line 114: Shouldn't we leave the `?` after `\r` here (twice)?

dscho (approved):
Only one request: I would like to still be accepting of LF-only input, even if we expect only CR/LF input on GitHub.

dscho (approved):

dscho:
Thank you so much!

PR #784, opened 2021-11-29
Disallow bi-directional code in ESLint, and run it in CI, too

dscho:
This should safeguard our code a little bit.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -105,6 +105,10 @@ jobs:
       run: npm ci
       working-directory: gitgitgadget
 
+    - name: Run linter
+      run: npm run lint
+      working-directory: gitgitgadget
+
     - name: Run build
       run: npm run build
       working-directory: gitgitgadget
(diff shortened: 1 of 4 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -109,6 +109,10 @@ jobs:
       run: npm run build
       working-directory: gitgitgadget
 
+    - name: Run linter
```
webstech, line 112: The `test` script in package.json already runs lint.  Do we want a script that only runs jest?  I don't always appreciate waiting for linting to run when the change is minor.  Having the CI workflow really reduces the need to run linting with each test run.  The VS Code ESLint extension gets the linting done in real time.

If  you update the yml, could you move this step before the build step?

dscho: Makes sense!

dscho: Fixed.

webstech (approved):

PR #785, opened 2021-12-03
Make `/preview` safer

dscho:
As reported by @ldennington, [one instance of a `/preview` command](https://github.com/gitgitgadget/git/pull/1050#issuecomment-985770245) updated GitGitGadget's notes ref in such a way that it thought it had sent a new iteration, and then failed to find the corresponding tag in gitgitgadget/git.

My best guess is that the notes ref (which was updated locally by that `/preview` command, but never pushed) actually _did_ get pushed by _another_ run (which shares a persistent clone of gitgitgadget/git, to avoid cloning/updating unnecessarily). Granted, this other run should have force-fetched the notes ref first, overwriting any local changes. But it did not.

To avoid that situation, let's prevent `/preview` runs from even modifying that notes ref.

### DIFF
--- a/lib/patch-series.ts
+++ b/lib/patch-series.ts
@@ -850,6 +850,8 @@ export class PatchSeries {
             await this.sendMBox(mails.join("\n"));
         }
 
+        if (this.options.noUpdate) return this.metadata;
+
         logger.log("Updating the mail metadata");
         let isCoverLetter: boolean = mails.length > 1;
         for (const mail of mails) {
@@ -904,7 +906,7 @@ export class PatchSeries {
             await this.notes.set(key, this.metadata, true);
         }
 
-        if (!this.options.noUpdate && publishTagsAndNotesToRemote) {
+        if (publishTagsAndNotesToRemote) {
             await git(["push", publishTagsAndNotesToRemote, this.notes.notesRef,
                        `refs/tags/${tagName}`],
                       { workDir: this.notes.workDir });


dscho:
The commit that got pushed accidentally was https://github.com/gitgitgadget/git/commit/d3f2d065ee1da5f6be378dafdd451c794036fbb6. The next commit on `refs/notes/gitgitgadget` is https://github.com/gitgitgadget/git/commit/1211182256c13bbd4cf6747e3bb6632a8d49d640, which looks as if it was generated by the `script/misc-helper.ts handle-new-mails` command run by the [Mirror Git List to GitGitGadget's PRs Pipeline](https://dev.azure.com/gitgitgadget/git/_build?definitionId=5&_a=summary).

And indeed, [this](https://dev.azure.com/gitgitgadget/git/_build/results?buildId=119852&view=logs&j=275f1d19-1bd8-5591-b06b-07d489ea915a&t=704ea7c9-8eb0-53b9-6b30-87d5a7befe2c&l=15) is where the offending commit was pushed:

> ```
> To https://github.com/gitgitgadget/git
>   9609dade671..d3f2d065ee1  refs/notes/gitgitgadget -> refs/notes/gitgitgadget
> ```

The problem here is that the notes ref is pushed even if a previous step failed (on purpose, in case that some comment has been posted already, and does not need to be posted again, which information is stored in the notes ref). And the previous step failed. Unfortunately, it failed even before it could force-fetch the notes ref to a clean state.

To remedy that, I introduced [another step](https://dev.azure.com/gitgitgadget/git/_build/results?buildId=119883&view=logs&j=275f1d19-1bd8-5591-b06b-07d489ea915a&t=29b11093-f299-52c9-fc3e-b4e81abf4ac1&l=11) that force-fetches the notes ref. It is not a complete fix: the force-fetch could fail for network reasons. In that case, I would expect the push to fail for the same reasons, too, though.

webstech (approved):
Sorry, I did not realize this unrefreshed notes situation could occur in conjunction with an arbitrary push.  Adding the notes refresh step to the pipeline is a good idea.

dscho:
> Sorry, I did not realize this unrefreshed notes situation could occur in conjunction with an arbitrary push.  Adding the notes refresh step to the pipeline is a good idea.

I hadn't realized, either :grin:

ldennington:
Thanks so much for the fix @dscho!

PR #803, opened 2021-12-14
Add one bit of defensive programming

dscho:
This just came up, where GitGitGadget (for an unknown reason) failed not only to send [`[PATCH v2 4/4]`](https://lore.kernel.org/git/7bacb3760f3d8d55b9a8d370642b093e9c60e9a6.1639149490.git.gitgitgadget@gmail.com/) of https://github.com/gitgitgadget/git/pull/1069 (see the [thread](https://lore.kernel.org/git/pull.1069.v2.git.1639149490.gitgitgadget@gmail.com/#r)), but also failed to record the Message-ID of that patch.

While it is unclear why it failed (the [log](https://dev.azure.com/gitgitgadget/git/_build/results?buildId=121013&view=logs&j=fd490c07-0b22-5182-fac9-6d67fe1e939b&t=28e14b79-1d95-5195-ee03-3a68ac48a418) is not illuminating), GitGitGadget should handle this more gracefully than to stop processing _all_ PRs.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -28,6 +28,31 @@
             "outFiles": [
                 "${workspaceRoot}/build/**/*.js"
             ]
+        },
+        {
+            "name": "Debug update-open-prs",
+            "type": "node",
+            "request": "launch",
+            "runtimeArgs": [
+                "--inspect-brk",
+                // "--nolazy",
+            ],
+            "args": [
+                "${workspaceRoot}/build/script/misc-helper.js",
+                "update-open-prs",
+            ],
+            "stopOnEntry": false,
+            "cwd": "${workspaceRoot}",
+            "preLaunchTask": null,
+            "runtimeExecutable": null,
+            "env": {
+                "NODE_ENV": "development"
+            },
+            "console": "integratedTerminal",
+            "sourceMaps": true,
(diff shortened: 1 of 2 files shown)


webstech:
>  While it is unclear why it failed (the [log](https://dev.azure.com/gitgitgadget/git/_build/results?buildId=121013&view=logs&j=fd490c07-0b22-5182-fac9-6d67fe1e939b&t=28e14b79-1d95-5195-ee03-3a68ac48a418) is not illuminating), GitGitGadget should handle this more gracefully than to stop processing _all_ PRs.

As noted in the referenced PR the problem appears to be an empty commit.  I am approving this change as a preventive measure but have not totally convinced myself the `not yet sent` state may not occur.  Back to the problem at hand, will the lack of any later emails cause a problem if the fix is implemented?  Knowing more about the problem now, I will let you decide about merging this patch.

I have verified an empty commit will generate fewer emails than expected.  I did not test with empty commits between non-empty commits.

May need to add empty commits to the checks.

webstech (approved):
Concerns are in the comments.

dscho:
> May need to add empty commits to the checks.

Hmm. That's probably a better way forward. Thanks for the sanity check!

PR #804, opened 2021-12-15
Reject commits expecting to be rebased out

webstech:
Commit messages starting with `squash!`, `fixup!` or `amend!` are special commits to simplify updating a previous Commit.  A rebase should be run to remove them.

### DIFF
--- a/lib/commit-lint.ts
+++ b/lib/commit-lint.ts
@@ -34,6 +34,7 @@ export class LintCommit {
 
         if (this.commitViable()) {
             this.commitMessageLength();
+            this.bangPrefix();
             this.lowerCaseAfterPrefix();
             this.signedOffBy();
             this.moreThanAHyperlink();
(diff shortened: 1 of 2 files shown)


lib/commit-lint.ts
```diff
@@ -99,6 +100,14 @@ export class LintCommit {
         }
     }
 
+    // Reject commits that appear to require rebasing
+
+    private bangPrefix(): void {
+        if (this.lines[0].match(/^(squash|fixup|amend)!/)) {
+            this.block(`Rebase needed to squash commit: ${this.lines[0]}`);
```
dscho, line 107: Every once in a while I send patch series with a `fixup!` for a buggy patch series. Could we turn this into a warning?

derrickstolee: I've tended to use `git format-patch` and paste my `fixup!` commits into a reply, giving me an opportunity to add extra context and keep the patch next to the discussion of the original topic. If that doesn't get the traction needed to apply the `fixup!`, then sending a forward fix as a non-`fixup!` is appropriate.

Would you consider changing your workflow a bit to allow us to help others avoid silly mistakes by keeping this as an error?

dscho: Sure!

dscho (approved):

PR #816, opened 2021-12-21
ci-helper: move checkCommits() before e-mail checking for /preview

softworkz:
This will apply the same order as for `/submit.

It has the advantage that the preview command will be rejected less often, because checkCommits() can autoDiscover users' e-mail addresses from additional places.

### DIFF
--- a/lib/ci-helper.ts
+++ b/lib/ci-helper.ts
@@ -662,7 +662,8 @@ export class CIHelper {
                     }
                 } catch (reason) {
                     throw new Error(`User ${
-                        accountName} is not a valid GitHub username: ${reason}`);
+                        accountName} is not a valid GitHub username: ${
+                        reason}`);
                 }
 
                 if (await gitGitGadget.allowUser(comment.author, accountName)) {
--- a/lib/ci-helper.ts
+++ b/lib/ci-helper.ts
@@ -662,7 +662,7 @@ export class CIHelper {
                     }
                 } catch (reason) {
                     throw new Error(`User ${
-                        accountName} is not a valid GitHub username: ` + (reason || ''));
+                        accountName} is not a valid GitHub username: ${reason}`);
                 }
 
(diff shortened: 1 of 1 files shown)


lib/ci-helper.ts
```diff
@@ -662,7 +662,7 @@ export class CIHelper {
                     }
                 } catch (reason) {
                     throw new Error(`User ${
-                        accountName} is not a valid GitHub username.`);
+                        accountName} is not a valid GitHub username: ` + (reason || ''));
```
dscho, line 665:
```suggestion
                        accountName} is not a valid GitHub username: ${reason}`);
```

lib/ci-helper.ts
```diff
@@ -662,7 +662,7 @@ export class CIHelper {
                     }
                 } catch (reason) {
                     throw new Error(`User ${
-                        accountName} is not a valid GitHub username: ` + (reason || ''));
+                        accountName} is not a valid GitHub username: ${reason}`);
```
dscho, line 665: Oops. I guess this line is still longer than 80 columns.
```suggestion
                        accountName} is not a valid GitHub username: ${
                        reason}`);
```

softworkz: I have disabled the 80 char lint limit locally, because I can't properly read the code with all that weird wrapping applied...

dscho:
I squashed the commits and fixed the commit message a bit, but I could not update your branch because you unchecked the "let maintainers edit" checkbox. Here is what I would suggest to force-push to your branch: https://github.com/gitgitgadget/gitgitgadget/compare/main...dscho:emailcheck

softworkz:
I know that checkbox but I don't see it here..

Let's not  over-exercise this - you can make those changes yourself and close the PRs, that's totally OK for me. ;-)

dscho (approved):

PR #817, opened 2021-12-22
send-mail: fix parsing of mail headers with line breaks between key and value

softworkz:


softworkz:
This is fixing wrapped mail headers like this:

image

The regex is splitting those to a single "line", but it doesn't remove the line breaks within those lines, which causes failure when parsing.

softworkz:
I pushed an update as there still was an issue regarding message id parsing (had a trailing space)

lib/ci-helper.ts
```diff
@@ -662,7 +662,8 @@ export class CIHelper {
                     }
                 } catch (reason) {
                     throw new Error(`User ${
-                        accountName} is not a valid GitHub username.`);
+                        accountName} is not a valid GitHub username: ${
+                        reason}`);
```
dscho, line 666: Could you rebase? This branch still has commits that I had squashed.

lib/send-mail.ts
```diff
@@ -60,13 +60,14 @@ export function parseMBox(mbox: string, gentle?: boolean):
     let subject: string | undefined;
     let to: string | undefined;
 
-    for (const line of header.split(/\n(?![ \t])/)) {
+    for (let line of header.split(/\n(?![ \t])/)) {
+        line = lines[i].replace("\r\n", "\n").replace("\n", " ");
```
dscho, line 64: This causes plenty of errors in the PR's Checks.

How about doing this instead?

```typescript
    for (const line of header.split(/\n(?![ \t])/)
        .map(e => e.replace("\r\n", " ").replace("\n", " "))) {
```

dscho (changes requested):

lib/ci-helper.ts
softworkz: Done

lib/send-mail.ts
softworkz: Your project - your style 😄 
Done.

softworkz: ..even though the column wrap limit is weird. Are you working with a Unix terminal from the 80ies..?

lib/send-mail.ts
```diff
@@ -60,15 +60,22 @@ export function parseMBox(mbox: string, gentle?: boolean):
     let subject: string | undefined;
     let to: string | undefined;
 
-    for (const line of header.split(/\n(?![ \t])/)) {
+    for (const line of header.split(/\n(?![ \t])/)
+        .map(e => e.replace("\r\n", " ").replace("\n", " "))) {
```
webstech, line 64: Should this be done in `mail-archive-helper.ts`?  The 'other' user of this method does not generate `\r` in emails.  Is the split missing something that the second `replace`  is covering?  ie what use case are you seeing?

lib/send-mail.ts
```diff
@@ -60,15 +60,22 @@ export function parseMBox(mbox: string, gentle?: boolean):
...
         const colon = line.indexOf(": ");
+
+        if (colon < 0 && line.lastIndexOf(":") === line.length - 1) {
+            // Keys without value are valid
+            continue;
+        }
```
webstech, line 70: Can this be combined with the following `if` to only check `colon < 0` once?

lib/send-mail.ts
```diff
@@ -60,15 +60,22 @@ export function parseMBox(mbox: string, gentle?: boolean):
...
         if (colon < 0) {
             throw new Error(`Failed to parse header line '${line}`);
         }
         const key = line.substr(0, colon);
-        const value = replaceAll(line.substr(colon + 2), "\n ", " ");
+        const value = replaceAll(line.substr(colon + 2), "\n ", " ").trim();
```
webstech, line 75: Can the `trim` be removed from `case "from":`  below?

lib/send-mail.ts
```diff
@@ -60,15 +60,22 @@ export function parseMBox(mbox: string, gentle?: boolean):
...
         switch (key.toLowerCase()) {
-            case "cc": cc = (cc || []).concat(value.split(", ")); break;
+            case "cc": cc = (cc || []).concat(value.split(", ")
+                .map(e => e.trim())); break;
```
webstech, line 78: Could the split be done with a regex to include whitespace+ and avoid the map?  There is no trailing whitespace at this point.

Just curious, what mail server/agent is including trailing whitespace?  Have not reviewed the RFCs on what is expected;

lib/send-mail.ts
softworkz: IIRC, from the RFCs regarding SMTP mail, CRLF is the correct EOL mode for e-mails and I suspect that it's probably Git's EOL normalization behavior that is changing EOLs to LF only.
On the other hand, a LF without CR doesn't have a separate meaning, that's why from my POV:

- yes, it should be done here to handle both cases
- it is safe to do so

> Is the split missing something that the second replace is covering? ie what use case are you seeing?

One of the reasons why is dislike regex is that there's often some unexpected "twist" in the way it behaves, and this is one example where the answer is: Yes, the regex split is missing something

The intention of the regex is splitting of mail header lines while treating lines which are starting with a space or a tab as being part of the previous line.

The regex does that, and retains such consecutive as a single line in its output.
BUT: without removing the line breaks in those cases.

Please see the screenshot above where the message id is wrapped.

You can reproduce with an e-mail account from outlook.com

lib/send-mail.ts
softworkz: Why? These are two different cases with different conditions and different actions to take.

Do you mean nested `if` statements where the outer one is checking `colon < 0`?

It's not a hot path in a performance-critical loop, so code clarity should be given precedence - IMO.
But I really don't care. If you want that then I can change it.

lib/send-mail.ts
softworkz: Two different cases:

- This is for handling cases where the (whole) value part is starting on a second line.
- The below is about splitting the value part itself into multiple elements

lib/send-mail.ts
softworkz: Yes, there's no trailing whitespace, but I have encountered cases with leading whitespace after the split.

lib/send-mail.ts
webstech: Probably a little confusion here.  The other user is git generated mbox format emails which will only have LF characters.  These are the result of /submit and /preview commands.  Only `mail-archive-helper.ts` would be seeing SMTP conformant emails.

How does this work if the CRLF is in the messages being processed?  Could message content be processed as headers?

```
const headerEnd = mbox.indexOf("\n\n");
```

Please add test cases for the conditions you are encountering so that future changes will not accidentally ignore them.  The current tests do not handle the edge situations you are seeing.

lib/send-mail.ts
webstech: Yes I meant nested.  That is clearer to me for visually scanning code.

lib/send-mail.ts
webstech: If this is staying, can it be:
```
.map(e => e.replace(/\r?\n\s+", " ").trim())) {
```
Would that remove the need for the later `replaceAll`?  No leading whitespace on any lines.

lib/send-mail.ts
webstech: Thanks for clarifying.  Please see earlier comment.

lib/send-mail.ts
webstech: Split with regex /,\s+/?

softworkz: I'm totally fine when you just close the PR and do the changes to your liking ;-)

It's not worth the time for those little changes; anyway you won't see me using regex as long as it's avoidable, but just go ahead, I don't mind how you do it, neither when it's authored by you.

webstech: > I'm totally fine when you just close the PR and do the changes to your liking ;-)

I don't want to close it.  I would like to have test cases with the problems you have encountered so they do not get regressed.  You have the knowledge in this area.

softworkz: OK, will do.

lib/send-mail.ts
dscho: I would prefer that form, too.

lib/send-mail.ts
dscho: I concur with @webstech: nesting it would make it easier to follow the code flow.

lib/send-mail.ts
dscho: I am looking forward to the updates!

lib/send-mail.ts
dscho: > Are you working with a Unix terminal from the 80ies..?

It is the Git project's requirement, not any hardware's. And personally, I don't believe that I am in the business of questioning successful projects' idiosyncrasies too much, myself 😉 I mean, if I started, the FFMPEG project would lend itself as a splendid target, wouldn't it? 😁

lib/send-mail.ts
dscho: An even better idea would probably be to change the search term from `": "` to `":"`: the key names cannot contain colons, therefore the result would be correct, still, and you would not have to special-case the empty value case.

lib/send-mail.ts
dscho: > Split with regex /,\s+/?

For extra goodness, we could use `/,\s*/`, I think. The only remaining problem would be that something like this would not be handled correctly:

```
Cc: "Schindelin, Johannes" <johannes.schindelin@gmx.de>
```

But that's not a new problem, and therefore not the responsibility of this PR to fix.

lib/send-mail.ts
softworkz: Are you sure that there's no value that could contain a colon?
I think a 100% correct approach would be to check for quotes and ignore colons inside quotes.
Checking for space is a somewhat "lazy" solution but still better than the current.

lib/send-mail.ts
softworkz: How sure are you about the suggested code?
I don't want to spend the time to double check and verify that it behaves 100% correct in all possible cases.
Actually it was an unexpectedly behaving regex that caused the incorrect parsing in the first place.

lib/send-mail.ts
softworkz: I don't like regex...

lib/send-mail.ts
softworkz: Absolutely! I'm in no way standing behind what the ffmpeg project does. I'd have a long list to tell ;-)

lib/send-mail.ts
dscho: > Are you sure that there's no value that could contain a colon?

The first `colon` assignment uses `indexOf()`, not `lastIndexOf()`. So it would at most find a _key_ that contains a colon, but AFAIR that is not allowed in emails.

lib/send-mail.ts
dscho: In this instance, it seems to be the best tool for the job, though.

lib/send-mail.ts
dscho: > How sure are you about the suggested code?

It was left as an exercise for the reader to figure out that the first double-quote should be a forward slash followed by a `g`, I think.

softworkz: LOL. You won't get me doing exercises with regex...

Let me explain. My experiences with regex had always been a pattern like this:

- Study the regex syntax
- Build a regex expression for the required situation
- Test it on a number of strings
- Implement code using the regex
- Publish code
- EACH TIME
  at some point there appeared a situation where the regex didn't do as expected
  => bug
- Needed to fix the bug
- Republish code
- Quickly afterwards: forget the regex syntax details

I had the same experience like 4 or 5 times in the past 20 years. Each time, something unexpected has happened.

On the other hand, I've written countless parsers for protocols or formats, bit-based or string-based. 
I write them once, I know exactly what they are doing, they are working rock-solid without surprises and I never need to touch it again (unless there are changes or extensions to the spec).

It goes without saying that there are regex experts that are capable of creating expressions that work without bad surprises, but I'm not such regex expert, and I prefer to write code in a way that I know it is solid.

Change my code, do it with regex, even close the PR and redo it yourself - all fine for me. It's surely a way to do things but not one of mine.  ;-)

lib/send-mail.ts
softworkz: Right, that's not allowed in header keys. 
It's been a few days, I guess I wanted to separate the cases (key+value and key-without-value) for some reason, but I don't remember why.

softworkz:
Another idea for handling this might be to use the Mailparser node component: http://nodemailer.com/extras/mailparser/

We're already using Nodemailer and Mailparser is just a part of it. Probably it does a much better and reliable job in parsing mail headers than we could achieve with the code we have..?

dscho:
> Another idea for handling this might be to use the Mailparser

Sure. I use this in another project and it works pretty well.

webstech:
> Another idea for handling this might be to use the Mailparser

I have pushed a [branch](https://github.com/webstech/gitgitgadget/tree/mailparser) that implements the `simpleParser` front-end to Mailparser.  It removes the need for our own handling of base-64 and libqp.

@softworkz  Does this meet your needs?  I added the empty header support.  Not sure if there are other things to add.
@dscho What do you think?

softworkz:
Thanks for your effort!

Why does it still require the manual checking for a line ending with a colon? Can't the library handle this?

Also, weren't there two cases when a line ends with a colon: 1. empty value, 2. continued at next line when it starts with a space (iirc)?

dscho:
@webstech I like the approach and the branch looks ready, could you open a Pull Request, please?

webstech:
> @webstech I like the approach and the branch looks ready, could you open a Pull Request, please?

PR #889 has been opened. @softworkz Please take a look.

softworkz:
Hmpf, I had already commented that it looks good to me. Probably I forgot to hit the button ;-)

softworkz:
Superseded by https://github.com/gitgitgadget/gitgitgadget/pull/889

webstech:
> Probably I forgot to hit the button

You commented here but I wanted your approval in the actual PR.  Thanks for that.

softworkz:
> > Probably I forgot to hit the button
> 
> You commented here but I wanted your approval in the actual PR. Thanks for that.

Hell, I've gone through all those hundred browser tabs to see what happened 😡 

BTW, how do you delete a comment without it still being shown?
(e.g. like here, below this comment: https://github.com/ffstaging/FFmpeg/pull/18#issuecomment-1012650340)

PR #819, opened 2021-12-22
Update .gitignore to exclude VS files

softworkz:
Would be great when we could get this in first, so I can switch branches without loosing all my environment ;-)


dscho (approved):

dscho:
Huh. I was not even aware that you could develop GitGitGadget using Visual Studio. I always use Visual Studio Code instead.

PR #820, opened 2021-12-25
patch-series: make stripDuplicateHeaders private

softworkz:
... otherwise ISingletonHeader would need to be exported

### DIFF
--- a/lib/patch-series.ts
+++ b/lib/patch-series.ts
@@ -371,7 +371,7 @@ export class PatchSeries {
         });
     }
 
-    protected static stripDuplicateHeaders(headers: string,
+    private static stripDuplicateHeaders(headers: string,
                                            header: ISingletonHeader): string {
         const needle = "\n" + header.key + ":";
         let offset;


webstech:
I assume this is a linting error you are seeing.  eslint (and previously tslint) do not complain.  Obviously if something extending  `PatchSeries` used `stripDuplicateHeaders()` the export would be needed.  Would it be better to make stripDuplicateHeaders() private?  

I suspect when GitGitGadget was developed, most methods were just specified as protected and others just copied that without considering if they should be private.

softworkz:
Making it private would do it as well. Either way is fine I think.

softworkz:
Changed this PR, making stripDuplicateHeaders private

webstech (approved):
Updated PR description from commit message.

webstech:
Thanks.

dscho:
> I suspect when GitGitGadget was developed, most methods were just specified as protected and others just copied that without considering if they should be private.

Right. I also learned Typescript on the go, and TSLint was still a thing (and it also was improved over the years, then replaced by ESLint).

PR #821, opened 2021-12-25
misc-helper/delete-test-branch: use consistent interface names

softworkz:


dscho (approved):
Right, interface names should start with `I`. Thank you!

PR #822, opened 2021-12-25
package.json: add missing dependencies

softworkz:


webstech:
Are these from some experimenting you were doing?

softworkz:
ts-node is never used for execution and seemingly not for tests (otherwise the checks would have failed for this PR).

gulp and chai are dependencies of many other modules. I'm not sure why they aren't installed automatically. They might not be required for execution, but without having them I get a lot of type resolution errors.
When type resolution doesn't work properly all the way through the code, it adversely affects code navigation and reference search (find usages, find derived, find interface implementations, etc.), When that doesn't work properly you might miss things when making changes and possibly make mistakes and miss certain things - especially when you're not deeply familiar with the code. In other words: it affects quality and I hate to make contributions with an aftermath that needs to be addressed.

That's actually the intention behind most of these PRs: getting it clean and correct (even at a higher level than it's required to run successfully). This will allow me to make future contributions at a higher quality than when I would just switch all those checks off.
Anyway, the individual changes are correct and justified, I think.

webstech:
> gulp and chai are dependencies of many other modules. I'm not sure why they aren't installed automatically. They might not be required for execution, but without having them I get a lot of type resolution errors.

They do not appear to be dependencies of anything in GGG since they are not installed.  Considering `chai` is an alternative to `jest`, with some naming overlaps, is there a configuration problem?  Still getting familiar with the files you provided to see if there is something missing from the .njsproj (jest as tester for example).

What kind of type resolution errors does `gulp` resolve?

No one has raised this issue and I am not seeing issues in Visual Studio (although I have not got it configured to run tests yet).  If there are configuration issues, it would be nice to get them resolved and a reference set of solution files could be posted in the wiki (just a suggested location) for anyone wanting to develop GGG using VS.

softworkz:
After deleting gulp. turbo-gulp and chai from the node_modules folder, I'm seeing this after a while (takes several minutes after VS start and opening a number of files in the lib folder:

image

softworkz:
Also, after deleting the modules, and before restarting VS, I have deleted the build, bin and obj folder.

webstech:
Those 'missing' deps are in the devDependencies for `@bcoe/v8-coverage`.  Analysis tools should not be looking at code in node_modules should they?  Sorry but it may be a config issue for whatever tool is providing that window.  If it is standard VS, please tell me how to get there.  My project still says it is happy.

softworkz:
Made another test: it only happens with VS 2022, not with VS 2019 - interesting..

softworkz:
No differences in Settings >> Text Editor >> Javascript/TypeScript

webstech:
> Made another test: it only happens with VS 2022, not with VS 2019 - interesting..

Okay, I have recreated on VS 2022.  Need to figure out the cause.  It is not something missing from `package.json`.

webstech:
This seems to be a new 'feature' of VS 2022.  Haven't found a specific issue opened for it but did find these instructions to get rid of the errors.
```
Check "only report diagnostics for files opened in the editor" in Tools -> Options -> Text Editor -> JavaScript/TypeScript -> Project -> General
```

Using excludes in `tsconfig.json` does not seem to stop the search for more tsconfig files.  btw, the tsconfig you sent has `node-modules`  instead of `node_modules` in the excludes.

softworkz:
That option already existed in VS 2019...

webstech:
> That option already existed in VS 2019...

Agreed.  The doc indicates there was a bit of change in the ts/js tooling for 2022.  I saw that suggestion on an MS VS page (which I can't find anymore) and in the community pages.  I suggested it because it got rid of the extra errors for me.

softworkz:
> > That option already existed in VS 2019...
> 
> Agreed. The doc indicates there was a bit of change in the ts/js tooling for 2022. I saw that suggestion on an MS VS page (which I can't find anymore) and in the community pages. I suggested it because it got rid of the extra errors for me.

Probably only as long as you don't open the code files?

Meanwhile I tried something else: 

I started from a package.json with all module dependencies being empty. From there, I only added those that are required to compile the code and I ended up with this:

```
  "devDependencies": {
    "@types/html-to-text": "^8.0.1",
    "@types/json-stable-stringify": "^1.0.33",
    "@types/libqp": "^1.1.1",
    "@types/marked": "^4.0.1",
    "@types/nodemailer": "^6.4.4",
    "@types/rfc2047": "^2.0.1",
    "typescript": "^4.5.4"
  },
  "dependencies": {
    "@octokit/auth-app": "^3.6.1",
    "@octokit/rest": "^18.12.0",
    "commander": "^8.3.0",
    "dugite": "^1.104.0",
    "html-to-text": "^8.1.0",
    "libqp": "^1.1.0",
    "marked": "^4.0.8",
    "nodemailer": "^6.7.2",
    "rfc2047": "^3.0.1"
  },
  "engines": {
    "node": ">= 16.0.0"
  }
```

=> No more errors in VS 2022!

Of course it's lacking the modules for running the tests now. But at least it's a clean starting point..

softworkz:
The errors appear after adding

```
    "jest": "^27.4.5",
    "ts-jest": "^27.1.2",
```

to the dev-dependencies

webstech:
> Probably only as long as you don't open the code files?

I get some 'errors' when opening files but not the 40+ errors related to the deps you identified.  The errors relate to ggg files and not deps.  Are you saying there was no change for you with that option?

softworkz:
> > Probably only as long as you don't open the code files?
> 
> I get some 'errors' when opening files but not the 40+ errors related to the deps you identified. The errors relate to ggg files and not deps. Are you saying there was no change for you with that option?

I haven't tried that option yet. The concern that I have regarding that option is that it might also affect code navigation and reference search. I can try that, but it's interesting that it's "jest" which is causing all this.

webstech:
> I haven't tried that option yet. The concern that I have regarding that option is that it might also affect code navigation and reference search. 

Not seeing any issue.

>it's interesting that it's "jest" which is causing all this.

Sorry I do not see the relation to "jest".  It is just a package with lots of deps.  The errors are only on packages which have a tsconfig file.  There are a couple of packages with a tsconfig that do not get errors, just to make sure we are confused.

softworkz:
> Sorry I do not see the relation to "jest".

See above. With this dependency set, there aren't any errors: https://github.com/gitgitgadget/gitgitgadget/pull/822#issuecomment-1001814751

The errors are only appearing after adding 

```
 "jest": "^27.4.5",
 "ts-jest": "^27.1.2",
```

to package.json

(of course you need to delete bin,obj, build, and node_modules when testing this)

softworkz:
We don't need to go crazy about VS issues, though. 
We can just stick to VS 2019 for now, and in one or a few months they might have addressed this in VS 2022 anyway :-)

webstech:
Okay.  @dscho can comment on the use of `ts-node` but the rest of this seems moot.  Need to look at the more minimal deps/devdeps you identified.

softworkz:
> Okay. @dscho can comment on the use of `ts-node`

From my understanding, ts-node is obsolete, because the pre-transpiled output is used for execution rather than jit'ed ts files.

Otherwise, yes - I'll use VS 2019 and avoid further headache..

dscho:
> @dscho can comment on the use of `ts-node`

I use this for running manual tests via `ts-node script/misc-helper.ts [...]`...

softworkz:
Ah alright, understood. Closed it anyway and for the VS2022 problem: somebody else will get sufficiently annoyed by this sooner or later and get it fixed by MS.

PR #823, opened 2021-12-25
various files: use explicit types in variable declarations

softworkz:


lib/ci-helper.ts
```diff
@@ -754,7 +755,7 @@ export class CIHelper {
         if (result) {
             const results = await Promise.all(
                 commits.map((commit: IPRCommit) => {
-                const linter = new LintCommit(commit);
+                const linter: LintCommit = new LintCommit(commit);
```
webstech, line 758: Is this really needed?

lib/send-mail.ts
```diff
@@ -117,7 +117,7 @@ export function parseMBoxMessageIDAndReferences(parsed: IParsedMBox):
         if (header.key === "In-Reply-To" || header.key === "References") {
             let value: string = header.value;
             while (value) {
-                const match = value.match(msgIdRegex);
+                const match: RegExpMatchArray | null = value.match(msgIdRegex);
```
webstech, line 120: Is this really needed?

lib/ci-helper.ts
softworkz: It shouldn't. Of course it should be inferred, don't know why it shows an issue.

lib/send-mail.ts
softworkz: same as above..

lib/ci-helper.ts
webstech: What linter is saying there is an issue?

softworkz: Seems to be a ReSharper bug.

Without type spec:

image

With type spec:

image

lib/send-mail.ts
softworkz: I have updated the PR to remove those changes.

lib/ci-helper.ts
softworkz: have updated the PR to remove those changes.

dscho (approved):
I am not necessarily convinced that those changes are _required_, but they don't hurt either, so I'll merge this PR.

softworkz:
No they aren't logically required. But this solves the problem now while waiting for R# to fix it would take weeks or months :-)
Thank you!

PR #824, opened 2021-12-25
sous-chef/delete-test-branches: add missing semicolons

softworkz:


webstech (approved):
Thanks.  This is not the last of these.

dscho:
> Thanks. This is not the last of these.

I wonder whether there is an ESLint rule we could use to auto-fix these issues?

softworkz:
> Thanks. This is not the last of these.

These were in fact _all_ missing semicolons (R# has such checks).

webstech:
> These were in fact _all_ missing semicolons (R# has such checks).

Turning on the `semi` rule in eslint reveals six more missing `;`.  I will open a PR to add the rule and update the files.

softworkz:
Oh, sorry for misinformation, I'm curious where they are..

webstech:
> Oh, sorry for misinformation, I'm curious where they are..

Sorry it was five (I had one in a file I was editing).  Please see first commit in PR #840.

PR #825, opened 2021-12-25
use explicit and compiler-friendly type conversions

softworkz:
In all of those cases, I'm seeing type conversion errors from static analysis. The "extra-tours" in this PR are clearing up typing and type conversions for the compiler


softworkz:
Merged into other PR https://github.com/gitgitgadget/gitgitgadget/pull/823

PR #826, opened 2021-12-25
github-glue/patch-series: avoid parameterless catch clauses

softworkz:
It's allowed since ECMA2019. 
I'm not sure whether typescript allows it, but VS doesn't like it and fails to parse the whole files when catch parameters are omitted

### DIFF
--- a/lib/github-glue.ts
+++ b/lib/github-glue.ts
@@ -137,11 +137,11 @@ export class GitHubGlue {
                             throw new Error("Found");
                         }
                     });
-                } catch {
+                } catch (_) {
                     // quick exit for cc matched (comment to quiet linter)
                 }
             });
-        } catch {
+        } catch (_) {
             found = false;          // ensure it was not a cc: false positive
             footerSeparator = "\r\n"; // reset
         }
--- a/lib/patch-series.ts
+++ b/lib/patch-series.ts
@@ -252,7 +252,7 @@ export class PatchSeries {
             // \r\n line endings.
             prTemplate = prTemplate.replace(/\r\n/g, "\n");
             prBody = prBody.replace(prTemplate, "");
-        } catch {
+        } catch (_) {
             // Just ignore it
         }
 


webstech:
> VS doesn't like it and fails to parse the whole files when catch parameters are omitted

I am not seeing any issue running pretty current:
VisualStudio.16.Release/16.11.8+32002.261
TypeScript Tools for Microsoft Visual Studio16.0.30526.2002

What versions are you getting the error on?

softworkz:
With both, 16.11.8 and 17.0.4, it looks like this:

image

I should add that I have set up a nodejs (.njsproj) project for this. 
Here's a zip: [gitgitgadget.zip](https://github.com/gitgitgadget/gitgitgadget/files/7776336/gitgitgadget.zip)
(can't publish as branch as the project files are git-ignored.

How are you using VS with the code?

softworkz:
PS: the code in the zip file has all of today's PRs applied.

webstech:
> I should add that I have set up a nodejs (.njsproj) project for this. Here's a zip: [gitgitgadget.zip](https://github.com/gitgitgadget/gitgitgadget/files/7776336/gitgitgadget.zip) (can't publish as branch as the project files are git-ignored.

Thanks for the solution/project files.

> How are you using VS with the code?

I had not been - I have been using vscode.

I am still not seeing the problem you are getting.  Are you using another extension to lint the code?

softworkz:
> I am still not seeing the problem you are getting. Are you using another extension to lint the code?

So, when you 

- open the solution in my zip file in VS 
- and you open the github-glue.ts file
- and you remove the (err) part from the catch clauses

you don't see any errors?

I have nothing special installed. In VS 2019, I installed TypeScript Extensions 4.5.2
Never used NodeJS projects before, so it's unlikely that I messed with some settings.

From a command prompt in the project folder:

```
PS F:\Projects\folder\gitgitgadget> npm -version
6.14.13
PS F:\Projects\folder\gitgitgadget> node --version
v14.17.2
```

softworkz:
Oh...it's ReSharper. The situation where I had disabled it for counter-checking was in a different context probably.

webstech:
> Oh...it's ReSharper. 

Is this PR still needed?

softworkz:
> > Oh...it's ReSharper.
> 
> Is this PR still needed?

It would be nice because I can't change R# and the change doesn't really do any harm.

dscho:
> (can't publish as branch as the project files are git-ignored.

You can `git add -f <files>` to add those files anyway.

lib/github-glue.ts
```diff
@@ -137,11 +137,11 @@ export class GitHubGlue {
                             throw new Error("Found");
                         }
                     });
-                } catch {
+                } catch(e1) {
```
dscho, line 140: We seem to consistently leave a space between `catch` and the parameter list, to distinguish this from a function call.

Also, I seem to recall that the convention for unused parameters is to prefix them with `_`, or to just use an underscore to begin with.

dscho:
Once again, I tried to force-push to your branch to get this PR moving again, but was greeted with this:

```console
$ git push --force-with-lease ffstaging HEAD
remote: Permission to ffstaging/gitgitgadget.git denied to dscho.
fatal: unable to access 'https://github.com/ffstaging/gitgitgadget/': The requested URL returned error: 403
```

Do you see a checkbox like this on your PR, in the sidebar to the upper right?

image

If yes, please make sure that it is checked. That way we can avoid unnecessary back-and-forths.

In this instance, I would like to ask you to `git pull --rebase=i https://github.com/dscho/gitgitgadget initial_patches_07`, review the merge conflict, then `git rebase --skip` and then force-push.

lib/github-glue.ts
softworkz: The underscore as a "discard qualifier" works in C#, but not in TS - I tried as I had the same idea.
But I can prefix with underscore of course.

softworkz:
> Once again, I tried to force-push to your branch to get this PR moving again, but was greeted with this:
> 
> ```
> $ git push --force-with-lease ffstaging HEAD
> remote: Permission to ffstaging/gitgitgadget.git denied to dscho.
> fatal: unable to access 'https://github.com/ffstaging/gitgitgadget/': The requested URL returned error: 403
> ```
> 
> Do you see a checkbox like this on your PR, in the sidebar to the upper right?
> 
> image
> 
> If yes, please make sure that it is checked. That way we can avoid unnecessary back-and-forths.
> 
> In this instance, I would like to ask you to `git pull --rebase=i https://github.com/dscho/gitgitgadget initial_patches_07`, review the merge conflict, then `git rebase --skip` and then force-push.

No, that checkbox doesn't exist in this constellation - I have no idea why. I know that checkbox but it doesn't show in any of my PRs here..

image

softworkz:
I have invited you to the repo with full push access

softworkz:
> > (can't publish as branch as the project files are git-ignored.
> 
> You can `git add -f <files>` to add those files anyway.

If you want to include this, I could also re-setup the vs project and solution in a way that it resides in a sub-folder (it seems to be supported to have a different root folder for the actual node content). 
This way it could be added without cluttering the root folder.

lib/github-glue.ts
dscho: > The underscore as a "discard qualifier" works in C#, but not in TS

It seems to work for me, though.

dscho:
> No, that checkbox doesn't exist in this constellation - I have no idea why.

Here is the official documentation: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

I guess the problem is that you pushed the branch to https://github.com/ffstaging/, which is not your personal place on GitHub, but an org instead.

dscho (approved):

softworkz:
> I guess the problem is that you pushed the branch to [@ffstaging](https://github.com/ffstaging/), which is not your personal place on GitHub, but an org instead.

Ah yes, that makes sense. I created an org, so I can easily transfer ownership to the project eventually. 

Thanks for merging.

lib/github-glue.ts
softworkz: When I used it in two places, I got a warning about a re-used variable, so it seems that it's just a variable named `_`.
The semantics of a "discard qualifier" would be that you can use it multiple times without the compiler assuming it to be a named variable.

The warning could also be a linting mistake, but a search for "typescript discard qualifier _" didn't show any hints that the underscore would be supported as such.

PR #827, opened 2021-12-25
tslint: don't require parameter and argument alignment

softworkz:
The current code doesn't adhere to this and this also contradicts
with max-line-length: aligning parameters or argument often
requires to indent subsequent lines so far that the line
length requirement would need to be violated

### DIFF
--- a/tslint.json
+++ b/tslint.json
@@ -3,8 +3,6 @@
     "rules": {
         "align": [
             true,
-            "parameters",
-            "arguments",
             "members",
             "elements",
             "statements"


dscho (approved):

dscho:
> The current code doesn't adhere to this

That's not a good reason.

> and this also contradicts with max-line-length

But this is.

webstech:
> Maybe it should be removed altogether then?

That is being done in the second commit in PR #840.

Related to other R# issues, what version are you running?  Can you turn off some of the warnings?  Their doc seems to indicate you can.  Really thinking about the `catch` parameter issue here.

softworkz:
> > Maybe it should be removed altogether then?
> 
> That is being done in the second commit in PR #840.

👍 

> Related to other R# issues, what version are you running? Can you turn off some of the warnings? Their doc seems to indicate you can. Really thinking about the `catch` parameter issue here.

No, this is different: it breaks all subsequent parsing of the file, it's not just one of those hints you can configure.
(otherwise I wouldn't have bothered you with it)

PS: latest R# version, long-time user for C# and C/C++; (JS/TS just occasionally)

Issue #818, opened 2021-12-22
Commit Checks


softworkz:
https://github.com/gitgitgadget/gitgitgadget/blob/41979ba3c50892eb3f618e9b0f45e9732757faf9/lib/commit-lint.ts#L35-L41

The requirements are similar to ours but there are differences, so it's moot to discuss individual ones. Just the above seems a bit odd: 

- at first, you might get a reply that says your message would be too short
  - you might end up writing a longer message, but you still get the same reply
  - because it is not checking for message length at all but for additional lines instead
- The above condition can be satisfied, alone by adding signed-off 
- Only after having satisfied the initial condition (msg length), you get the others, e.g. : capitalization
  - make you wonder: "hey, why didn't it say that in the first place?"

I just wonder why it's done that way..

webstech:
The original intent was to check if a commit was viable based on minimal linting.  If it passed that, all the other checks were run, so they would all report issues.   The viable checks assumed if the minimal check was not met, more checking would probably fail.

Can this be adjusted for your needs (and linting requirements) or is a plug-in framework needed (that could help with generic project support as well).

Not sure if this answers your issue.  If a PR is viable, all lints should run.  If they don't, we messed up conversion from asynch to synch methods (my opinion).

softworkz:
My point is that the logic is flawed:

You can't report that the message would be "too short" from a check that doesn't measure the message length but the line count.

commitViable() should be removed and all would be good .

softworkz:
> Can this be adjusted for your needs (and linting requirements) or is a plug-in framework needed (that could help with generic project support as well).

I think a configuration file would be sufficient where you can specify like:

```
check1: true,
check2: false,
check3: true,
check3_maxLen: 80,
```

softworkz:
> My point is that the logic is flawed:

I'm not nit-picking about hypothetical situations. I've gone through this myself when I submitted a (real) patchset as PR for testing:

- I ended up duplicating the commit message texts to satisfy the conditions
- I force-pushed
- then I got new errors about missing signed-off text
- I added the signed-off lines to all commit messages 
- force-pushed again

then I realized that I wouldn't have been required to increase the text lengths at all. the addition of a signed-off line alone is silencing the "message too short" check. And the annoyed user has been me in that case. ;-)

dscho:
> The user should be told right away what he needs to change

[...]

PR #841, opened 2022-01-07
lint: run all possible checks on short commit msg

webstech:
Actually two changes here.

### lint: run all possible checks on short commit msg

- Refactor linting into 3 phases where phase 3 only runs if phase 1 succeeds.
- The commit message will be shown before any linting messages to make it easier to identify.  
- Added check on long Lines in commit description.
- Cleaned up some dead comments.

### linting: add options interface for customization
The initial option allows setting of the max line length in the commit message.  This will help position for when some method of customizing is selected.

### DIFF
--- a/lib/commit-lint.ts
+++ b/lib/commit-lint.ts
@@ -5,6 +5,10 @@ export interface ILintError {
     message: string;
 }
 
+export interface ILintOptions {
+    maxColumns?: number | undefined; // max line length
+}
+
 /*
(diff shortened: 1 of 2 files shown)


webstech:
Re-pushed to address semicolon  problems with function declarations and late rebase.  Lots of LOL for me.

This should address the concerns (or many of them) in issue #818.

dscho (approved):
Great!

PR #842, opened 2022-01-09
Allow some long lines in commit lint

dscho:
When a line simply cannot be wrapped (because there is no white-space where we could wrap the line), we should exempt the line from the too-long-line check.

Requested by @newren in https://github.com/gitgitgadget/git/pull/1113#issuecomment-1008229286.

### DIFF
--- a/lib/commit-lint.ts
+++ b/lib/commit-lint.ts
@@ -110,7 +110,11 @@ export class LintCommit {
         }
 
         for (let i = 1; i < this.lines.length; i++) {
-            if (this.lines[i].length > this.maxColumns) {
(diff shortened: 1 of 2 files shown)


webstech (approved):
Nice cleanup of the tests.  I had got into mundane cut/paste mode.

Thanks for fixing.  I saw the failures but had not got to it yet.  Column 10 is kind of arbitrary but it should be fine until it is not (eg _full-commit: long-url_).

dscho:
> Column 10 is kind of arbitrary but it should be fine until it is not

Right. My idea was to just go with it until it breaks.

newren:
Wow, that was quick.  Thanks!

Issue #829, opened 2021-12-25
Factoring Out all Project Specific Values


softworkz:
At this point, I have just plainly replaced all ggg specific parameters in my fork, which is undoubtedly not a great way for moving forward.

As I already know all the places that require parametrization, I would get this done and contribute that change. I'd just like to avoid doing it more than once, so it would be great when we could agree about a pattern for this up-front.
I don't have any preferences for this, other than it should probably be a single file at the end where you can do the adaption for using ggg in another project.

Maybe you can just make a suggestion?

dscho:
> Maybe you can just make a suggestion?

So far, we did a pretty good job of allowing things to be configured via the `[gitgitgadget]` section of the Git config...

softworkz:
Those few things yes - but I'm talking about the many things that are hardcoded now but need to be configurable.
Look at project-options.ts. I think there's a reason why you didn't make all this configurable via git config..?

dscho:
> there's a reason why you didn't make all this configurable via git config..?

Yes. The reason is that all of GitGitGadget used to be [a shell script](https://github.com/dscho/mail-patch-series), with no configuration whatsoever. We made configurable what we needed to be configurable. And now we need more things to be configurable.

softworkz:
So you want to have it all in the Git config? OK we can do that.

webstech:
I'm not sure the git config is the best place.  A json file may be better.  Future changes requiring git config changes would be a problem for random contributors.

This is project level configuration that users should not have to manage.  The current ggg related git config values tend to be user specific.

softworkz:
From my point of view: the more direct and connected to the code - the better.
Perfect would be something that that connects even at dev-time - i.e. some form of code file (basically in a way like this: https://github.com/gitgitgadget/gitgitgadget/issues/818#issuecomment-1006054970).

The advantage of this would be to have the ability to see how things are connected and which values are configured already during development, just by navigating through the source code (when the file would be included in your IDE).

One problem with this project is that the actual execution is happening somewhere "far away" which makes it impossible to debug and read runtime values. 
When having an indirect mechanism like via a config file, and there's some issue to troubleshoot, you often want to verify that the right values are loaded from configuration. This is somewhat tedious because it forces you to add some code to dump/log those values to verify. Besides that, you need to be careful which values to dump, because the logs are publicly visible...

For those reasons, a configuration that is part of the code would be my preference. 
[...]

PR #849, opened 2022-01-11
maint: expand maximum line length and update node version

webstech:
### editor: increase max line length to 120
Reduce the excessive wrapping of lines.  Suggested in #829.
### maint: update compiler options and node version
Node v12 goes out of support soon.

The workflows are already using node v16 so may as well make any
developers use it.

Make the compiler targets `es2021`.  There are some new features
that may be nice to use.  Added one of the recommeneded options
from https://github.com/tsconfig/bases/.


dscho (approved):
Nice work!

PR #850, opened 2022-01-14
lookup-commit.sh: we entered the next epoch on lore.kernel.org/git

dscho:
Let's follow suite so that we do not fall behind in GitGitGadget.

This PR is needed to fix the failures of the "Update GitGitGadget's commit to mail notes" builds, e.g. [this one](https://dev.azure.com/gitgitgadget/git/_build/results?buildId=124673&view=logs&j=275f1d19-1bd8-5591-b06b-07d489ea915a&t=33e5d3ec-87e7-5f80-0281-074c6962cb44&l=36)


webstech (approved):
Not that it will be needed for awhile but probably should be configurable.

dscho:
> Not that it will be needed for awhile but probably should be configurable.

That, or GitGitGadget should automatically detect lore's epoch changes.

Issue #150, opened 2019-11-19
Replace the many `parsePullRequestURL()` calls with a proper object


dscho:
Rather than parsing the URL over and over and over again, let's add a class that is constructed with that URL and then parses the thing once, and only once.

Suggested by @webstech.

PR #852, opened 2022-01-17
dry part 1: GitHubGlue support for keys or url as parm

webstech:
This is part of a series to make GitGitGadget more generic for use with other repos.  As part of this it also starts to address issue #150.  

The `pullRequestKey` may not be the best name or object but it is also not the first name or iteration.  Suggestions are welcome.  The existing `parsePullRequestURL` should be able to go away with the next iteration.

A next step would be to update the `azure` function to add the repo name to the trigger parms.  This can then be handled in the pipeline to set a third parm of the repo.  A full key would then be available for PR processing.  

### test: ci-helper check last parm for value

The number of parms to GitHubGlue::updatePR could change.  This
removes a dependency on the number of call parameters.

### dependency: remove repo url from notes

Part of a set of changes to make GitGitGadget more supportive of    other projects.

### regex reduction: introduce PR keys object

First phase in removing reparsing of pull request URL.  This positions   `GitHubGlue` to accept a URL string or object for most methods.

### DIFF
--- a/lib/ci-helper.ts
+++ b/lib/ci-helper.ts
@@ -11,6 +11,7 @@ import { MailArchiveGitHelper } from "./mail-archive-helper";
 import { MailCommitMapping } from "./mail-commit-mapping";
 import { IMailMetadata } from "./mail-metadata";
 import { IPatchSeriesMetadata } from "./patch-series-metadata";
+import { getPullRequestKeyFromURL } from "./pullRequestKey";
 
 const readFile = util.promisify(fs.readFile);
 type CommentFunction = (comment: string) => Promise<void>;
@@ -567,6 +568,11 @@ export class CIHelper {
         }
         const command = match[1];
         const argument = match[3];
+        const prKey = {
+            owner: repositoryOwner,
+            repo: "git",
+            pull_number: comment.prNumber
+        };
 
         const pullRequestURL = `https://github.com/${
             repositoryOwner}/git/pull/${comment.prNumber}`;
@@ -588,8 +594,7 @@ export class CIHelper {
             }
 
             const getPRAuthor = async (): Promise<string> => {
-                const pr = await this.github.getPRInfo(repositoryOwner,
-                                                       comment.prNumber);
+                const pr = await this.github.getPRInfo(prKey);
                 return pr.author;
             };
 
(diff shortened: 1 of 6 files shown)


lib/ci-helper.ts
```diff
@@ -11,6 +11,7 @@ import { MailArchiveGitHelper } from "./mail-archive-helper";
 import { MailCommitMapping } from "./mail-commit-mapping";
 import { IMailMetadata } from "./mail-metadata";
 import { IPatchSeriesMetadata } from "./patch-series-metadata";
+import { getPullRequesKeyFromtURL } from "./pullRequestKey";
```
dscho, line 14: Typo? I think the `t` before `URL` belongs before the `Key` instead.

lib/ci-helper.ts
```diff
@@ -834,11 +845,10 @@ export class CIHelper {
         return await mailArchiveGit.processMails(prFilter);
     }
 
-    private async getPRInfo(prNumber: number, pullRequestURL: string):
+    private async getPRInfo(_prNumber: number, pullRequestURL: string):
```
dscho, line 848: Why not remove the `prNumber` parameter altogether?

lib/github-glue.ts
```diff
@@ -227,20 +226,20 @@ export class GitHubGlue {
...
 
         const status = await this.client.rest.pulls.createReplyForReviewComment(
             {
                 body: comment,
                 comment_id: id,
-                owner,
-                pull_number: nr,
-                repo,
+                owner: prKey.owner,
+                pull_number: prKey.number,
```
dscho, line 241: If we called the attribute `prKey.pull_number`, we could use the `...prKey` syntax to DRY up the code even further (not only here, but elsewhere, too). Even if we don't, `pull_number: prKey.number, ...prKey` might work.

dscho (dismissed):
I love it!

I left a few suggestions, hopefully you find them useful.

lib/ci-helper.ts
webstech: Sorry, I usually catch that sort of thing.

lib/ci-helper.ts
webstech: I was trying to minimize the changes but yes, that makes more sense.

lib/github-glue.ts
webstech: Great idea!  Felt awkward using a type as a name anyway.

I did not use `...prKey` when `issue_number` was part of the request.  Concerned about future rejection of unexpected parameters (pull_number).  Should this be a concern?

dscho: > I did not use ...prKey when issue_number was part of the request.

Makes sense!

dscho (approved):
What a delightful PR. Thank you!

PR #858, opened 2022-01-19
dry part 2: expand use of `pullRequestKey` over URL string

webstech:
Warning: many continued lines have been joined using new longer line length.

###  GitHubGlue: reduce updatePR parameter typings

Currently only used internally with `pullRequestKey`.

### CIHelper: more pullRequestKey use and longer lines

More changes to use `pullRequestKey` and rely less on reparsing URL.

A number of multi-line statements were re-wrapped to one line,  making the code a little more obvious.

### GitGitGadget: remove parsePullRequestURL

Convert last user to new function.


dscho (approved):
Awesome clean-up!

PR #868, opened 2022-01-27
dry part 3: more prep for using generic config

webstech:
### cihelper: validate owner/repo before submission

Moving from GitGitGadget::genAndSend.  It is more consistent with other checks being done at that layer.

### patch-series-meta-data: remove unused type

The type will resurface as part of configuration.

### githubglue::constructor - pass repo name as needed

While the default would be okay in these cases, specifying the name makes it clear this will need to specify a custom name to support other projects.

### githubglue::annotateCommit: add baseOwner parm

This additional parm (combined with changes to always specify the repo on the constructor) makes GitHubGlue generic.

### code compaction: remove many remaining line wraps

An initial cleanup of ci-helper.

### DIFF
--- a/lib/github-glue.ts
+++ b/lib/github-glue.ts
@@ -61,7 +61,7 @@ export class GitHubGlue {
     }
 
     public async annotateCommit(originalCommit: string, gitGitCommit: string,
-                                repositoryOwner: string): Promise<number> {
+                                repositoryOwner: string, baseOwner: string): Promise<number> {
         const output =
             await git(["show", "-s", "--format=%h %cI", gitGitCommit],
                       { workDir: this.workDir });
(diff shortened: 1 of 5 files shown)


lib/github-glue.ts
```diff
@@ -70,7 +70,7 @@ export class GitHubGlue {
             throw new Error(`Could not find ${gitGitCommit}: '${output}'`);
         }
         const [, short, completedAt] = match;
-        const url = `https://github.com/git/git/commit/${gitGitCommit}`;
+        const url = `https://github.com/${baseOwner}/${this.repo}/commit/${gitGitCommit}`;
```
dscho, line 73: The fork could have been renamed, though. See e.g. https://github.com/git-for-windows/msys2-runtime/, which was forked from https://github.com/Alexpux/Cygwin.

dscho (approved):
Thank you for continuing to work on this!

lib/github-glue.ts
dscho: I merged the PR anyway because the "base repo has a different name than fork" problem is not one we have to face _now_.

PR #877, opened 2022-02-01
dry part 4a: more code reduction

webstech:
### code compaction: remove line wraps

Cleanup of project-options.

###  ci-helper: remove redundant check for options

`getGitGitGadgetOptions()` will throw an error if there are no     options so no need to check on the return.

### debug: add config to jest debug current file 

Provide ability to set breakpoint in current test file and start debugging using a test run on that file.

Added source directories to search because ts-jest appears to build there.

Renamed workspaceRoot to workspaceFolder (deprecated).

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -13,11 +13,11 @@
                 // "--nolazy",
             ],
             "args": [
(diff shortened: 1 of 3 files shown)


dscho (approved):

dscho:
Thank you!

PR #880, opened 2022-02-03
ci(approve): add the required permissions

dscho:
This fixes the permissions of the Dependabot PR auto-approver, so that we can re-enable it again.

### DIFF
--- a/.github/workflows/auto-approve-dependabot-prs.yml
+++ b/.github/workflows/auto-approve-dependabot-prs.yml
@@ -2,6 +2,9 @@ name: Auto-approve Dependabot Pull Requests
 
 on: [pull_request_target]
 
+permissions:
+  pull-requests: write
+
 jobs:
   auto-approve:
     if: github.actor == 'dependabot[bot]' || github.actor == 'dependabot-preview[bot]'


webstech (approved):
The endless cycle of rebases is more problematic for me.  Using the GitHub cli to approve changes has made that easier.  Probably going to write some code to automate it - check checks, approve if needed, merge, wait a bit and repeat until done.

dscho:
> Probably going to write some code to automate it - check checks, approve if needed, merge, wait a bit and repeat until done.

Why not automate it in a workflow? The [`on.check_suite`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#check_suite) syntax allows us to trigger a workflow when a check suite has finished. I _think_ we could then use an `if:` statement to skip the job unless the check suite failed due to a merge conflict, and otherwise add a comment `@dependabot rebase`.

Of course, we could also try to rebase ourselves, figuring out which package was updated, and then push the result, but if we try to do that via a workflow with `permissions: contents: write`, then the check won't run again because an event triggered by a workflow cannot trigger another workflow, whereas an event triggered by a workflow can trigger Dependabot which in turn can trigger another workflow. (Yes, this is a bit convoluted.)

dscho:
Oh, we could instead trigger [`on: pull_request: types: ['synchronize']`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request and skip the job if `github.event.pull_request.mergeable_state != "dirty"` (which seems to be the indicator of merge conflicts, if there are no merge conflicts, the value will be `"clean"` according to my tests).

PR #889, opened 2022-02-13
sendMail: change parseMBox to use simpleParser

webstech:
simpleParser handles base-64 and quoted-printable as well as specializing in email parsing.

The cc test is expanded and the to test now includes multiple recipients.

The tests using mbox2markdown for base-64 and quoted-printable just use the parser now.

### DIFF
--- a/lib/gitgitgadget.ts
+++ b/lib/gitgitgadget.ts
@@ -168,7 +168,7 @@ export class GitGitGadget {
         const email = userInfo.email;
 
         const send = async (mail: string): Promise<string> => {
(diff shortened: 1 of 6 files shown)


softworkz:
From a quick look, this seems to be fine. I've just too many projects in parallel that I can't switch back and forth too many times and I'd like to combine the adoption of this with the adoption of your (great) work regarding config abstraction.

Anyway - IMO this is the best way to go and even when there would be an issue, it wouldn't question the PR in genera, but only require a fix on top of it.

dscho (approved):

PR #897, opened 2022-02-16
misc-helper has a const config

webstech:
Ensure only one config is set.  Added some basic validation.

This was requested in PR #887.  If more validation is needed, it should be moved to a separate file.  If [genConfig](https://github.com/webstech/ggguser/blob/init/src/genConfig.ts) is added to gitgitgadget, it would use the validator. 

A reason to add `genConfig` is to do this determination in the workflow instead of `ProjectOptions` (i.e. removing more ggg specific code):
```
} else if (await commitExists("cb07fc2a29c86d1bc11", workDir) &&
        await revParse(`${baseCommit}:git-gui.sh`, workDir) !== undefined) {
        // Git GUI
        to = "--to=git@vger.kernel.org";
        cc.push("Pratyush Yadav <me@yadavpratyush.com>");
        upstreamBranch = "git-gui/master";
```

### DIFF
--- a/script/misc-helper.ts
+++ b/script/misc-helper.ts
@@ -46,39 +46,33 @@ if (commander.args.length === 0) {
 }
 
 const commandOptions = commander.opts<ICommanderOptions>();
-let config: IConfig = getConfig();
-
-async function getGitGitWorkDir(): Promise<string> {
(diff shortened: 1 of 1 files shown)


dscho (approved):
Looks good, thank you very much!

Issue #904, opened 2022-02-24
Problem handling Message-IDs


dscho:
I noticed [here](https://dev.azure.com/gitgitgadget/git/_build/results?buildId=129968&view=logs&j=275f1d19-1bd8-5591-b06b-07d489ea915a&t=704ea7c9-8eb0-53b9-6b30-87d5a7befe2c&l=16), [here](https://dev.azure.com/gitgitgadget/git/_build/results?buildId=129979&view=logs&j=275f1d19-1bd8-5591-b06b-07d489ea915a&t=704ea7c9-8eb0-53b9-6b30-87d5a7befe2c&l=15) and [here](https://dev.azure.com/gitgitgadget/git/_build/results?buildId=129982&view=logs&j=275f1d19-1bd8-5591-b06b-07d489ea915a&t=704ea7c9-8eb0-53b9-6b30-87d5a7befe2c&l=15) that there is a new bug where GitGitGadget fails to parse some "Message-ID" that is not actually a single message ID but a space-separated list of message IDs.

I suspect that the switch to using the `mailparser` might have introduced this bug, maybe when looking at `In-reply-to:` headers?

dscho:
> I suspect that the switch to using the `mailparser` might have introduced this bug, maybe when looking at `In-reply-to:` headers?

Actually, looking at [this failure](https://dev.azure.com/gitgitgadget/git/_build/results?buildId=129982&view=logs&j=275f1d19-1bd8-5591-b06b-07d489ea915a&t=704ea7c9-8eb0-53b9-6b30-87d5a7befe2c&l=15) a bit more closely, by inspecting the commit range [21ca1511c592bef2ee91a36692a62200e05dcc55..c4620617e12433ba1b2de3e22a8046a4d0b053a3](https://github.com/gitgitgadget/git-mailing-list-mirror/compare/21ca1511c592bef2ee91a36692a62200e05dcc55..c4620617e12433ba1b2de3e22a8046a4d0b053a3), I see [`<pull.986.git.1624559401.gitgitgadget@gmail.com>` in the `References:` header](https://github.com/gitgitgadget/git-mailing-list-mirror/compare/21ca1511c592bef2ee91a36692a62200e05dcc55..c4620617e12433ba1b2de3e22a8046a4d0b053a3), not the `In-reply-to` header.

dscho:
[...]

PR #905, opened 2022-02-24
Fix parsing of the `References:` header

dscho:
When we replaced our home-grown mbox parser with a version that uses the `mailparser`, we did not adjust the regex when parsing the `References:` header.

As a consequence, our parser got confused when the references stretched over multiple lines (because the regex did not expect Carriage Returns nor Line Feeds).

Let's fix this.

This fixes https://github.com/gitgitgadget/gitgitgadget/issues/904.

### DIFF
--- a/lib/send-mail.ts
+++ b/lib/send-mail.ts
@@ -115,7 +115,7 @@ export function parseMBoxMessageIDAndReferences(parsed: IParsedMBox):
         /^\s*<([^>]+)>(\s*|,)(\([^")]*("[^"]*")?\)\s*|\([^)]*\)$)?(<.*)?$/;
     for (const header of parsed.headers ?? []) {
         if (header.key === "In-Reply-To" || header.key === "References") {
-            let value: string = header.value;
+            let value: string = header.value.replace(/[\r\n]/g, " ");
             while (value) {
                 const match = value.match(msgIdRegex);
                 if (!match) {
--- a/tests/send-mail.test.ts
+++ b/tests/send-mail.test.ts
@@ -1,6 +1,6 @@
 import { expect, test } from "@jest/globals";
 import { MailArchiveGitHelper } from "../lib/mail-archive-helper";
-import { parseMBox } from "../lib/send-mail";
+import { parseMBox, parseMBoxMessageIDAndReferences } from "../lib/send-mail";
 
 const mbox0 =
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -1,3 +1,4 @@
+123
```
webstech, line 1: Was this an accident?

lib/send-mail.ts
```diff
@@ -115,7 +115,7 @@ export function parseMBoxMessageIDAndReferences(parsed: IParsedMBox):
         /^\s*<([^>]+)>(\s*|,)(\([^")]*("[^"]*")?\)\s*|\([^)]*\)$)?(<.*)?$/;
     for (const header of parsed.headers ?? []) {
         if (header.key === "In-Reply-To" || header.key === "References") {
-            let value: string = header.value;
+            let value: string = header.value.replace(/[\r\n]/g, " ");
```
webstech, line 118: Should it be `\r?\n` in case this is called in the future with an mbox that has already removed all `\r`?

webstech (commented):
Sorry about this.  I 'think' I did not remove all the line-ends at line 66 was because something in unicode was getting messed up.

Do you think there is a possibility of `Message-ID` being split over two lines (ie a line-end may need to be removed)?

README.md
dscho: Yes.

lib/send-mail.ts
dscho: I'd rather treat even stray `\r` as whitespace.

README.md
dscho: Fixed.

dscho:
> Do you think there is a possibility of `Message-ID` being split over two lines (ie a line-end may need to be removed)?

There _might_ be such a thing, potentially, but I'd have to research more how this would look. For now, I think we should be good enough with the current version.

webstech (approved):
Thanks.

lib/send-mail.ts
dscho: Sorry for being so curt... it was a little late for me and I wanted to go to bed :smile:

webstech: > I wanted to go to bed

No problem.  Any time you post during my day I pay attention.

Speaking of going to bed...

Issue #869, opened 2022-01-31
Be more helpful when pointing out issues


dscho:
In https://github.com/git/git/pull/1207#issuecomment-1025479875, I felt the need to explain why a `noreply.github.com` email address is not good enough. Which makes me believe that GitGitGadget could be taught to be more helpful in the first place.

We probably want to vet all linter messages, imagining ourselves as new contributors to the project who do not know anything about the customs of the Git project. If any linter message would leave such a new contributor behind even more puzzled than before, it is safe to assume that the message can be easily improved.

PR #906, opened 2022-02-25
Be more helpful when pointing out that a missing public email

dscho:
This is one step in the direction of addressing https://github.com/gitgitgadget/gitgitgadget/issues/869.

### DIFF
--- a/lib/ci-helper.ts
+++ b/lib/ci-helper.ts
@@ -508,7 +508,10 @@ export class CIHelper {
     }
 
     protected warnOnMissingPublicEmail(username: string): string {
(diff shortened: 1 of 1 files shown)


webstech (approved):

webstech:
@dscho The message text needs to be adjusted to avoid embedded new lines.  I have some changes to put in and can include it.  You can see the current format [here](https://github.com/webstech/gggmonitored/pull/1#issuecomment-1053157047).

dscho:
Thank you for cleaning up after me! Just one thing: there's a trailing `}` :smile:

webstech:
> Thank you for cleaning up after me!

Just returning the favour.

PR #978, opened 2022-05-08
MailArchiveGitHelper: format replies as emails in the GGG PR

phil-blain:
GitHub's Linguist, which powers Markdown code block syntax highlighting
[1], understands ```` ```email```` as an indication to format
the following code block as en email [2]. This allows quoted text from
previous emails to be colored differently.

Syntax-highlight the replies from the mailing list in the corresponding
PR comments.

[1] https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-and-highlighting-code-blocks#syntax-highlighting
[2] https://github.com/github/linguist/blob/cddf7476af4c95d1572956ffc5c0cb84f7e431c5/lib/linguist/languages.yml#L1467-L1481

### DIFF
--- a/lib/mail-archive-helper.ts
+++ b/lib/mail-archive-helper.ts
@@ -48,8 +48,10 @@ export class MailArchiveGitHelper {
             return "";
         }
 
-        const wrap = "``````````\n";
-        return `${wrap}${body}${body.endsWith("\n") ? "" : "\n"}${wrap}`;
+        const backTicks = "``````````";
+        const wrapTop = `${backTicks}email\n`;
+        const wrapBottom = `${backTicks}\n`;
+        return `${wrapTop}${body}${body.endsWith("\n") ? "" : "\n"}${wrapBottom}`;
     }
 
     protected readonly branch: string;


dscho:
Hmm. Wouldn't we have to provide the _unparsed_ mbox, then? Does that work with base64-encoded, or quoted-printable mboxes? Do you have an example to show how the result would look like?

dscho:
Let's just go with it and iterate if necessary.

phil-blain:
Yes, I was going to give a similar example.

> Wouldn't we have to provide the unparsed mbox, then? Does that work with base64-encoded, or quoted-printable mboxes?

I'm not sure about that.

dscho (approved):
Looks good! I probably would have used `${backticks}email\n` instead of `${wrapTop}`, but that's just personal style, and not a blocker.

Thank you for improving GitGitGadget!

dscho:
> > Wouldn't we have to provide the unparsed mbox, then? Does that work with base64-encoded, or quoted-printable mboxes?
> 
> I'm not sure about that.

Since current version of the patch uses the parsed mbox, it's not really of any concern.

PR #993, opened 2022-05-25
lint: allow long lines if prefixed with whitespace

phil-blain:
Quoted error messages can often be longer than 80 characters and it
makes sense to allow them to be kept on a single line.

Loosen the "line too long" check to allow long lines if they are
prefixed with whitespace, as it is customary to prefix copied error
messages, or terminal session excerpts, with whitespace.

Add a test and adjust two existing tests to not prefix long lines with a
space.


webstech (approved):
Thanks!

dscho:
I'm late to the party, but would like to add my "very nice!", too ;-)

PR #994, opened 2022-05-30
Updates to support packaging and more config support

webstech:
Three commits (package.json has the latest versions of all dependencies):

- Make workdir required and add owner to github-glue

   Make workdir required for CIHelper (always passed anyway).  Also set some values using the config instead of assuming git.

   Make workdir required for GitHubGlue and add owner for authentication use.  Having owner removes the need to authenticate ahead of any other calls.
   Note: Both workdir and repo are iffy as class members.  eg. annotateCommit could be in CIHelper with just the check create in this class.
- Add mailparser to package.json

   All external packages must be identified in order for gitgitgadget to be used as an npm package.
- Change include 'require' to 'from'

   The ncc tool does not seem to support 'require' on the import statement.  Changing to 'from', which works with current tooling (tsc, ncc).

### DIFF
--- a/lib/ci-helper.ts
+++ b/lib/ci-helper.ts
@@ -1,6 +1,6 @@
 import * as fs from "fs";
 import * as util from "util";
-import addressparser = require("nodemailer/lib/addressparser");
+import addressparser from "nodemailer/lib/addressparser";
 import { ILintError, LintCommit } from "./commit-lint";
 import { commitExists, git, emptyTreeName } from "./git";
 import { GitNotes } from "./git-notes";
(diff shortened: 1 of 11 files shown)


dscho (dismissed):
Thank you for your tireless work on improving GitGitGadget!

dscho (approved):

webstech:
> Thank you for your tireless work on improving GitGitGadget!

Thanks for straightening out my little mess.  I did a rebase and somehow lost the merges when it stopped for the package.json conflict.  Then everything showed up as new commits, which I did not notice before pushing.

dscho:
> Thanks for straightening out my little mess.

No worries. I'm happy if I can contribute using my modest skills (for which I usually make up via tenacity).

PR #1002, opened 2022-06-05
Prefer environment vars over config values

webstech:
When running as part of a workflow, the secrets are set in the environment.  This removes the need to add them to a git config, reducing the risk of exposure.  The '--gitgitgadget-work-dir' is no longer needed (except for the generated app key).

Still thinking about the app key for actions.  The action could generate one to be used with just that run.  Haven't looked too closely but I thought an app key may have more privileges, such as creating checks.

### DIFF
--- a/lib/gitgitgadget.ts
+++ b/lib/gitgitgadget.ts
@@ -18,13 +18,18 @@ export interface IGitGitGadgetOptions {
     activeMessageIDs?: { [messageID: string]: string };
 }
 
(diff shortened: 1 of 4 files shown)


webstech:
[Sample action](https://github.com/webstech/ggguser/runs/6743021807?check_suite_focus=true) using env vars.  Still a WIP but getting close.

dscho (approved):

dscho:
Eventually, I would like to see us move GitGitGadget to become a GitHub Action.

PR #1023, opened 2022-06-24
Make email repo more generic via config

webstech:
Added to mailrepo:
  - url - link to access files.  Should be buildable at runtime  but that is asking for more than I can currently offer
  - descriptiveName - term to be used for messages

### DIFF
--- a/lib/gitgitgadget-config.ts
+++ b/lib/gitgitgadget-config.ts
@@ -17,6 +17,8 @@ const defaultConfig: IConfig = {
         owner: "gitgitgadget",
         branch: "master",
         host: "lore.kernel.org",
+        url: "https://lore.kernel.org/git",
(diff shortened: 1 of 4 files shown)


webstech:
[gitgitgadget-config](https://github.com/gitgitgadget/gitgitgadget/pull/1023/files#diff-26e2b151673711b88598fafdac0342be6f4c6353f51ce798e6778af017d218a1) has a couple of comments to show what would be desired (ie deferred value resolution).  There may be an elegant solution but it was not obvious in searches.  Another option is to require config providers to include a function to post complete a loaded config.  Have not thought this through too much but it may have implementation issues.  The problem being solved is repetition.

dscho (approved):

dscho:
Thank you!

Issue #1024, opened 2022-06-27
Is it possible to delete useless range-diff?


adlternative:
Hi, sometime we may have a big change from vN to vN+1, the range-diff output
is useless, so is here anyway to suppress its output by some option?
@dscho

dscho:
There is not yet any option, but it is easy to see how this could be valuable.

One strategy (easy to implement) would be a footer in the first comment similar to `Cc:`.

Another strategy (harder to implement) would be to compare the size of the range-diff to the size of the overall patches, and at a certain threshold replace the range-diff with a message like "The range-diff is too huge to be helpful."

All it needs is someone with enough time to implement either strategy (or a strategy I have not thought of, yet).

webstech:
Looking at the second strategy.  Using [this diff](https://lore.kernel.org/git/pull.1250.v2.git.1654778272871.gitgitgadget@gmail.com/) as an example.

`/^\s+[+-][\s\n]/` is the count of changes.
`^\s+[+-]\+[\s\n]/ is the count of range changes.
In this example, there are 42 changes and 60 range changes.  The range changes are 27 ++ and 33 -+.

Possible rules (**VERY** open to suggestions and numbers):
- if the ++ count is  _x%_ greater than the -+ count, a lot of changes were added so the range-diff could still be useful
- if the -+ count is  _x%_ greater than the ++ count, a lot of changes were removed so the range-diff may not be needed
- if the ++ and -+ counts are similar and are _x%_ of the total changes (approximately 40% in the example), then the range-diff may not be needed

Given the possibilities of getting it wrong, the footer comment would still need to be supported but as a boolean: `version-diff: (yes|no)`.

How about this message?
No range-diff provided.  Please review the new patches as a replacement.

I guess that means both strategies. 🤔

dscho:
> Using [this diff](https://lore.kernel.org/git/pull.1250.v2.git.1654778272871.gitgitgadget@gmail.com/) as an example.

That is actually a good example of a useful range-diff that is still larger than the patch itself (I think, at least) because it touches every line.

To me, this implies that there is no good "relative" measure.

Maybe a better measure would simply be the number of characters in the range-diff. Like, if the range-diff is even "only" 5kB, I would wager a bet that it is not very useful because it would be too hard to review.

> No range-diff provided. Please review the new patches as a replacement.

That would make it sound to me as if the contributor did not contribute a range-diff. Which they didn't, but that's not their fault. I would prefer something along these lines:

[...]

PR #1030, opened 2022-06-30
Add feature to suppress range-diff on /submit

webstech:
This change introduces support for a new pull request description
footer:

Range-Diff: false

This will bypass generation of the range-diff between patch versions.
The range-diff output will be replaced with instructions for
generating it locally.  The users of this feature should comment if
the range-diff is not relevant.

## Random bits
- ~~choice of `version-diff` is trying to look at this from a user perspective~~
- readme description could be expanded (usual developer lack of doc)
- tests need to be developed.  Testing was done using webstech/gggmonitored repo.
- change does not include `mail-patch-series` support using `getFromTag`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,8 +26,11 @@ their behalf, and people mentioned in the Cc: footer of the Pull Request
 description.
 
 Furthermore, for all iterations of a patch series but the first one,
-GitGitGadget will insert a machine-generated representation of what changed,
-and reply to the cover letter of the previous iteration.
+GitGitGadget will insert a machine-generated representation of what changed
+between revisions,
+and reply to the cover letter of the previous iteration.  This patch revision
+diff can be supressed if the change may be too large or irrelevant by adding
+a `Range-Diff: false` footer in the Pull Request description.
 
 For convenience of reviewers, GitGitGadget will generate tags for each
 iteration it sent, and push those to [https://github.com/gitgitgadget/git](https://github.com/gitgitgadget/git). Links
--- a/lib/git.ts
+++ b/lib/git.ts
@@ -195,3 +195,8 @@ export async function gitCommandExists(command: string, workDir?: string):
     const result = await GitProcess.exec([command, "-h"], workDir || ".");
     return result.exitCode === 129;
 }
+
+// rev-parse does not have enough info in shallow repos to determine a safe short name
+export function gitShortHash(longHash: string): string {
+    return longHash.substring(0, 8);
+}
--- a/lib/patch-series.ts
+++ b/lib/patch-series.ts
@@ -1,7 +1,7 @@
 import addressparser from "nodemailer/lib/addressparser";
 import mimeFuncs from "nodemailer/lib/mime-funcs";
 // import { encodeWords } from "nodemailer/lib/mime-funcs";
-import { commitExists, git, gitConfig, revListCount, revParse, } from "./git";
+import { commitExists, git, gitConfig, gitShortHash, revListCount, revParse, } from "./git";
 import { GitNotes } from "./git-notes";
 import { IGitGitGadgetOptions } from "./gitgitgadget";
 import { IMailMetadata } from "./mail-metadata";
@@ -49,6 +49,13 @@ const singletonHeaders: ISingletonHeader[] = [
     },
 ];
 
+interface IRangeDiff {
+    previousRange: string;
+    currentRange: string;
+    baseCommit: string;
+    headCommit: string;
+}
+
 export class PatchSeries {
     public static async getFromTag(options: PatchSeriesOptions, project: ProjectOptions): Promise<PatchSeries> {
         const latestTag: string = await this.getLatestTag(project.branchName, options.redo);
@@ -68,7 +75,8 @@ export class PatchSeries {
             headLabel: project.branchName,
             iteration: 1,
         };
-        let rangeDiff = "";
+        let rangeDiffRanges: IRangeDiff | undefined;
+        const currentRange = `${baseCommit}..${headCommit}`;
 
         if (latestTag) {
             const range = latestTag + "..." + project.branchName;
@@ -106,13 +114,19 @@ export class PatchSeries {
                 }
             });
 
-            rangeDiff = await git(["range-diff", "--creation-factor=95", "--no-color", range]);
+            const tagCommit = await revParse(latestTag);
+            if (!tagCommit) {
+                throw new Error(`Cannot determine commit for tag: ${latestTag}`);
+            }
+
+            const previousRange = `${baseCommit}..${tagCommit}`;
+            rangeDiffRanges = { previousRange, currentRange, baseCommit, headCommit };
         }
 
-        const patchCount = await revListCount(["--no-merges", `${baseCommit}..${headCommit}`], project.workDir);
+        const patchCount = await revListCount(["--no-merges", currentRange], project.workDir);
 
         const notes = new GitNotes(project.workDir, "refs/notes/mail-patch-series");
-        return new PatchSeries(notes, options, project, metadata, rangeDiff, patchCount);
+        return new PatchSeries(notes, options, project, metadata, rangeDiffRanges, patchCount);
     }
 
     public static async getFromNotes(notes: GitNotes,
@@ -138,7 +152,7 @@ export class PatchSeries {
             throw new Error(`Invalid commit range: ${currentRange}`);
         }
 
-        let rangeDiff = "";
+        let rangeDiffRanges: IRangeDiff | undefined;
         if (metadata === undefined) {
             metadata = {
                 baseCommit,
@@ -156,8 +170,7 @@ export class PatchSeries {
             }
 
             const previousRange = `${metadata.baseCommit}..${metadata.headCommit}`;
-            rangeDiff = await git(["range-diff", "--no-color", "--creation-factor=95",
-                                    previousRange, currentRange], { workDir });
+            rangeDiffRanges = { previousRange, currentRange, baseCommit: metadata.baseCommit, headCommit };
 
             metadata.iteration++;
             metadata.baseCommit = baseCommit;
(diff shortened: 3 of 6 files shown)


webstech:
Issue #1024 was the impetus for this.  @dscho suggested manual and code to decide.  This provides a manual implementation that can override future automation (bad) decisions.

lib/patch-series.ts
```diff
@@ -718,9 +740,42 @@ export class PatchSeries {
...
+
+                // please set rangeDiff empty if post analysis decides it is too big or messy
+                rangeDiff = await git(["range-diff", "--no-color", "--creation-factor=95", ...args],
+                    { workDir: this.project.workDir });
+            }
+
+            if (rangeDiff) {
+                // split the range-diff and prefix with a space
+                footers.push(`Range-diff vs v${this.metadata.iteration - 1}:\n\n${
+                            rangeDiff.replace(/(^|\n(?!$))/g, "$1 ")}\n`);
```
dscho, line 761: It would probably be better to move that into the previous `if` block, and if the range-diff _is_ empty, offer a different message (e.g. `Range-diff vs v${this.metadata.iteration - 1} is empty`).

lib/patch-series.ts
```diff
@@ -718,9 +740,42 @@ export class PatchSeries {
...
+                    if (hashes) {
+                        return `${hashes[1]}${hashes[2]}${hashes[3]}`;
+                    } else {
+                        throw Error(`Range parse failed for ${range}`);
+                    }
+                };
+
+                footers.push(`Range-diff too large. Please review it locally by calling:
+   git fetch https://github.com/gitgitgadget/git ${this.rangeDiff.baseCommit.substring(0, 7)} ${
+                    this.rangeDiff.headCommit.substring(0, 7)}
```
dscho, line 775: I would like to avoid hard-coding the `7` here, maybe we can do something like adding `gitShortHash(revision)` to `lib/git.ts` and use that here? Oh, and increase the length to 8, to stay on the safe side.

dscho (commented):
I agree that this would be valuable to do. A couple of comments/suggestions below:

README.md
```diff
@@ -26,8 +26,11 @@ their behalf, and people mentioned in the Cc: footer of the Pull Request
 description.
 
 Furthermore, for all iterations of a patch series but the first one,
-GitGitGadget will insert a machine-generated representation of what changed,
-and reply to the cover letter of the previous iteration.
+GitGitGadget will insert a machine-generated representation of what changed
+between revisions,
+and reply to the cover letter of the previous iteration.  This patch revision
+diff can be supressed if the change may be too large or irrelevant by adding
+a `version-diff: no` footer in the Pull Request description.
```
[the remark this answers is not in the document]
webstech: Sure.  I chose a name specific to how range-diff is used here (as described in the command doc) but have no objection to range-diff.

lib/patch-series.ts
webstech: Will the diff ever be empty without an error?  I have adjusted the code. The previous iteration allowed post-range-diff analysis to set rangeDiff to empty but everything can be re-adjusted if that ever happens.

lib/patch-series.ts
```diff
@@ -718,9 +740,42 @@ export class PatchSeries {
...
+                const getRange = (range: string): string => {
+                    const hashes = range.match(/([a-z,0-9]{7}).+?(\.+)([a-z,0-9]{7})/);
+                    if (hashes) {
+                        return `${hashes[1]}${hashes[2]}${hashes[3]}`;
+                    } else {
+                        throw Error(`Range parse failed for ${range}`);
+                    }
+                };
+
+                footers.push(`Range-diff too large. Please review it locally by calling:
```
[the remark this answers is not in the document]
webstech: Please see the latest verbiage.

lib/patch-series.ts
webstech: Please see the updates.

webstech:
Thanks for the review.

lib/patch-series.ts
dscho: It could be empty if a force-push happened after `git commit --amend` without editing the commit message (the commit date would have changed, and therefore the commit hash, but nothing that the range-diff cares about).

lib/git.ts
```diff
@@ -195,3 +195,8 @@ export async function gitCommandExists(command: string, workDir?: string):
     const result = await GitProcess.exec([command, "-h"], workDir || ".");
     return result.exitCode === 129;
 }
+
+// rev-parse does not have info for shallow repos
```
dscho, line 199: How about:

```typescript
// rev-parse does not have enough info in shallow repos to determine a safe short name
```

lib/patch-series.ts
```diff
@@ -106,7 +113,7 @@ export class PatchSeries {
                 }
             });
 
-            rangeDiff = await git(["range-diff", "--creation-factor=95", "--no-color", range]);
+            rangeDiff = { previousRange: range, baseCommit, headCommit };
```
dscho, line 116: I just noticed that this is a `...` range.. Maybe we would want to adjust this to define `previousRange` to `<current-branch>..<latest-tag>` and `currentRange` to `<latest-tag>..<current-branch>`?

lib/patch-series.ts
```diff
@@ -198,6 +205,9 @@ export class PatchSeries {
         const publishToRemote: string | undefined = undefined;
 
         const project = await ProjectOptions.get(workDir, headCommit, cc, basedOn, publishToRemote, baseCommit);
+        if (typeof rangeDiff !== "undefined") {
```
dscho, line 208: Can't we just use `if (rangeDiff)` for simplicity?

lib/patch-series.ts
```diff
@@ -288,6 +303,12 @@ export class PatchSeries {
                                 }
                             });
                             break;
+                        case "range-diff:":
```
dscho, line 306: Thank you for changing it in the code. Could you also adjust the commit message, please?

lib/patch-series.ts
```diff
@@ -288,6 +303,12 @@ export class PatchSeries {
...
+                            if (doRangeDiff) {
+                                throw new Error(`Duplicate version-diff`);
+                            }
+                            doRangeDiff = match2[2].toLowerCase() === "yes" ? true : false;
```
dscho, line 310: Maybe we should use `true` here?

A later PR might imitate Git's config setting strategy where some settings can be Boolean (and which accept `true` and `false` but not `yes`), but also contain more information. For example, I imagine that we could allow passing arbitrary `range-diff` options, such as `--creation-factor=9999` or `-U30`.

lib/patch-series.ts
```diff
@@ -718,9 +740,35 @@ export class PatchSeries {
             if (footers.length > 0) {
                 footers.push(""); // empty line
             }
-            // split the range-diff and prefix with a space
-            footers.push(`Range-diff vs v${this.metadata.iteration - 1}:\n\n${
-                         this.rangeDiff.replace(/(^|\n(?!$))/g, "$1 ")}\n`);
+
+            if (typeof this.options.doRangeDiff === "undefined" || this.options.doRangeDiff) {
```
dscho, line 744: Could we do it the other way round? I do not feel very strongly about it, but I did stumble over this. The order "if it is set to something true-ish, or if it is not set at all" would have helped me.

lib/patch-series.ts
```diff
@@ -718,9 +740,35 @@ export class PatchSeries {
...
+                const getRange = (range: string): string => {
+                    const hashes = range.match(/([a-z,0-9]+?)(\.+)([a-z,0-9]+)/);
+                    if (hashes) {
+                        return `${gitShortHash(hashes[1])}${hashes[2]}${gitShortHash(hashes[3])}`;
+                    } else {
+                        throw Error(`Range parse failed for ${range}`);
+                    }
+                };
+
+                footers.push(`Contributor requested no range-diff. You can review it using these commands:
```
dscho, line 766: Wonderful!

dscho (commented):
Thank you for the updates! Just a couple minor things, still, and I think we're good to go.

adlternative:
I am not a TypeScript user, but thanks to both of you! 👍

lib/patch-series.ts
webstech: After `git commit --amend --no-edit`:

```
git range-diff --creation-factor=95 f939873..64efcea5 f939873..a3b9bf7
1:  64efcea = 1:  a3b9bf7 Versioning a submit
```
So it still tells us if the commits are the same.

lib/patch-series.ts
webstech: No, that was an error on my part during renames (thanks for catching it).  It is supposed to be a check on `doRangeDiff`, which may not be defined.  Since it is a boolean, existence needs to be checked.

lib/patch-series.ts
webstech: Doh!

lib/patch-series.ts
webstech: Okay.  Converting to a string.  This affects the next comment.

lib/patch-series.ts
webstech: Need to know if present before checking a boolean.  Awkward so I  am changing the if/else order (even though it is now a string and I prefer the more positive side of the if first).

lib/patch-series.ts
webstech: I am glad you brought this up.  I was meaning to ask.  Do I need to convert this range to hashes?  `gitShortHash` could just return the value if it does not look like a hash (or do not call `gitShortHash` if it does not look like a hash.  Certainly, changing to specify `currentRange` will make other code simpler.  Will other reviewers be able to get the tags if they are shown in the 'no range-diff' message?

lib/patch-series.ts
webstech: > Since it is a boolean, existence needs to be checked.

Except now it is a string!

dscho (approved):
Great, thank you!

PR #1037, opened 2022-07-11
"Bug report " urls and line length exception advice

PhilipOakley:
The GitGitGadget lint checker can leave users in a Catch 22 situation
without advice on how best to resolve their reasonable conundrum.
E.g gitgitgadget/git Issue #1283 used a "Bug report " Long_URL which
reported such a fail.

1. Allow sufficient characters to fit the "Bug report " into the long
   line preamble.
2. Add guidance to the error message on allowable exceptions.

Signed-off-by: Philip Oakley <philipoakley@iee.email>

### DIFF
--- a/lib/commit-lint.ts
+++ b/lib/commit-lint.ts
@@ -114,8 +114,9 @@ export class LintCommit {
                 // Allow long lines if prefixed with whitespace (ex. quoted error messages)
                 (! this.lines[i].match(/^\s+/)) &&
                 // Allow long lines if they cannot be wrapped at some
-                // white-space character, e.g. URLs. To allow ` [1] <URL>`
-                // lines, we skip the first 10 characters.
+                // white-space character, e.g. URLs. To allow a short
+                // preamble such as `ref [1] <URL>` lines, we skip the
+                // first 10 (arbitrary) characters.
                 this.lines[i].slice(10).match(/\s/)) {
                 this.block(`Lines in the body of the commit messages ${""
                     }should be wrapped between 60 and ${
--- a/lib/commit-lint.ts
+++ b/lib/commit-lint.ts
@@ -120,7 +120,8 @@ export class LintCommit {
                 this.lines[i].slice(10).match(/\s/)) {
                 this.block(`Lines in the body of the commit messages ${""
                     }should be wrapped between 60 and ${
-                    this.maxColumns} characters.`);
+                    this.maxColumns} characters.\n`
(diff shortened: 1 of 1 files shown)


lib/commit-lint.ts
```diff
@@ -115,8 +115,8 @@ export class LintCommit {
                 (! this.lines[i].match(/^\s+/)) &&
                 // Allow long lines if they cannot be wrapped at some
                 // white-space character, e.g. URLs. To allow ` [1] <URL>`
-                // lines, we skip the first 10 characters.
-                this.lines[i].slice(10).match(/\s/)) {
+                // lines, we skip the first 11 characters ("Bug report ").
+                this.lines[i].slice(11).match(/\s/)) {
```
dscho, line 119: I don't think this is a good thing to do. You could just as easily write `Bug report:` into its own line and put the overly-long line into the next line. And that allows for things like `The problem was reported at:` (which is much longer than 11 columns, but still informative and short).

lib/commit-lint.ts
```diff
@@ -119,7 +119,8 @@ export class LintCommit {
                 this.lines[i].slice(11).match(/\s/)) {
                 this.block(`Lines in the body of the commit messages ${""
                     }should be wrapped between 60 and ${
-                    this.maxColumns} characters.`);
+                    this.maxColumns} characters.\n`
+                     + `Exceptions are: indented lines and very long URLs/words.`);
```
dscho, line 123: This is misleading. We do not allow very long URLs or words unless they are in their own line.

PhilipOakley: `"Single" words` maybe. I was just rephrasing the 'URL' example in the comment lines above.

Did you have a particular suggestion?

dscho: How about "Indented lines, and lines without whitespace, are exempt from this check."?

I don't know. _You_ are the native speaker, not I.

dscho: "and very long URLs/words" makes it sound as if the lines could still contain whitespace.

PhilipOakley: I've updated, rebased and pushed the branch. I dropped the plurals (which could apply to the whole message being linted) and made the commit message be about individual 'lines'. I also included that the error message avoids mention of the preamble that can contain spaces.

Not quite sure which of the GitHub buttons needs pressing to get the PR to update for GGG.

dscho (approved):
Sounds good!

PhilipOakley:
Thanks!

PR #1055, opened 2022-07-20
Add two sources to support use as actions

webstech:
A couple of thin layers to allow gitgitgadget functions to be accessed from GitHub actions.  The actions are similar to misc-helper but more specialized.  Currently, nothing is published as a package.  Action providers must  run `npm pack` on gitgitgadget and specify the package on the `npm install` command in the action repo.

`pull-action` is used for pull request related events (create, synchronize, comment).

`misc-action` is used for other scheduled events such as checking mailing list emails.

Add *.tgz to .gitignore for untracked package.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@
 /obj/**
 /*.sln*
 /*.njsproj*
+/*.tgz
--- a/lib/misc-action.ts
+++ b/lib/misc-action.ts
@@ -1,0 +1,82 @@
+import { CIHelper } from "./ci-helper";
+import { isDirectory } from "./fs-util";
+import { getConfig } from "./gitgitgadget-config";
+import { getVar } from "./gitgitgadget";
+import { IConfig, loadConfig, setConfig } from "./project-config";
+import path from "path";
+
+export interface actionInterface {
+    action: string;
+    repositoryDir: string;
+    configRepositoryDir: string;
+    config: string;
+    repoOwner: string;
+    repoName: string;
+    skipUpdate?: string | undefined;
+}
+
+/**
+ * Handle various gitgitgadget requests.
+ *
+ * @param parms
+ */
+
+export async function handleAction(parms: actionInterface): Promise<void> {
+    const config: IConfig = parms.config ? setConfig(await getExternalConfig(parms.config)) : getConfig();
+
+    // Update with current values
+    config.repo.name = parms.repoName;
+    config.repo.owner = parms.repoOwner;
+    setConfig(config);
+
+    lintConfig(config);
+
+    if (!(await isDirectory(parms.repositoryDir))) {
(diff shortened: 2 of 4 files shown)


webstech:
[Sample repo](https://github.com/webstech/gitgitgadget-pr-action) with two actions using the new functions.  These could (should) be part of gitgitgadget.  The doc suggests all actions should be in separate repos but it is not mandatory.  Since these could be used by external projects using gitgitgadget, should they be in separate projects?

dscho (approved):
Thank you so much for pushing this along!

dscho:
Since such a lot of code is shared, I would think that it makes sense to keep things in a single repository.

I would be in favor of refactoring so that it is a single Action running in different modes.

Eventually, I want to use `ncc` to pack everything into a single (committed) `.js` file, much like I do it in `setup-git-for-windows-sdk`: https://github.com/git-for-windows/setup-git-for-windows-sdk/commits/7da470f0d3b358ecffc95788839430faab761963/dist

webstech:
> Since such a lot of code is shared, I would think that it makes sense to keep things in a single repository.

The actions in the sample have a duplicated `setupGitEnvironment()` but that is about it.  The reason for separate repos is to keep the workflow invocation cleaner.  This is a bit verbose:
```
uses: webstech/gitgitgadget-pr-action/.github/actions/pull-request@main
```
(location suggested in GitHub docs for multi-action repo).

> I would be in favor of refactoring so that it is a single Action running in different modes.

The pull request action has different inputs so it should have its own `action.yml`.  The code could check what action was running and do the appropriate work but it does make it more complex.

> Eventually, I want to use `ncc`

The sample is doing this.  The outputs are in the same folders as the `action.yml`.

dscho:
> > Since such a lot of code is shared, I would think that it makes sense to keep things in a single repository.
> 
> The actions in the sample have a duplicated `setupGitEnvironment()` but that is about it. The reason for separate repos is to keep the workflow invocation cleaner. This is a bit verbose:
> 
> ```
> uses: webstech/gitgitgadget-pr-action/.github/actions/pull-request@main
> ```
> 
> (location suggested in GitHub docs for multi-action repo).

I agree, this would be ugly.

But maybe we can use the shorter names `gitgitgadget/pr` and `gitgitgadget/push`? FWIW feel free to move your repository into the `gitgitgadget` org, unless you want to play with it more before moving it.

> > would be in favor of refactoring so that it is a single Action running in different modes.
> 
> The pull request action has different inputs so it should have its own `action.yml`. The code could check what action was running and do the appropriate work but it does make it more complex.

Now I understand. And I agree with your reasoning. Thank you for being patient with me.

> > Eventually, I want to use `ncc`
> 
> The sample is doing this. The outputs are in the same folders as the `action.yml`.

I totally missed this (I was on my phone, I did not work last week). It looks beautiful.

PR #1057, opened 2022-07-21
parseMBox should match first colon in header

webstech:
Date headers will now be parsed correctly.

Noticed bad output in the log.


dscho (approved):
Obviously correct!

PR #1112, opened 2022-09-27
maint: update package-lock.json and GitHubGlue deprecated parameter

webstech:
Two commits because one update required the other.

1. maint: update package-lock.json
    
    Node v18 starts to use the http fetch api by default.  This caused a problem for octokit when running the test suite.  This update includes the octokit version with a fix.

2. maint: update GitHubGlue for deprecated parameter
    
    pulls.createReviewComment api has deprecated `position` for `line`. Unfortunately, the typing is requiring `line` so may as well change to it.  Easier than trying to get the types corrected.

### DIFF
--- a/lib/github-glue.ts
+++ b/lib/github-glue.ts
@@ -209,7 +209,7 @@ export class GitHubGlue {
             body: comment,
             commit_id: commit,
             path,
-            position: 1,
+            line: 1,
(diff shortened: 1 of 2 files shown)


dscho (approved):

dscho:
Thank you!!!

PR #1113, opened 2022-09-28
Do not break actions if comment is not found

webstech:
Occasionally a comment may be deleted before it is seen by ggg.  This should not cause a failed action to show on the GitHub dashboard.

### DIFF
--- a/lib/ci-helper.ts
+++ b/lib/ci-helper.ts
@@ -5,7 +5,7 @@ import { ILintError, LintCommit } from "./commit-lint";
 import { commitExists, git, emptyTreeName } from "./git";
 import { GitNotes } from "./git-notes";
 import { GitGitGadget, IGitGitGadgetOptions } from "./gitgitgadget";
-import { GitHubGlue, IGitHubUser, IPRCommit, IPullRequestInfo } from "./github-glue";
+import { GitHubGlue, IGitHubUser, IPRComment, IPRCommit, IPullRequestInfo, RequestError } from "./github-glue";
 import { toPrettyJSON } from "./json-util";
 import { MailArchiveGitHelper } from "./mail-archive-helper";
 import { MailCommitMapping } from "./mail-commit-mapping";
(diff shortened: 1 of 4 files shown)


lib/ci-helper.ts
```diff
@@ -521,7 +521,16 @@ GitGitGadget needs an email address to Cc: you on your contribution, so that you
      * @param commentID the ID of the PR comment to handle
      */
     public async handleComment(repositoryOwner: string, commentID: number): Promise<void> {
-        const comment = await this.github.getPRComment(repositoryOwner, commentID);
+        let comment: IPRComment;
+
+        try {
+            comment = await this.github.getPRComment(repositoryOwner, commentID);
+        } catch (e) {
+            const error = e as Error;
```
dscho, line 529: I wonder whether we can ensure that this exception was thrown because the comment was not found, i.e. a 404?

We would still want the run to fail upon, say, 500...

dscho: Judging by [this code](https://github.com/gitgitgadget/gitgitgadget/blob/403cc505b4749d866bbccdc472754b078215661a/tests/github-glue.test.ts#L526), I believe we should be able to filter by this via `if (e.status === 404) [...]`.

webstech: > Judging by [this code](https://github.com/gitgitgadget/gitgitgadget/blob/403cc505b4749d866bbccdc472754b078215661a/tests/github-glue.test.ts#L526), I believe we should be able to filter by this via `if (e.status === 404) [...]`.

Updated to check the status and check it is a `RequestError`.  I chose to export `RequestError` from `github-glue.ts` to hide octokit from `ci-helper.ts`.  If you prefer a direct import I can change it (back).

Tested by running the code as part of `github-glue.test.ts` with a bad comment number.

dscho: This is excellent!

dscho (approved):

PR #1128, opened 2022-10-14
Format checkboxes

rscharfe:
GitHub task lists are represented in HTML as unordered lists with checkboxes.  These checkboxes are currently ignored when converting the HTML to text.  This affects PRs with task lists (e.g. https://github.com/git/git/pull/1359); the resulting text list has no indication anymore whether an item was checked (done) or not.

E.g. this:

- [x] done item
- [ ] item still to do

Is turned into:

` * done item`
`* item still to do`

Format checkboxes by turning checked ones into "[x]" and unchecked ones into "[ ]" to retain that information during the conversion. This turns the text result for the example above into:

` * [x] done item`
` * [ ] item still to do`

Unfortunately the simple formatter callback requires two ESLint annotations to avoid a warning and an error because it doesn't use all of its parameters and needs to access elem.attribs, which is untyped for some reason.  Hopefully there's a better way, but that's the best I  can do for now.


dscho:
@rscharfe I force-pushed your branch, please review.

<details><summary>Range-diff</summary>



* 1:  e5b1da3 ! 1:  df287d1 Format checkboxes

   ``````diff
   @@
     ## Metadata ##
   -Author: rscharfe <l.s.r@web.de>
   +Author: René Scharfe <l.s.r@web.de>
    
     ## Commit message ##
        Format checkboxes
   @@ Commit message
        for some reason.  Hopefully there's a better way, but that's the best I
        can do for now.
    
   +    Test-added-by: Johannes Schindelin <johannes.schindelin@gmx.de>
   +    Signed-off-by: René Scharfe <l.s.r@web.de>
   +    Signed-off-by: Johannes Schindelin <johannes.schindelin@gmx.de>
   +
     ## lib/markdown-renderer.ts ##
    @@ lib/markdown-renderer.ts: export function md2text(markdown: string, columns = 76): string {
                             .replace(/(^|\n)(\n)(?!$)/g, "$1>$2"); // quote empty
                         }});
                 },
   -+            // eslint-disable-next-line @typescript-eslint/no-unused-vars
   -+            checkBoxFormatter: (elem, _walk, builder, _options) => {
   -+                // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
   -+                builder.addInline(elem.attribs.checked === undefined ? "[ ]" : "[x]");
   ++            checkBoxFormatter: (elem, _walk, builder) => {
   ++                const attribs = elem.attribs as { checked?: boolean };
   ++                builder.addInline(attribs.checked === undefined ? "[ ]" : "[x]");
    +            },
             },
             selectors: [
   @@ lib/markdown-renderer.ts: export function md2text(markdown: string, columns = 76
             ],
         };
     
   +
   + ## tests/markdown-renderer.test.ts ##
   +@@ tests/markdown-renderer.test.ts: const bq7 = `Over 56 levels still has 20 char (exact):
   + >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> 3 5 7 9 1 3 5 7 9 1
   + >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> 3 5 7`);
   + });
   ++
   ++test("task lists are rendered correctly", () => {
   ++    const taskList = `This is a task list:
   ++* [x] done item
   ++* [ ] item still to do`;
   ++    expect(md2text(taskList)).toEqual(`This is a task list:
   ++
   ++ * [x] done item
   ++ * [ ] item still to do`);
   ++});
   + \ No newline at end of file
   ``````
</details>

lib/markdown-renderer.ts
```diff
@@ -36,6 +36,10 @@ export function md2text(markdown: string, columns = 76): string {
                         .replace(/(^|\n)(\n)(?!$)/g, "$1>$2"); // quote empty
                     }});
             },
+            checkBoxFormatter: (elem, _walk, builder) => {
+                const attribs = elem.attribs as { checked?: boolean };
```
rscharfe, line 40: The cast looks tricky -- hopefully just to me, lacking TypeScript experience. It would be nice if the type of elem could be inferred from the caller of this callback, but that's probably a bridge too far. elem.attrib also has a property named "type" (not optional, it seems), but we can simply ignore it here, I guess? https://github.com/html-to-text/node-html-to-text/blob/ba8697718d9995efbfe6255fe27d629e9d556b0b/lib/formatter.js#L147 has "const attribs = elem.attribs || {};"; would that work as well?

rscharfe (commented):
The paragraph lamenting the ESLint annotations should be replaced by a note saying that you fixed them, Dscho.

I give two more line-specific comments, hopefully just demonstrating lack of experience with JavaScript and TypeScript.

Thank you for the test!

lib/markdown-renderer.ts
```diff
@@ -36,6 +36,10 @@ export function md2text(markdown: string, columns = 76): string {
                         .replace(/(^|\n)(\n)(?!$)/g, "$1>$2"); // quote empty
                     }});
             },
+            checkBoxFormatter: (elem, _walk, builder) => {
```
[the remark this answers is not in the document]
dscho: Yes, if you are used to C it looks terrifying to "just drop" a parameter. Yet that's totally normal in Javascript.

lib/markdown-renderer.ts
dscho: This cast basically says that we know that this `any` type is actually an object that has an optional `checked` attribute.

However, there is a bug: the attribute has type string ;-)

Will fix.

dscho (approved):
@rscharfe please have another look. As soon as you give the thumbs-up, I'll merge.

rscharfe (commented):
Looks good to me! Indeed, checked is either absent or a string.

PR #1141, opened 2022-10-28
lib: prepare for an interim maintainer

ttaylorr:
This pull request prepares GitGitGadget for the interim maintainer. The changes here are broken out into two sets:

- First, a handful of patches which replace "gitster" in comments and code with "maintainer", but does not actually change the location from gitster/git.
- Then, a final patch which replaces the location of the upstream gitster/git with ttaylorr/git.

The final patch 572cb9316f5ddec710243d720955153318ae1794 should be reverted once Junio is back online.

##

/cc @dscho

### DIFF
--- a/script/lookup-commit.sh
+++ b/script/lookup-commit.sh
@@ -239,7 +239,7 @@ test notes != "$mode" || {
 
 		# gitk,git-gui patches are often taken via Pull Requests, let's look only at patches from known git.git (interim) maintainers
 		case "$(git -C "$GITGIT_DIR" show -s --format=%cn "$commit")" in
-		"Linus Torvalds"|"Junio C Hamano"|"Shawn O. Pearce"|"Jeff King"|"Jonathan Nieder") ;; # these should be mostly on the mailing list
+		"Linus Torvalds"|"Junio C Hamano"|"Shawn O. Pearce"|"Jeff King"|"Jonathan Nieder"|"Taylor Blau") ;; # these should be mostly on the mailing list
 		*) continue;; # these are not expected to have been discussed on the mailing list
 		esac
 
--- a/script/lookup-commit.sh
+++ b/script/lookup-commit.sh
@@ -70,10 +70,12 @@ update_gitgit_dir () {
 
 	if git -C "$GITGIT_DIR" rev-parse --verify refs/remotes/gitster/seen >/dev/null 2>&1
 	then
+		# Let's take 'seen' from the official source at git.git.
+		git -C "$GITGIT_DIR" remote set-url gitster https://github.com/git/git
 		git -C "$GITGIT_DIR" fetch gitster ||
 		die "Could not update the 'gitster' remote to $GITGIT_DIR"
 	else
-		git -C "$GITGIT_DIR" remote add -f gitster https://github.com/gitster/git ||
+		git -C "$GITGIT_DIR" remote add -f gitster https://github.com/git/git ||
 		die "Could not add the 'gitster' remote to $GITGIT_DIR"
 	fi
 }
--- a/script/lookup-commit.sh
+++ b/script/lookup-commit.sh
@@ -237,15 +237,6 @@ test notes != "$mode" || {
 			continue
 		fi
 
-		amlog="$(git -C "$GITGIT_DIR" notes --ref=refs/notes/gitster-amlog show "$commit" 2>/dev/null | head -n 1)"
-		if test -n "$amlog"
-		then
-			amlog=${amlog%>}
-			amlog=${amlog#Message-Id: <}
-			printf 'M 100644 inline %s\ndata <<EOF\n%s\nEOF\n' "$commitpath" "$amlog"
-			continue
-		fi
-
 		# gitk,git-gui patches are often taken via Pull Requests, let's look only at patches from known git.git (interim) maintainers
 		case "$(git -C "$GITGIT_DIR" show -s --format=%cn "$commit")" in
 		"Linus Torvalds"|"Junio C Hamano"|"Shawn O. Pearce"|"Jeff King"|"Jonathan Nieder") ;; # these should be mostly on the mailing list


dscho:
Hey @ttaylorr, thank you for thinking about adjusting GitGitGadget, too.

However, since this is only about an interim maintainership, I am somewhat concerned that we're changing a lot of code here, with all kinds of ramifications to code stability, for at most temporary benefit.

Also, a _lot_ about GitGitGadget is actually outside of this source code. For example, the branches are mirrored from `gitster/git` into `gitgitgadget/git` via https://dev.azure.com/gitgitgadget/git/_build?definitionId=8, which would need to be adjusted separately. For the record, this Pipeline is _not_ a GitHub workflow because Azure Pipelines can be triggered on repositories owned by someone else, while GitHub workflows are required to live in that same code base and therefore we would have to ask the Git maintainer (whether interim or not) to accept a file into `.github/workflows/` of `gitster/git`'s default branch solely for the benefit of the `gitgitgadget/git` fork, which I doubt would be accepted, so I did not even try.

I also have to admit that I doubt that "gitster" -> "maintainer" is a descriptive rename. The short-hand "gitster branch" stands for "a branch that is not available in `git/git` but that is nevertheless a topic branch created from individual patch series on the Git mailing list and that are updated with new iterations and that usually live in `gitster`'s personal fork". Reading "maintainer branch" I think about something completely different: a branch designed to help the maintainer and nobody else, i.e. something like [that infamous `todo` branch](https://github.com/git/git/commits/todo). I understand that naming is hard, which is why it needs to be done carefully.

lib/patch-series-metadata.ts
```diff
@@ -9,7 +9,7 @@ export interface IPatchSeriesMetadata {
     latestTag?: string;
     referencesMessageIds?: string[];
     tipCommitInGitGit?: string;
-    branchNameInGitsterGit?: string;
+    branchNameInMaintainerGit?: string;
```
dscho, line 12: @ttaylorr this would break GitGitGadget: this interface is a description of the data that is stored in [the `refs/notes/gitgitgadget`](https://github.com/gitgitgadget/git/commits/refs/notes/gitgitgadget) notes.

To see an example, take a peek at https://github.com/gitgitgadget/git/blob/8e58afdf46074debd5cf65693f2906d6dc67cf02/4c/c9/8ffa1e29134751b57d34fc8dd723bf7db1cc

If you rename the field in this interface (and in all of its users in GitGitGadget's code base), every time GitGitGadget reads the notes it will assume that that information is missing. That's a bug.

script/lookup-commit.sh
```diff
@@ -235,7 +235,7 @@ test notes != "$mode" || {
 			continue
 		fi
 
-		amlog="$(git -C "$GITGIT_DIR" notes --ref=refs/notes/gitster-amlog show "$commit" 2>/dev/null | head -n 1)"
+		amlog="$(git -C "$GITGIT_DIR" notes --ref=refs/notes/maintainer-amlog show "$commit" 2>/dev/null | head -n 1)"
```
dscho, line 238: A quick `git grep amlog` will reveal that this ref is updated outside of the `gitgitgadget/gitgitgadget` code base. Therefore, changing the name here will simply break the script.

DESIGN.md
```diff
@@ -112,7 +112,7 @@ automatically.
 
 ### Background tasks
 
-The repositories <https://github.com/gitster/git> (and
+The repositories <https://github.com/ttaylorr/git> (and
```
dscho, line 115: GitGitGadget was designed long before you were designated interim maintainer. Therefore, this change is incorrect.

ttaylorr:
> I also have to admit that I doubt that "gitster" -> "maintainer" is a descriptive rename.

My feeling was that it would be odd (even in the short term) to have a variable called `gitsterBranch` that points at "ttaylorr/git". If you don't think that would be awkward, you could just pick up something that resembles just the final patch.

It's no problem if you don't want to take this PR, I just figured that it would be useful to keep GGG up-to-date with the new, temporary locations while they exist.

script/lookup-commit.sh
```diff
@@ -73,7 +73,7 @@ update_gitgit_dir () {
 		git -C "$GITGIT_DIR" fetch maintainer ||
 		die "Could not update the 'maintainer' remote to $GITGIT_DIR"
 	else
-		git -C "$GITGIT_DIR" remote add -f maintainer https://github.com/gitster/git ||
+		git -C "$GITGIT_DIR" remote add -f maintainer https://github.com/ttaylorr/git ||
```
dscho, line 76: @ttaylorr On its own, this would not have the intended effect because it only initializes the remote if it does not exist yet.

Since you _also_ want to change the nickname of the remote, it would work this time.

But once Junio takes back the reigns, it would no longer be as easy as reverting this change because then the `maintainer` remote is already registered and will pull from the wrong fork!

A better idea would probably be to update this script (which _is_ used in https://dev.azure.com/gitgitgadget/git/_build?definitionId=9) to install a `ttaylorr` remote and (temporarily) use that one, but keep the name `gitster` for the existing functionality and use it again after the revert.

However, I think you missed https://github.com/gitgitgadget/gitgitgadget/blob/609d70813de217a8385e7405eb49f92481885d0c/script/lookup-commit.sh#L247-L251 which _does_ need to be updated.

FWIW this seems to be [the only instance where GitGitGadget's source code mentions the word "interim"](https://github.com/search?q=repo%3Agitgitgadget%2Fgitgitgadget%20interim&type=code).

dscho (changes requested):
Some of these changes would introduce bugs, which makes me wonder how useful it would _actually_ be to rename everything with "gitster" in it. Since we're talking about interim maintainership, it would probably be a much better idea to change the bare minimum necessary.

This bare minimum includes, off the top of my head:
- (temporarily) changing the Azure Pipeline that mirrors `gitster/git` to `gitgitgadget/git`
- updating the `lookup-commit.sh` script to fetch `seen` from `git/git` instead of `gitster/git`

As a bonus, the `lookup-commit.sh` script could lose the code looking at `gitster-amlog` because that ref has turned out to be highly unreliable in identifying commit <-> mail mappings.

lib/ci-helper.ts
```diff
@@ -319,30 +319,30 @@ export class CIHelper {
         const maintainerBranch = `refs/remotes/${this.config.repo.maintainerBranch}/`;
         const maintainerRepo = `${this.config.repo.maintainerBranch}.${this.config.repo.name}`;
 
-        let gitsterBranch: string | undefined =
+        let maintainerBranch: string | undefined =
```
dscho, line 322: FWIW the variable named `maintainerBranch` has already been declared 3 lines above, but it refers to the `refs/remotes/gitster/` prefix.

That is the culprit why all CI builds fail, see e.g. https://github.com/gitgitgadget/gitgitgadget/actions/runs/3345902406/jobs/5542023393#step:9:16

dscho:
> This bare minimum includes, off the top of my head:
> 
>     * (temporarily) changing the Azure Pipeline that mirrors `gitster/git` to `gitgitgadget/git`
> 
>     * updating the `lookup-commit.sh` script to fetch `seen` from `git/git` instead of `gitster/git`

And of course I forgot to add:

- add your name to https://github.com/gitgitgadget/gitgitgadget/blob/609d70813de217a8385e7405eb49f92481885d0c/script/lookup-commit.sh#L249

But then, this is only a work-around for patches like git/git@45c9f05c44b1cb6bd2d6cb95a22cf5e3d21d5b63, i.e. patches that have never seen the Git mailing list at all. So it's probably not all that crucial.

The entire `lookup-commit.sh` business is not super crucial, anyway, because the commit <-> mail mapping is notoriously fragile (because Junio's processes do not accommodate for the need of people or tools like GitGitGadget to have a reliable mapping in order to track the provenance of code changes). I would not mind leaving it "broken" for the few weeks the Git maintainer is away.

dscho:
> it would be odd (even in the short term) to have a variable called `gitsterBranch` that points at "ttaylorr/git".

@ttaylorr `gitsterBranch` does not point to a branch in `ttaylorr/git`. It points to a (mirror) branch in `gitgitgadget/git`.

ttaylorr:
@dscho: I pushed up the result of our impromptu pairing session to this branch, please have a look at a more minimal approach!

script/lookup-commit.sh
```diff
@@ -237,15 +237,6 @@ test notes != "$mode" || {
 			continue
 		fi
 
-		amlog="$(git -C "$GITGIT_DIR" notes --ref=refs/notes/gitster-amlog show "$commit" 2>/dev/null | head -n 1)"
```
dscho, line 238: For the record: `gitster-amlog` was not fetched automatically by any code in GitGitGadget or its Azure Pipelines for quite a while, therefore this code was functionally dead anyway.

dscho (approved):

dscho:
> * (temporarily) changing the Azure Pipeline that mirrors `gitster/git` to `gitgitgadget/git`

@ttaylorr changed: https://dev.azure.com/gitgitgadget/git/_build?definitionId=8

PR #1147, opened 2022-11-05
Fix the URL of the "This branch is now known as <name>" links

dscho:
I noticed that https://github.com/gitgitgadget/git/pull/1404#issuecomment-1304362150 pointed to an incorrect location (hence fixed) and used the opportunity to fix an underlying problem: we should really point to the branches mirrored into GitGitGadget's fork, from where GitGitGadget determines which maintainer branch corresponds to the respective PR branch, anyway.

### DIFF
--- a/lib/ci-helper.ts
+++ b/lib/ci-helper.ts
@@ -317,7 +317,7 @@ export class CIHelper {
 
         // Identify branch in maintainer repo
         const maintainerBranch = `refs/remotes/${this.config.repo.maintainerBranch}/`;
-        const maintainerRepo = `${this.config.repo.maintainerBranch}.${this.config.repo.name}`;
+        const maintainerRepo = `${this.config.repo.owner}/${this.config.repo.name}`;
 
         let gitsterBranch: string | undefined =
             await git(["for-each-ref", `--points-at=${tipCommitInGitGit}`,


lib/ci-helper.ts
```diff
@@ -317,7 +317,7 @@ export class CIHelper {
 
         // Identify branch in maintainer repo
         const maintainerBranch = `refs/remotes/${this.config.repo.maintainerBranch}/`;
-        const maintainerRepo = `${this.config.repo.maintainerBranch}.${this.config.repo.name}`;
+        const maintainerRepo = `${this.config.repo.owner}/${this.config.repo.name}`;
```
webstech, line 320: Unfortunately, now line 329 is not correct.  Do you want to fix that?

dscho: You mean because it mentions `gitgitgadget/git` to contain two separate branches pointing to the same commit?

I would think the comment now to be "more correct": We did have instances in the past where the Azure Pipeline mirrored a branch with a typo in its name from `gitster/git` to `gitgitgadget/git`, later the typo was fixed and the branch with the fixed name was mirrored, too, but while in `gitster/git` the branch with the incorrect name was deleted, it remained in the `gitgitgadget/git` fork. In such an instance, the comment should not claim that there are two candidate branches in `gitster/git` when they really exist only in `gitgitgadget/git`.

webstech: > Unfortunately, now line 329 is not correct.

Okay, maybe not.  I went back to PR #887 where I introduced this bug and see that that line had a slash as well.  It seems there were a couple of other places (`misc-helper.ts`) where `git.git` was used and I probably copied from there.  Sorry my multiple reviews of the changes missed this.

webstech (approved):
Thanks for finding this.

lib/ci-helper.ts
dscho: No apologies necessary, we all work together to make GitGitGadget better, right?

PR #1180, opened 2022-12-12
Update node and actions versions in workflow

webstech:
Use node 18 LTS and latest versions of the workflow actions.

This will get rid of the warnings in the actions jobs.


dscho (approved):
Thank you!

PR #1187, opened 2022-12-21
mail archive - catch comment errors

webstech:
Email comments on specific commits (not 0 in series) fail if the first line is not part of the change.  This is part of a couple of patches to get past this problem.  Failed commit commenting will be a PR comment.   First step to make sure comments are not missed.

Future patches will 
- track the initial changed line in a patch in the metadata
- check if the commit is no longer part of the patch

Noting this is not tested by the CI at this point.  This patch is to ensure emails are added as comments.

### DIFF
--- a/lib/mail-archive-helper.ts
+++ b/lib/mail-archive-helper.ts
@@ -202,25 +202,31 @@ export class MailArchiveGitHelper {
                     (parsedMbox.from ?
                      parsedMbox.from.replace(/ *<.*>/, "") : "Somebody") +
                      ` wrote ([reply to this](${replyToThisURL})):\n\n`;
-                const comment = header +
-                    MailArchiveGitHelper.mbox2markdown(parsedMbox);
+                const comment = MailArchiveGitHelper.mbox2markdown(parsedMbox);
+                const fullComment = header + comment;
 
                 if (issueCommentId) {
                     await this.githubGlue.addPRCommentReply(pullRequestURL,
                                                             issueCommentId,
-                                                            comment);
+                                                            fullComment);
                 } else if (originalCommit) {
-                    const result = await this.githubGlue
-                        .addPRCommitComment(pullRequestURL, originalCommit,
-                                            this.gggNotes.workDir, comment);
-                    issueCommentId = result.id;
+                    try {
+                        const result = await this.githubGlue
+                            .addPRCommitComment(pullRequestURL, originalCommit,
+                                                this.gggNotes.workDir, fullComment);
+                        issueCommentId = result.id;
+                    } catch (error) {
+                        const regarding = `${header.slice(0,-3)}, regarding ${originalCommit}:\n\n`;
+                        await this.githubGlue.addPRComment(pullRequestURL, regarding + comment);
+                        originalCommit = undefined;
+                    }
                 } else {
                     /*
                      * We will not use the ID of this comment, as it is an
                      * issue comment, really, not a Pull Request comment.
                      */
                     await this.githubGlue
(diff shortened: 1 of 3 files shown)


webstech:
Not sure how the package updates got in this patch.  @dscho  Let me know if you want them removed.

dscho:
> Not sure how the package updates got in this patch. @dscho Let me know if you want them removed.

No worries!

lib/mail-archive-helper.ts
```diff
@@ -202,25 +202,31 @@ export class MailArchiveGitHelper {
...
-                    issueCommentId = result.id;
+                    try {
+                        const result = await this.githubGlue
+                            .addPRCommitComment(pullRequestURL, originalCommit,
+                                                this.gggNotes.workDir, fullComment);
+                        issueCommentId = result.id;
+                    } catch (error) {
+                        const regarding = `${header.slice(0,-3)}, regarding ${originalCommit}:\n\n`;
+                        const result = await this.githubGlue.addPRComment(pullRequestURL, regarding + comment);
+                        issueCommentId = result.id;
```
dscho, line 221: I would prefer to omit setting `issueCommentId` and instead to set `originalCommit = undefined` here, so that we [do not record the comment as being attached to a commit](https://github.com/gitgitgadget/gitgitgadget/blob/7de60a2dba77d4b0c138ee375da68a655090b4d5/lib/mail-archive-helper.ts#L235-L240).

From a cursory look at the code, it seems as if it still would work, but it just looks wrong that we detect commit-attached replies by their `issueCommentId` and record that for a comment that was _not_ attached to a commit.

dscho (changes requested):

lib/mail-archive-helper.ts
dscho: > I would prefer to omit setting `issueCommentId`

BTW this is precisely what we do in the [already-existing `addPRComment()` call](https://github.com/gitgitgadget/gitgitgadget/blob/7de60a2dba77d4b0c138ee375da68a655090b4d5/lib/mail-archive-helper.ts#L223-L230).

dscho: > it seems as if it still would work

Oh, I take that back!

Imagine that a comment was posted, as a regular "issue comment" because the "PR review comment" did not work because the commit was force-pushed outside of the PR commit range.

Now, someone replies to that. GitGitGadget sees the `issueCommentId` and tries [to call `createReplyForReviewComment()`](https://github.com/gitgitgadget/gitgitgadget/blob/7de60a2dba77d4b0c138ee375da68a655090b4d5/lib/github-glue.ts#L237), using the recorded comment ID. But that ID does not refer to a review comment (but to an issue comment instead), so the call will fail.

In short, I am now rather convinced that `issueCommentId` must not be set, and that `original Commit` must be explicitly unset in this code block.

webstech: Updated.  Wanted to get something out there to not lose comments.  I had not verified `addPRCommentReply` ended with  `createReplyForReviewComment`.  

Future changes can validate the commit is still part of the patch.

dscho (approved):
Thank you!

And sorry for the delay in re-reviewing. In my defense, I was battling an epic fight with the automation that dearly refused to build and deploy `curl` (MSYS/i686). But I won. After six hours or so.

webstech:
> No worries!

fyi, hours of research later, I think I accidentally did a commit amend in the vscode source control commit menu.  I've slept since then.  Hours of research had me learning about the .git/logs dir and finally learning all I really needed was the `reflog` command and a couple of minutes. 😳

PR #1195, opened 2022-12-30
git api: resolve promise in lineHandler

webstech:
The returned api promise will now only be resolved by the end of stdout processing for git command output handling.  The promise was being resolved early before the output handlers had run.

This was found during testing of an upcoming mail-archive-helper.test.ts contribution.  In successful processing everything is fine as promises are completed and notes are updated correctly.

The problem is that in the `git function`, a call to `GitProcess.exec` always resolves the promise.  If there is a callback handler, it may run after this and do a second (pointless) resolve on the promise.  For mail archive handling, this means the `git log` command completes before the log output has been processed and the updated `state.latestRevision` may be set before exceptions are thrown.  During log handling in the callback, it may set notes that are not yet available to the caller of the log processor due to the early promise resolve.

### DIFF
--- a/lib/git.ts
+++ b/lib/git.ts
@@ -107,8 +107,10 @@ export function git(args: string[], options?: IGitOptions | undefined):
                                     }':\nstderr: ${result.stderr
                                     }\nstdout: ${result.stdout}\n`);
             }
-            resolve(!options || options.trimTrailingNewline === false ?
-                    result.stdout : trimTrailingNewline(result.stdout));
+            if (!options?.lineHandler) { // let callback resolve the promise
+                resolve(!options || options.trimTrailingNewline === false ?
+                        result.stdout : trimTrailingNewline(result.stdout));
+            }
         }).catch((reason) => {
             reject(reason);
         });
--- a/lib/git.ts
+++ b/lib/git.ts
@@ -45,20 +45,10 @@ export function git(args: string[], options?: IGitOptions | undefined):
                 if (!process.stdout) {
                     throw new Error(`No stdout for "git ${args.join(" ")}`);
                 }
-                let linePromise: Promise<void> | undefined;
+                const promises: Array<Promise<void>> = [];
                 const handleLine = (line: string): boolean => {
                     try {
-                        if (!linePromise) {
-                            linePromise = lineHandler(line);
-                        } else {
-                            linePromise = linePromise.then(() => {
-                                return lineHandler(line);
-                            });
-                        }
-                        linePromise.catch((reason) => {
-                            reject(reason);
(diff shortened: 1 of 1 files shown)


dscho (approved):

dscho:
@webstech hopefully you're okay with me slipping in https://github.com/gitgitgadget/gitgitgadget/pull/1195/commits/49b15b872117e91c58a2dadf503313920d4ed4cd? I noticed this before leaving for the holidays but lacked the time to turn it into a full-blown PullRequest...

webstech:
> @webstech hopefully you're okay with me slipping in [49b15b8](https://github.com/gitgitgadget/gitgitgadget/commit/49b15b872117e91c58a2dadf503313920d4ed4cd)? I noticed this before leaving for the holidays but lacked the time to turn it into a full-blown PullRequest...

I am a little concerned about this.

1. In theory the promises all start processing 'at the same time'.  The line handler is building a buffer so it really needs the lines in sequence.  In reality, the lines are probably passed to the line handler in sequence so the mbox is built okay.  Will it always be this way?

1. Consider there are two emails, a reply to a patch email and a reply to that reply.  The first email starts getting processed and gets suspended while posting the commit comment (waiting for the response).  The second email starts getting processed and can't find `references` for the first commit comment because the notes have not been set for it (waiting to get the comment id back from GitHub).

So, even if the email mboxes are built correctly, they still need be processed in sequence.  Do you see how this can happen?  I'm going to put a wait into the new [tester](https://github.com/gitgitgadget/gitgitgadget/compare/main...webstech:gitgitgadget:mail-archive-helper-test) to verify.

webstech:
> I'm going to put a wait into the new [tester](https://github.com/gitgitgadget/gitgitgadget/compare/main...webstech:gitgitgadget:mail-archive-helper-test) to verify.

@dscho I added a wait at line 382 in the tester to simulate network latency and both emails got processed concurrently so there is a problem.

webstech:
For my future self, this is the code for the wait:

```
            console.log("waiting"); await new Promise<void>((resolve) => { setTimeout(() => { resolve(); }, 10000); });
````

webstech:
> I noticed this before leaving for the holidays

I am curious about what you observed.  Was there a failure but the new endpoint was updated?  Just trying to make sure the points of failure are covered.

dscho:
Hmm. I was pretty certain that the line handler I created for testing (which artificially slept for a while) failed to be `await`ed without the `Promise.all()` change...

I am not on a computer right now so I cannot test it (nor add an actual unit test for this) :-(

webstech:
PR #1199 opened to address the issues discussed above.

PR #1196, opened 2022-12-30
Support mostly lowercase `In-reply-to` header

webstech:
Some email handlers may only capitalize the first header character, making the relationships unknown.

Both `In-Reply-To` and  `In-reply-to` are now accepted.

If the mail server chooses to use some form of `In-Reply-To` instead of `References`, the relationship will be created.

This was observed in some messages on the [git-mailing-list-mirror](https://github.com/gitgitgadget/git-mailing-list-mirror/commits/lore-1/m).


dscho (approved):
Thank you!

dscho:
I hope you don't mind that I turned the regex into a case-insensitive one, @webstech ?

webstech:
> I hope you don't mind that I turned the regex into a case-insensitive one, @webstech ?

No problem.  I chose to be less permissive but since these are mostly software generated I am good with your change.

dscho:
IIRC the specs say that the header's keys are to be matched case-insensitively, so any parser (including GitGitGadget's) should do that. I might misremember, though.

PR #1198, opened 2022-12-31
Mail archive - track first patch line when sending patch email

webstech:
There are multiple commits in this series.

1.  Reformat for longer allowed line length
  Reduce the number of lines split across multiple lines.

2.  Add firstPatchLine to IMailMetadata
    Track the first patch line from patch email so it can be used on the commit comment.

1. Mail Archive: flag stale commit comments
   During mail archive handling, flag commit comments as outdated if the commit has been superseded in the patch.


dscho (commented):
Very nice and clean. Thank you!

dscho (approved):

PR #1199, opened 2023-01-02
CI: add tests for git serialization and sequencing

webstech:
Multiple commits in the patch series.

1. Ci: add a logger if --GGGVERBOSE is specified
   Utility to allow tracing to be left in test cases.  It would be displayed if --GGGVERBOSE is specified on the npm command.

    Usage:
    import logger from "./logger";
    logger("some  test related data");

2. CI: Add tests for git  line handlers
Three new tests to check serial execution, sequence of execution and slow git command processing.  The first two are similar but slightly different.  Serial checks to ensure other promises do not start before the current function has returned. Sequencing checks to ensure the output is processed in the proper order.
The slow processing tests verifies the command output is handled okay even when the command completes earlier.

3. git.ts: revert 49b15b8 (git linehandler issue)

    There are some issues with this patch.  Reverting to prior code pending another solution (a recent change appears to fix the problem).

Prior to the last commit to remove a previous commit (not sure of the best way to do this), [jobs](https://github.com/gitgitgadget/gitgitgadget/actions/runs/3820754267) were failing.

### DIFF
--- a/lib/git.ts
+++ b/lib/git.ts
@@ -45,10 +45,20 @@ export function git(args: string[], options?: IGitOptions | undefined):
                 if (!process.stdout) {
                     throw new Error(`No stdout for "git ${args.join(" ")}`);
                 }
-                const promises: Array<Promise<void>> = [];
+                let linePromise: Promise<void> | undefined;
                 const handleLine = (line: string): boolean => {
                     try {
-                        promises.push(lineHandler(line));
+                        if (!linePromise) {
+                            linePromise = lineHandler(line);
+                        } else {
+                            linePromise = linePromise.then(() => {
+                                return lineHandler(line);
+                            });
+                        }
+                        linePromise.catch((reason) => {
+                            reject(reason);
+                            process.kill();
+                        });
                     } catch (reason) {
                         reject(reason);
                         process.kill();
@@ -74,7 +84,12 @@ export function git(args: string[], options?: IGitOptions | undefined):
                     if (buffer.length > 0) {
                         handleLine(buffer);
                     }
-                    Promise.all(promises).then(() => resolve("")).catch(reject);
+                    if (linePromise) {
+                        linePromise.then(() => { resolve(""); })
+                            .catch((reason) => { reject(reason); });
+                    } else {
+                        resolve("");
+                    }
                 });
             };
         }
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "transform": {
       "\\.(ts|tsx)$": "ts-jest"
     },
-    "testRegex": "/tests/(?!test-lib).*\\.(ts|tsx|js)$"
+    "testRegex": "/tests/.*\\.test\\.(ts|tsx|js)$"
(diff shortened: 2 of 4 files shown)


dscho:
Thank you for working on this!

I do have a couple of concerns that I think could be addressed easily:

- In my experience, any sort of randomized behavior in tests ends up very hard to reproduce and hence causes more trouble than it is worth. Could you turn that into deterministic behavior?
- The `await ... setTimeout()` pattern is reasonably convoluted, so much so that I would prefer it to be abstracted into its own function (`const sleep = async (ms) => new Promise((resolve, reject) => setTimeout(ms, resolve))` or something similar).
- The `git` behavior we are concerned about is not really exposed by `git config`. I suspect that it would be better to simulate the patterns of concern explicitly, e.g. via `git -c alias.node="!node" node -e "(async () => { [... code ...] })().catch(console.log)"`, where `code` is carefully designed to produce output patterns like "print numbers 10 to 1, waiting the same number of milliseconds after each line".

@webstech what do you think?

webstech:
Thanks for the input.

> * In my experience, any sort of randomized behavior in tests ends up very hard to reproduce and hence causes more trouble than it is worth. Could you turn that into deterministic behavior?

I was not happy with it either.  I had it fixed but did not like everything running in sequence.  Changed it to now have odd numbers wait longer.  This is a consistent test.
> * The `await ... setTimeout()` pattern is reasonably convoluted, so much so that I would prefer it to be abstracted into its own function (`const sleep = async (ms) => new Promise((resolve, reject) => setTimeout(ms, resolve))` or something similar).

Hopefully the code is similar to what you were thinking.

> * The `git` behavior we are concerned about is not really exposed by `git config`. I suspect that it would be better to simulate the patterns of concern explicitly, e.g. via `git -c alias.node="!node" node -e "(async () => { [... code ...] })().catch(console.log)"`, where `code` is carefully designed to produce output patterns like "print numbers 10 to 1, waiting the same number of milliseconds after each line".

The problem is exposed by `git config`.  Both of those tests fail without the last commit.  The github actions history shows this.  I do like your suggestion of a slow output though.  I added it as a separate test.  Sounds easy right?  Not really 😣.  The alias specified as suggested works fine from the command line.  Using the `git.ts` function does not work.  Documenting for history.  The alias spec of `!node` is escaped by dugite or git (I did not determine which) and fails.  Tracing showed `trace: alias expansion: node => ''\!'node'`, which does not work.

Using this works:
```
    process.env.GIT_CONFIG_COUNT = `1`;
    process.env.GIT_CONFIG_KEY_0 = `alias.node`;
    process.env.GIT_CONFIG_VALUE_0 = `!node`;
```
It just takes time to figure it out.  `GIT_TRACE` was used to help debug this.  Sorry if I missed something obvious that would have made it work with `-c` on the command.

dscho (approved):
Thank you for going the extra mile to make the "slow stdout" test case work. It _really_ builds a lot more confidence in the code. ❤️

PR #1214, opened 2023-01-10
ci: new tests for MailArchiveGitHelper

webstech:
New CI tests to verify the archive handling works.

1. test not a pr related email
2. test already seen
3. test reply to cover letter
4. test reply to patch letter (including a reply to the reply)
5. test reply to outdated patch letter
6. test reply to not outdated patch letter


dscho (approved):
Wow, this is excellent (and was probably a ton of work). Thank you so much!

PR #1244, opened 2023-02-14
ci-helper tests: only show the `console.log()` output upon failure

dscho:
This is much quieter and easier to parse.

### DIFF
--- a/tests/ci-helper.test.ts
+++ b/tests/ci-helper.test.ts
@@ -9,6 +9,26 @@ import { testCreateRepo, TestRepo } from "./test-lib";
 
 jest.setTimeout(180000);
 
+// Only show the `console.log()` output when a test failed
+declare type AsyncFn = () => Promise<void>;
+function testQ(label: string, fn: AsyncFn) {
+    test(label, async () => {
+        const originalConsoleLog = console.log;
+        const mockLog = jest.fn();
+        try {
+            console.log = mockLog;
+            await fn();
+        } catch (e) {
+            mockLog.mock.calls.forEach(call => {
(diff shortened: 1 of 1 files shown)


webstech:
I agree the output is excessive and have wanted to but not done anything about it.  I don't disagree with this change but would it be better to use a jest reporter to do it for everything?

I gave [jest-silent-reporter](https://github.com/rickhanlonii/jest-silent-reporter) a try and it was okay but it only prints the error.  OTOH, if you run the specific test, all the output is displayed so maybe that is acceptable.  Another alternative, JestReporterLogValidator  could be a generic model to use switches to control if output is printed regardless of test completion status.

I ran with this in the `package.json`.  It does not give the detail info on each test suite completion.
```
    "reporters": [
        "summary",
        [
            "jest-silent-reporter", { "showWarnings": true, "useDots": true }
        ]
    ]
```

dscho:
Hmm. I actually like having the log shown explicitly. That opens the door for showing even more information _in the failure case only_, i.e. it equips us with more control.

webstech:
> Hmm. I actually like having the log shown explicitly. That opens the door for showing even more information _in the failure case only_, i.e. it equips us with more control.

I'm not sure what you are saying in the first sentence.  This patch suppresses the log unless there is a failure so the log is not shown.  Sorry for my confusion.

Using a reporter will cover all tests (ie not just ci-helper).  It appears the output already goes through the default reporter so using a different reporter is not adding a layer.

webstech (approved):
The default reporter does not provide this and alternates are not providing what the default does provide, so lets go with this.

PR #1391, opened 2023-08-17
Finalize the move of the GitHub App

dscho:
I received communication from Azure that the Azure Function we're using to back GitGitGadget's GitHub App still used node.js 16.x, but that support for that node.js version would end soon.

Naturally, I wanted to make sure that nothing would break when upgrading to 18.x, so I ran the test suite with that node.js version.

At least that's what I wanted to do, but I failed miserably because _there is no test suite_!

So I used the opportunity to implement one. To do so, I wanted to imitate how I did things in [Git for Windows' helper app](https://github.com/git-for-windows/gfw-helper-github-app), and naturally I wanted to have the same convenient deployment mechanism as that helper app has (just merge to `main` and a GitHub workflow will do the rest), so I separated out the Azure Function code into its own repository (https://github.com/gitgitgadget/gitgitgadget-github-app), implemented a test suite, added the GitHub workflows to run PR builds and the deployment, and then verified that things still work with node.js 18.x, which they did. I then registered a new Azure Function (because the old one was a Windows one, and I'd like to use a Linux-based one because they're quicker) and pushed to deploy. After that, I redirected the GitHub App's webhook URL[^1] to the new location and [performed a test](https://github.com/gitgitgadget/git/pull/1420#issuecomment-1682033321) to verify that GitGitGadget still does what it's supposed to do.

So now that all of that is said and done, it is time to remove the stale code from this here repository, and to add links to the new location of the GitHub App code.

[^1]: Actually, GitGitGadget has _two_ GitHub Apps, for technical reasons: the Git project wanted to prevent write access to the `git/git` repository, and [the `gitgitgadget-git` GitHub App](https://github.com/apps/gitgitgadget-git)  was created for that very purpose. I adjusted the webhook URLs of both the `gitgitgadget-git` and the `gitgitgadget` GitHub App.


webstech (approved):
There may be a run-on sentence in CONTRIBUTING.md but no big deal.

PR #1473, opened 2023-11-08
Add support for distributed notes updates

dscho:
One of the biggest blockers preventing GitGitGadget from moving away from Azure Pipelines toward GitHub workflows is the complication that the `refs/notes/gitgitgadget` note holds global state and must therefore be updated sequentially.

In the Azure Pipelines approach, we manage that by having an agent pool of one, which naturally can only serve one request at a time. That way, every job can first fetch the GitGitGadget notes ref, do whatever it needs to do, leisurely update the notes ref to reflect the new state and push the ref. Then, the next job can do the same and there is never a situation where a job tries to push the notes ref and fails because another, concurrent job has updated said ref in the meantime.

For various reasons, adding a pool of one to GitHub Actions is impractical.

In this PR, I propose a solution that allows for jobs to run concurrently, fetching the GitGitGadget notes, doing their thing, updating the notes ref along the way, and once it is time to push the notes ref and a situation is detected where it does not fast-forward, combine the changes in a way that is inspired by [Conflict-free replicated data types (CRTD)](https://en.wikipedia.org/wiki/Conflict-free_replicated_data_type), so that it can be pushed without problems after all.

The strategy implemented in this PR is to prefer the local changes in case of a hard conflict.

This means we will still have to be careful to avoid the same GitHub workflow to run concurrently: for example, we really will only ever want at most a single instance of the job to run whose responsibility it is to read new mails from the Git mailing list and mirror the relevant ones into the corresponding PRs. Should two jobs of that workflow run at the same time, they would end up adding PR comments _twice_, which is undesirable.

However, this PR unlocks the idea to let the workflow that mirrors the Git mailing list to the PRs run concurrently with multiple workflows that submit PRs as patches to the Git mailing list.

Note that this PR only adds that functionality, but it is in no way used yet. This will be the purpose of a follow-up PR.

### DIFF
--- a/lib/git-notes.ts
+++ b/lib/git-notes.ts
@@ -1,6 +1,22 @@
 import { emptyBlobName, git, revParse } from "./git";
 import { fromJSON, toJSON } from "./json-util";
 
+export type POJO = { [name: string]: string | string[] | number | number[] | boolean | POJO };
+
+/*
+ * Represents a temporary Git index that reflects a note tip commit, ready
+ * for making changes without committing them immediately.
+ *
+ * The purpose of this data structure is to support the `GitNotes.notesSync()`
+ * method.
+ */
+type TemporaryNoteIndex = {
+    appendNote: (oid: string, text: string) => Promise<void>,
+    setTextNote: (oid: string, text: string) => Promise<void>,
+    mutateObject: (oid: string, fn: (o: POJO) => void) => Promise<void>,
+    writeTree: () => Promise<string>
+};
+
 export class GitNotes {
     public static readonly defaultNotesRef = "refs/notes/gitgitgadget";
     public readonly workDir?: string;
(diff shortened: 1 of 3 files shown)


lib/git-notes.ts
```diff
@@ -1,6 +1,22 @@
 import { emptyBlobName, git, revParse } from "./git";
 import { fromJSON, toJSON } from "./json-util";
 
+export type POJO = { [name: string]: string | string[] | number | number[] | boolean | POJO };
+
+/*
+ * Represents a temporary Git index that reflects a note tip commit, ready
```
dscho, line 7:
```suggestion
 * Represents a temporary Git index that reflects the revision at the tip of the `refs/notes/gitgitgadget` branch, ready
```

dscho (commented):
Needs some grammar fix and clearer code comment.

dscho:
@webstech could you please have a look?

tests/git-notes.test.ts
```diff
@@ -53,15 +53,20 @@ test("set/get notes", async () => {
     expect(await notes.getLastCommitNote(commit as string)).toEqual("2");
 
     // error tests for update
-    await expect(notes.update(gitURL)).resolves.toBeUndefined();
+    for (const note of ["commit-to-mail", "mail-to-commit"]) {
```
webstech, line 56: Wasn't the network already being avoided by using `fakeRemote` instead of `gitURL` in PR #1786?

tests/ci-helper.test.ts
```diff
@@ -304,6 +311,7 @@ testQ("handle comment allow fail invalid user", async () => {
     };
 
     ci.setGHGetPRComment(comment);
+    ci.letGHGetGitHubUserInfoThrow("is not a valid GitHub username");
```
webstech, line 314: Would any error string work here?  ie, does a specific string need to be specified for the throw?  The error message is set by CIHelper.  This suggests the error is returned (in the throw) by Octokit.

Sorry I can't test this at the moment.  Windows 24H2 may have given me bad drivers and am slowing rebuilding in between system freezes.

lib/git.ts
```diff
@@ -21,6 +21,9 @@ function trimTrailingNewline(str: string): string {
 }
 
 export function git(args: string[], options?: IGitOptions): Promise<string> {
+    // allow the command to run in a bare repository
+    if (options?.workDir?.endsWith(".git")) args = [`--git-dir=${options.workDir}`, ...args];
```
webstech, line 25: Noting the `.git` suffix is not a git requirement but just a convention and maybe a best practice.  This would make it a requirement for users of ggg.  Not a big deal but would need to be documented.

Sorry if I missed doc that indicated it was a git requirement.

lib/patch-series.ts
```diff
@@ -819,7 +819,7 @@ export class PatchSeries {
 
         if (publishTagsAndNotesToRemote) {
             logger.log("Publishing tag");
-            await git(["push", publishTagsAndNotesToRemote, this.notes.notesRef, `refs/tags/${tagName}`], {
+            await git(["push", publishTagsAndNotesToRemote, `refs/tags/${tagName}`], {
```
webstech, line 822: I was wondering if this bit of code should be under the `dryRun` check just above it but it appears the only user of `dryRun` was removed in PR #1520.  Maybe `options.dryRun` is not needed any more?

webstech:
Still reviewing (re-reviewing really).

The `init-gitgitgadget-options` command should initialize a fully initialized options object.  This would remove some of the `optionsUpdated` settings for a new repo.  Requiring a repo to be initialized for supporting ggg should be reasonable now.  This is unlikely to cause a problem for current use.

The `handle-pr` command would be cleaner.  It is currently missing a push call if `optionsUpdated` is set.

dscho:
> Still reviewing (re-reviewing really).

Thank you!

> The `init-gitgitgadget-options` command should initialize a fully initialized options object.

I would like to have a nicer user experience, though, especially in the (hopefully not so far) future when I can use GitGitGadget also to contribute to Cygwin (i.e. run this Action in a fork of `gitgitgadget-workflows` with everything configured in a repository variable and some repository secrets after registering a new GitHub App via [`register-github-app-cli`](https://github.com/gr2m/register-github-app-cli)).

That means that I'd like to reduce the number of steps anyone has to go through to set up their own "GitGitGadget instance", which means: no more `init-gitgitgadget-options` command.

lib/ci-helper.ts
```diff
@@ -274,6 +277,7 @@ export class CIHelper {
 
         if (optionsUpdated) {
             await this.notes.set("", options, true);
+            await this.notes.push(this.urlRepo);
```
webstech, line 280: `CIHelper::handlePR`  is doing a push.  Is any of this necessary now?

lib/ci-helper.ts
```diff
@@ -420,6 +424,7 @@ export class CIHelper {
 
         if (optionsUpdated && updateOptionsInRef) {
             await this.notes.set("", options, true);
+            await this.notes.push(this.urlRepo);
```
webstech, line 427: Is `updateOptionsInRef` still useful?  It seems outdated and would allow the update if needed which would simplify upstream callers.  `optionsUpdated` does not need to return true if the notes are actually updated but that only matters if callers use it.

lib/git.ts
dscho: > Noting the `.git` suffix is not a git requirement but just a convention and maybe a best practice.

It is also what `git clone --bare <url>` appends to the newly-created directory's name.

> This would make it a requirement for users of ggg.  Not a big deal but would need to be documented.

Yup, I'll document it.

> Sorry if I missed doc that indicated it was a git requirement.

I should also have been clearer that the intention here is to prepare for [`safe.bareRepository = explicit`](https://git-scm.com/docs/git-config#Documentation/git-config.txt-safebareRepository) (which should have been made the default a long time ago but the Git project drags feet about it).

webstech:
> That means that I'd like to reduce the number of steps anyone has to go through to set up their own "GitGitGadget instance", which means: no more `init-gitgitgadget-options` command.

Maybe the options get needs to be able to push updates if it does not have a full object and callers can assume they have a complete object.  Having checks for missing fields is noise and complicates code.  Some code is deciding if the options is updated and then calling code that also has to decide if the options is updated.  Now that code (ie handlePR) is pushing the update but the caller is also thinking it may need to do an update.

dscho:
> The primary owner would need to be in the config to allow the init command to be removed.

That's probably the cleanest solution.

A more convenient solution would be to obtain the "collaborator permission" (similar to what [Git for Windows' GitHub App does](https://github.com/git-for-windows/gfw-helper-github-app/blob/main/GitForWindowsHelper/get-collaborator-permissions.js)), so that automatically all accounts with write permission would be added.

Instead of resorting to GraphQL, this could be done via [the `collaborators` endpoint](https://docs.github.com/en/rest/collaborators/collaborators?apiVersion=2022-11-28). Using GitHub CLI, it would look something like

```console
$ gh api repos/gitgitgadget/git/collaborators --jq \
  '.[] | select(.permissions.admin or .permissions.maintain or .permissions.push) | .login'
```

dscho:
> ```
> $ gh api repos/gitgitgadget/git/collaborators --jq \
>   '.[] | select(.permissions.admin or .permissions.maintain or .permissions.push) | .login'
> ```

I guess the following would be enough:

```console
$ gh api repos/gitgitgadget/git/collaborators --jq  '.[] | select(.permissions.push) | .login'
```

tests/git-notes.test.ts
dscho: You're correct! I guess I forgot about that when I rebased the branch. Will drop https://github.com/gitgitgadget/gitgitgadget/pull/1473/commits/615307bc55eab859e832aca55ec257fd7aebdd86 from this here PR.

tests/ci-helper.test.ts
dscho: Yes, any error string would work here! Will replace with "SOME ERROR" ;-)

lib/git.ts
dscho: > > This would make it a requirement for users of ggg.  Not a big deal but would need to be documented.
> 
> Yup, I'll document it.

Actually, this only affects GitGitGadget's _own_ use. So I think we'll be fine.

lib/patch-series.ts
dscho: I am not sure that I follow... I mean, I do agree that the `dryRun` option was only useful for the command-line interface when running this locally (originally as [the `mail-patch-series.sh` shell script](https://github.com/dscho/mail-patch-series)).

However, even expanding the context prior to this hunk, I see:

```typescript
        if (!this.options.dryRun) {
            const key = this.metadata.pullRequestURL || this.project.branchName;
            await this.notes.set(key, this.metadata, true);
        }
```

As such, I think you're absolutely correct that this should be guarded by that option, which I will do.

lib/ci-helper.ts
dscho: It is totally outdated, thank you for pointing that out! I will drop it in a new commit.

lib/ci-helper.ts
dscho: Good point! I'll drop this call.

tests/ci-helper.test.ts
webstech: > Windows 24H2 may have given me bad drivers and am slowly rebuilding in between system freezes.

fyi, the freezing appears to have been caused by the power management driver (shutting down the wrong cpu?).  There is a registry setting to turn it off.

lib/git.ts
webstech: > Actually, this only affects GitGitGadget's _own_ use. So I think we'll be fine.

Wouldn't this affect anyone trying to use a bare repo?

lib/patch-series.ts
webstech: > this should be guarded by that option

Well, in the end I was really wondering if `options.dryRun` support can be removed.  Not sure if it is checked everywhere it should be (ie should `publishBranch` check it?).

That could be a separate pull request.

webstech (commented):
I have finished reviewing the other commits with no additional comments.  Thanks for the education.

I will re-review when your changes arrive.

lib/patch-series.ts
dscho: > I was really wondering if `options.dryRun` support can be removed.

Me, too, and I am convinced that the answer is: Yes!

> That could be a separate pull request.

Totally!

lib/git.ts
webstech: > Wouldn't this affect anyone trying to use a bare repo?

By anyone I meant anyone using GitGitGadget to submit changes for repos other than git (by providing their own config).  Wouldn't these users also want to use bare repos in their workflows?

lib/patch-series.ts
```diff
@@ -812,15 +812,13 @@ export class PatchSeries {
             await this.notes.set("", globalOptions, true);
         }
 
-        logger.log("Publishing branch and tag");
-        await this.publishBranch(tagName);
```
webstech, line 816: I did not look at what cleanup could occur but there are quite a few branches and tags on the git repo.  Are there more than is currently needed?

lib/git.ts
dscho: > By anyone I meant anyone using GitGitGadget to submit changes for repos other than git (by providing their own config). Wouldn't these users also want to use bare repos in their workflows?

The idea is that the Azure Pipelines (and in the future, GitHub workflows) will initialize their local worktree (or bare repository) themselves, i.e. they have full control over the naming.

In any case, in the future, I think, we should _only_ use bare repositories, read: eventually this logic will go away and the callers will have to specify the `<worktree>/.git` path _iff_ they still want to work on non-bare rather than bare repositories.

lib/patch-series.ts
dscho: The code under discussion is a left-over from the CLI times, when the branch had not necessarily been published yet. The log message still talks about that branch.

However, what _does_ need to still be pushed, even in these times, is the tagged version that reflects the iteration that has been sent to the mailing list.

That function is already provided by [the conditional block `if (publishTagsAndNotesToRemote) { ... }`](https://github.com/gitgitgadget/gitgitgadget/pull/1473/files/13d86b95a6aac538b257d095c48e1387af60546e..90cf2b7ed6690553f6c33a08b038f6589864b322#diff-41cf3426d7f5952d827b30af22098f372998b86ab807647e443a376fd20b41ebL823-R824) below, though, and therefore the `publishBranch()` logic is no longer needed.

The branches that are in `gitgitgadget/git` are actually synchronized by [the `sync-*` workflows](https://github.com/gitgitgadget/gitgitgadget-workflows/actions) from some upstream sources, and serve only as potential PR target branches.

The vast majority of the tags in `gitgitgadget/git` are the ones reflecting submitted patch series iterations, the remainder have been mirrored by the `sync-*` workflows together with the branches.

lib/ci-helper.ts
```diff
@@ -257,24 +260,15 @@ export class CIHelper {
...
-        let optionsUpdated = false;
         for (const pullRequestURL in options.openPRs) {
             if (!Object.prototype.hasOwnProperty.call(options.openPRs, pullRequestURL)) {
                 continue;
             }
             console.log(`Handling ${pullRequestURL}`);
-            const [notesUpdated, optionsUpdated2] = await this.handlePR(pullRequestURL, options);
-            if (notesUpdated) {
+            const [notesUpdated, optionsUpdated] = await this.handlePR(pullRequestURL, options);
+            if (notesUpdated || optionsUpdated) {
```
webstech, line 269: The jsdoc return description no longer needs to specify `(and needs to be pushed)`.

tests/ci-helper.test.ts
```diff
@@ -304,9 +311,10 @@ testQ("handle comment allow fail invalid user", async () => {
     };
 
     ci.setGHGetPRComment(comment);
+    ci.letGHGetGitHubUserInfoThrow("SOME ERROR");
 
-    await expect(ci.handleComment("gitgitgadget", 433865360)).rejects.toThrow(/is not a valid GitHub username/);
-    expect(ci.addPRCommentCalls[0][1]).toMatch(/is not a valid GitHub username/);
+    await expect(ci.handleComment("gitgitgadget", 433865360)).rejects.toThrow(/SOME ERROR/);
```
webstech, line 316: The purpose of the test is to verify the bad user is reported by CIHelper.  Verifying the network access was bypassed seems like a different test.  In the CI test, github throws a bad request type error that CIHelper converts to the bad user with the GitHubGlue error message as part of the error.  Running locally, the test actually gets  a not authenticated error thrown by GitHubGlue (so not quite the test I thought it was).  Setting 
`getConfig().repo.owner = "gooduser";` fixes the auth error if the config is properly set up.  I understand what the update is proofing but don't think it is part of the test - the checks do not need to be changed.

dscho: Wait, you mean to say that the patch avoiding network access to GitHub REST should be dropped?

webstech:
@dscho Thanks for the updates.  Sorry but a couple of minor things.  fyi, due to the change overlap, I reviewed the complete change instead of individual commits - too much to track when comparing with local source.  I also assumed large commits I had not commented on were unchanged.

lib/ci-helper.ts
dscho: Fixed and pushed.

dscho:
@webstech thank you for the review!

I do not quite understand what you'd like me to do with c61a8531f4d6640aad05f8b6869aa3a14a5cbf1a; I'd like to get this over the finish line so that I can start playing with https://github.com/gitgitgadget/gitgitgadget/pull/1392 on top of this ;-)

tests/ci-helper.test.ts
webstech: Running locally, there is no network access due to no authentication (required to ask about users).  Running on workflows or pipelines, there is network access but is that a problem?  I'll let you decide on that.

I think, either way, the tests should still be checking for `is not a valid GitHub username` and not `SOME ERROR` since that is the purpose of the test.  CIHelper is setting the not valid message.  Sorry I was not clear.

dscho:
> I do not quite understand what you'd like me to do with [c61a853](https://github.com/gitgitgadget/gitgitgadget/commit/c61a8531f4d6640aad05f8b6869aa3a14a5cbf1a); I'd like to get this over the finish line so that I can start playing with #1392 on top of this ;-)

Okay, I still don't know, but I really want to merge this PR, so I simply dropped that patch from the PR and will merge as soon as CI finishes successfully.

PR #1512, opened 2023-12-18
build(deps): fix bump marked from 10.0.0 to 11.1.0

webstech:
Cast `marked.parse()` return value as string.

Fix for failing #1507.


webstech:
When I initially did this two weeks ago, an override was needed and that seems to have gone away.
```
    // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
    return htmlToText(marked.parse(markdown) as string, formatOptions);
```

dscho (approved):

PR #1520, opened 2024-01-08
Some cleanups

dscho:
Various cleanups just to reduce technical debt.


lib/project-options.ts
```diff
@@ -37,13 +18,12 @@ export class ProjectOptions {
             for (const user of project.cc) {
                 cc.push(user);
             }
-        } else if (await commitExists("cb07fc2a29c86d1bc11", workDir) &&
-            await revParse(`${baseCommit}:git-gui.sh`, workDir) !== undefined) {
+        } else if (await revParse(`${baseCommit}:git-gui.sh`, workDir) !== undefined) {
```
webstech, line 21: Most of these checks go away with unique configs for cygwin and busybox.  The git-gui check has stuck in my head but maybe an optional config property to identify a file to check for and overriding emails could work.  If another project had similar issues, a function may need to be defined in the config.  Getting way more complex.

webstech (approved):
Thanks for cleaning up.

lib/project-options.ts
dscho: It should not be too hard to add some (optional) array of hashes to the config where the key is a path whose presence will override the specified values. Something like this:

```json
{
  "project": {
    "repo": {
      [...]
    },
    [...],
    "override-if-file-is-present": {
      "git-gui.sh": {
        "repo": {
          [...]
        }
      }
    }
  }
}
```

PR #1540, opened 2024-02-04
WELCOME: add a note about single-commit PRs

phil-blain:
Contributors continue to send single-commit PRs to the mailing list with identical text in their commit message and below the three-dash line.

This is another simple attempt to try to prevent that.

In https://github.com/gitgitgadget/gitgitgadget.github.io/pull/8#issuecomment-1133758832 Dscho suggested adding such a check in Gitgitgadget itself, which I think is a good idea, but I lack the time to pursue it currently.

And, the current change to the welcome message does not make it _that much_ longer, which I think is a good compromise.

### DIFF
--- a/res/WELCOME.md
+++ b/res/WELCOME.md
@@ -2,7 +2,9 @@
 
 Hi @${username}, and welcome to GitGitGadget, the GitHub App to send patch series to the Git mailing list from GitHub Pull Requests.
 
-Please make sure that your Pull Request has a good description, as it will be used as cover letter. You can CC potential reviewers by adding a footer to the PR description with the following syntax:
+Please make sure that your Pull Request has a good description, as it will be used as cover letter.
+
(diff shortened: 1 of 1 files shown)


phil-blain:
> [phil-blain](https://github.com/phil-blain) wants to merge 3 commits into [gitgitgadget:main](https://github.com/gitgitgadget/gitgitgadget/tree/main) from [phil-blain:main](https://github.com/phil-blain/gitgitgadget/tree/main)

I just realized I forgot to create a topic branch... I can open a new PR (so that the merge commit has a better message) if needed (since I can't change the PR's branch).

dscho (approved):

dscho:
Thank you @phil-blain! It took a bit longer to review and merge this than I wanted because I had to find time after-hours to do so (unfortunately I am no longer funded in my day-job at GitHub to work on GitGitGadget).

PR #1644, opened 2024-06-30
WELCOME.md: warn new users against copy/pasting CC lists

listx:
Copy/pasting Markdown-formatted CC lists from one GGG PR to another can result in a broken CC list which needs manual fixing from the maintainer. So warn users about this issue early on for all first-time contributors.

See [1] for a discussion of some of the ways the community would like to mitigate this issue (one of which is the motivation for this change).

[1] https://lore.kernel.org/git/xmqqle2o26jy.fsf@gitster.g/

### DIFF
--- a/res/WELCOME.md
+++ b/res/WELCOME.md
@@ -11,6 +11,10 @@ You can CC potential reviewers by adding a footer to the PR description with the
 
     CC: Revi Ewer <revi.ewer@example.com>, Ill Takalook <ill.takalook@example.net>
 
+NOTE: DO NOT copy/paste your CC list from a previous GGG PR's description,
+because it will result in a malformed CC list on the mailing list. See
+[example](https://lore.kernel.org/git/owly4jd741ph.fsf@fine.c.googlers.com/).
+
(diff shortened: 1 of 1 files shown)


dscho (approved):
An even better solution would be to detect the faulty Cc: lines and refuse to send the mail(s), with a helpful error message to guide the user out of the situation.

Having said that, I'll gladly take this PR because it helps _right now_.

listx:
> An even better solution would be to detect the faulty Cc: lines and refuse to send the mail(s), with a helpful error message to guide the user out of the situation.

Fully agree. Any takers?

> Having said that, I'll gladly take this PR because it helps _right now_.

Thanks.

dscho:
> > An even better solution would be to detect the faulty Cc: lines and refuse to send the mail(s), with a helpful error message to guide the user out of the situation.
> 
> Fully agree. Any takers?

I always love to work on such things, but cannot afford any nerdsnipes at the moment.

Having said that, I will add a few pointers to https://github.com/gitgitgadget/gitgitgadget/issues/1645 to help any volunteer along.

PR #1671, opened 2024-08-05
ES Modules Support

webstech:
This patch series implements changes to use ES modules  in place of the default CommonJS.  This is needed to support octokit  package changes converting to ES modules.

The series covers several steps:

1. Update the config (break everything).
2. Update imports to specify qualified names for local files.
3. Update imports for non-exported nodemailer functions.
4. Update imports in test files to specify qualified names for local files.
5. Remove use of __dirname and __filename.
6. Update the config to enable jest to work with ES modules.  The qualified names are a problem.
7. Update the workflow old branch cleanup to work with modules.
8. Update the azure pipeline to use a more current node.

Builds and runs with the current package versions.  May need an update to actually move to the latest octokit packages.

It is unfortunate that Jest support of modules does not appear to be moving forward.


webstech:
@dscho Sorry for the noise.  I should have got the workflows running before opening the PR.

These patches are my understanding of what is needed for the ES module conversion.  Sorry if I got something wrong - the doc can be confusing with the multiple options available.

webstech:
@dscho This should be ready to go.  It has been updated with today's changes.

dscho (approved):
Wow, what a lot of work this required! Thank you for doing all of this.

PR #1688, opened 2024-08-20
eslint: new version preparation

webstech:
Prep work for moving to updated eslint with new flat config format.  Initial conversion shows `prettier` does not appear to be running in the current setup.  This set of patches reduces the number of problems from over 500 to ~350.  It is mostly fixing formatting problems that are resolved using the extended line length now in use.  Prettier may be more opinionated than we would like.

The patches are mostly single file per patch in case any need to be removed.

lint: use Object.prototype.hasOwnProperty for safety
lint: remove superfluous escapes
lint: add parens around await as per linter
lint: cleanup ',' versus ';' usage
lint: misc linter issues
lint: misc linter issues
lint: clean up terminators
lint: misc linter issues
lint: misc clean up
lint: mostly format updates and a grep change
lint: clean up and `Error` type casting
lint: formatting issues
lint: line length and space issues
lint: undent a section of code

### DIFF
--- a/lib/git-notes.ts
+++ b/lib/git-notes.ts
@@ -32,11 +32,9 @@ export class GitNotes {
         return await this.setString(key, toJSON(value), force);
     }
 
-    public async setString(key: string, value: string, force?: boolean):
-        Promise<void> {
+    public async setString(key: string, value: string, force?: boolean): Promise<void> {
         const obj = await this.key2obj(key);
-        if (obj !== emptyBlobName &&
-            !await revParse(`${obj}^{blob}`, this.workDir)) {
+        if (obj !== emptyBlobName && !(await revParse(`${obj}^{blob}`, this.workDir))) {
             try {
                 /*
                  * Annotate the notes ref's tip itself, just to make sure that
@@ -61,8 +59,7 @@ export class GitNotes {
         }
     }
 
-    public async appendCommitNote(commit: string, note: string):
-        Promise<string> {
+    public async appendCommitNote(commit: string, note: string): Promise<string> {
         return await this.notes("append", "-m", note, commit);
     }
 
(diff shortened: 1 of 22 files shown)


lib/gitgitgadget.ts
```diff
@@ -144,7 +143,7 @@ export class GitGitGadget {
             throw new Error(`User ${vouchingUser} lacks permission for this.`);
         }
 
-        if (!this.isUserAllowed(user)) {
+        if (!this.allowedUsers.delete(user)) {
```
dscho, line 146: It did take me a couple of moments to wrap my head around why this is correct, but yes, it is!

lib/git-notes.ts
```diff
@@ -76,12 +73,8 @@ export class GitNotes {
     }
 
     public async update(url: string): Promise<void> {
-        if (this.notesRef === "refs/notes/gitgitgadget" ||
-            this.notesRef === "refs/notes/commit-to-mail" ||
-            this.notesRef === "refs/notes/mail-to-commit") {
-            await git(["fetch", url,
-                       `+${this.notesRef}:${this.notesRef}`],
-                      { workDir: this.workDir });
+        if (this.notesRef.match(/^refs\/notes\/(gitgitgadget|commit-to-mail|mail-to-commit)$/)) {
```
dscho, line 76: An alternative would be `["refs/notes/gitgitgadget", "refs/notes/commit-to-mail", "refs/notes/mail-to-commit"].includes(this.notesRef)`. But the regex is fine, too (despite [some concerns](https://blog.codinghorror.com/regular-expressions-now-you-have-two-problems/), this seems outside [dangerous territory](https://stackstatus.tumblr.com/post/147710624694/outage-postmortem-july-20-2016)).

dscho (approved):
Wow. Big cleanup, I guess it's still spring somewhere ;-) Cheek aside, thank you so much for doing this!

lib/git-notes.ts
webstech: I am adding [eslint-plugin-security](https://github.com/eslint-community/eslint-plugin-security#readme) to replace the trojan attack detection.  It also includes unsafe regex checking.

PR #1690, opened 2024-08-22
eslint: more lint updates and conversion to new eslint flat config

webstech:
Changes to scripts and test code.  Lastly, convert to the latest eslint.

1. lint: update scripts for linting issues
2. lint: update tests for linting issues
3. maint: add missing object attributes
4. eslint: convert to new flat config format

General lint changes are to fix space formatting issues and  change a number of `expect()` to single liners. 

The last patch completes the conversion.  The config is based on using the migration tool as a basis.  The `prettier` recommended was not used so the warning could be set.  It was also hoped there could be some influence over prettier output.

- Convert from `.eslintrc` to `eslint.config` file format.  Remove the old config and ignore files.  The ignore was not needed anyway.
-  Replaced the trojan detection with `eslint-plugin-security`, which includes other checks.  Added `typescript-eslint` for ts checks.
- Added `@stylistic` for lints moved out of eslint.  Some of these were intended to override prettier but that is not working.
- Set `prettier` to warn.  It is too opinionated to be allowed to error.

While `prettier` is useful, `@stylistic` may give more flexibility.


webstech:
If there is a preference to migrate to `prettier` or `@stylistic` formatting I am open.  Removing `prettier` gets rid of most of the warnings.

dscho (approved):
I am so grateful that you go through this large effort!

PR #1758, opened 2024-11-12
test: handle some concurrency issues

webstech:
Some tests have recently failed due to test jobs running at the same time.  Directories or git global configs unexpectedly existed.

There may still be an edge case setting the global config.  This will be addressed when it shows up.

A second patch changes to using node native promise based functions.


dscho (approved):
Looks good to me!

PR #1798, opened 2025-01-10
Fix parsing of long header bodies

### LINKED ISSUE #1794: Fix parsing of long message IDs
I just noticed that https://lore.kernel.org/git/TYCPR01MB1173678128DCC128099726865AB162@TYCPR01MB11736.jpnprd01.prod.outlook.com/raw contains a new-line in the `Message-ID` header, which [GitGitGadget seems to be unable to cope with](https://dev.azure.com/gitgitgadget/git/_build/results?buildId=232613&view=logs&j=275f1d19-1bd8-5591-b06b-07d489ea915a&t=704ea7c9-8eb0-53b9-6b30-87d5a7befe2c&l=15).

This should be fixed, as I do not want Outlook users to be punished even more by the Git project (for years, you could not send emails to the Git mailing list if you had a `@outlook.com` email address).

H1ghBre4k3r:
Improve handling of line breaks in folded headers by removing leading new lines of field bodies in folded headers, since they should have no semantical nor syntactical meaning, according to the specification. 

This general handling also provides proper soundness for other fields which might get folded. 

Fixes #1794

### DIFF
--- a/lib/send-mail.ts
+++ b/lib/send-mail.ts
@@ -54,7 +54,7 @@ export async function parseMBox(mbox: string, gentle?: boolean): Promise<IParsed
 
     for (const entry of parsed.headerLines) {
         // try to parse header line and consume a leading line break after the colon in folded headers
-        const valueSet = entry.line.match(/(.*?):(?:\r?\n)? *([^]*)$/);
+        const valueSet = entry.line.match(/(.*?):(?:\r\n)? *([^]*)$/);
         if (!valueSet) {
             if (entry.line[entry.line.length - 1] === ":") {
                 continue;
--- a/lib/send-mail.ts
+++ b/lib/send-mail.ts
@@ -53,7 +53,8 @@ export async function parseMBox(mbox: string, gentle?: boolean): Promise<IParsed
     const parsed = await simpleParser(mbox, options);
 
     for (const entry of parsed.headerLines) {
-        const valueSet = entry.line.match(/(.*?): *([^]*)$/);
+        // try to parse header line and consume a leading line break after the colon in folded headers
+        const valueSet = entry.line.match(/(.*?):(?:\r?\n)? *([^]*)$/);
         if (!valueSet) {
             if (entry.line[entry.line.length - 1] === ":") {
                 continue;
--- a/tests/send-mail.test.ts
+++ b/tests/send-mail.test.ts
@@ -62,7 +62,7 @@ test("parse mbox", async () => {
         { key: "Content-Transfer-Encoding", value: "8bit" },
         { key: "MIME-Version", value: "1.0" },
         { key: "Header-with-no-value", value: "" },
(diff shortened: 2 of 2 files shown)


webstech:
Background:
The question of a multi-line Message-ID came up in PR #905.  It was left pending until seen in the wild and now we have seen it.  PR #889 changed the original parsing to use a generic parser that left the newlines in place.  Tests specifically added multi-line headers. 

The problematic handling of certain multi-line headers in PR #905 was done in `parseMBoxMessageIDAndReferences`.  Should this fix be there instead of here?  The current change removes the first newline but some headers may have multiple newlines.  Callers of `parseMBox` have always accepted newlines in the results (except for cc where they are removed).  Mail archive helper only uses the results of `parseMBox` to call `parseMBoxMessageIDAndReferences`.

It may be more consistent to remove all newlines in `parseMBox`  but need to ensure no callers are affected by long header lines.

lib/send-mail.ts
```diff
@@ -53,7 +53,8 @@ export async function parseMBox(mbox: string, gentle?: boolean): Promise<IParsed
     const parsed = await simpleParser(mbox, options);
 
     for (const entry of parsed.headerLines) {
-        const valueSet = entry.line.match(/(.*?): *([^]*)$/);
+        // try to parse header line and consume a leading line break after the colon in folded headers
+        const valueSet = entry.line.match(/(.*?):(?:\r?\n)? *([^]*)$/);
```
dscho: I think this will work.

While looking at this regex, I cannot fail to notice that it isn't high quality (probably my work, honestly): Header keys ["MUST be composed of printable US-ASCII characters (i.e., characters that have values between 33 and 126, inclusive), except colon"](https://datatracker.ietf.org/doc/html/rfc2822#section-2.2), and they must be at the start of the line. Therefore, I think we will want to consider using `/^([^ :]+):` instead of `/(.*?):`, but that's outside _this PR_'s scope.

lib/send-mail.ts
```diff
@@ -54,7 +54,7 @@ export async function parseMBox(mbox: string, gentle?: boolean): Promise<IParsed
 
     for (const entry of parsed.headerLines) {
         // try to parse header line and consume a leading line break after the colon in folded headers
-        const valueSet = entry.line.match(/(.*?):(?:\r?\n)? *([^]*)$/);
+        const valueSet = entry.line.match(/(.*?):(?:\r\n)? *([^]*)$/);
```
dscho, line 57: I do not quite understand why we cannot be lenient here, but I guess it does not matter much in practice 😉

dscho (approved):
I am sorry, this PR has languished a lot and it is all my fault. In my defense, I needed quite some time to recuperate from a quite stressful and unsuccessful security bug fix triple-sprint.

In any case, thank you so much for your contribution @H1ghBre4k3r!

PR #1852, opened 2025-03-14
Adjust the "reply to this" links

dscho:
Since the wiki was locked down for external contributors (to prevent spam and vandalism), it was not really a good place for the information. As of https://github.com/gitgitgadget/gitgitgadget.github.io/pull/21 it moved to GitGitGadget's home page.

Let's stop referring to the wiki for that reason, and refer to the new location instead.


webstech (approved):
Thank you.

PR #1972, opened 2025-08-15
Some fixes

dscho:
While working [towards a GitHub Action](https://github.com/gitgitgadget/gitgitgadget/pull/1392), I noticed a couple of issues that are really easy to resolve.


dscho:
@webstech I hope to be moving fast on these changes (as well as a couple of others, including one that will unbreak the Azure Pipelines after I merged #1473...). If you're available, I would appreciate timely reviews, but if this is an inopportune time for you, don't worry (I'll want to move relatively fast before my own availability dwindles again, hope you don't mind!).

webstech (approved):
Thanks for fixing up my style errors.  It was probably cut and pasted at the time but I should have corrected them.

Funny how the styles differ between mjs and json.

dscho:
@webstech thanks!

> Funny how the styles differ between mjs and json.

TBH I think this might just be a change in ESLint, I didn't manage to keep up with it ;-)

PR #1974, opened 2025-08-15
Re architecture token handling

dscho:
The primary motivation for this change is to move GitGitGadget towards a set of GitHub Actions that are used in GitHub workflows. The urgency comes from a secondary, originally unanticipated motivation: When I merged #1473, I broke the current Azure Pipelines (the fancy push-or-merge-and-push strategy does not work because the Azure Pipelines cannot push the Git notes to `gitgitgadget/git` due to missing authorization).

So here comes a set of changes to allow specifying the token(s) to use by `CIHelper` in a _much_ more intentional way. Previously, the idea was that the token to use was specified via the Git config options `gitgitgadget.githubToken` (and `gitgitgadget.git.githubToken` for `git/git`, where GitGitGadget requires less permissions; It does not need to push, just do things like adding PR comments). This still works, for backwards-compatibility (at least until I managed to move everything over to GitHub Actions/workflows), but now there is a much more explicit way to configure the tokens to use for specific GitHub rogs in the `CIHelper` object. This new method is now also called from `misc-helper`, which should fix the Azure Pipelines.

### DIFF
--- a/lib/ci-helper.ts
+++ b/lib/ci-helper.ts
@@ -576,7 +576,7 @@ export class CIHelper {
         };
 
         try {
-            const gitGitGadget = await GitGitGadget.get(this.gggConfigDir, this.workDir);
+            const gitGitGadget = await GitGitGadget.get(this.gggConfigDir, this.workDir, this.notesPushToken);
             if (!gitGitGadget.isUserAllowed(comment.author)) {
                 throw new Error(`User ${comment.author} is not yet permitted to use ${this.config.app.displayName}`);
             }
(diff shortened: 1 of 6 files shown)


dscho:
I'll push some updates shortly (to make this more comprehensive a business of configuring tokens explicitly).

dscho:
> I _think_ it will fix all those Azure Pipeline failures.

Seems to do the job: [re-run #1](https://dev.azure.com/gitgitgadget/git/_build/results?buildId=253297&view=logs&j=f4a2bb40-7e70-5f50-d570-7d7c5c8ca123&t=28db4af1-c17a-5ad6-7967-acadf04ad18b), [re-run #2](https://dev.azure.com/gitgitgadget/git/_build/results?buildId=253159&view=logs&j=92a4755a-6173-591e-5c07-8dbe1165e937&t=05dc4a6b-2706-5ea6-77c8-e14f00112193).

webstech:
@dscho Sorry I did not get a chance to comment on this.  I did review it and had no issues but was looking at the `publishRemote` aspect.  If the publishRemote env var/git config item no longer needs the auth token, isn't it just a value that can be generated from repo.host, repo.owner and repo.name in the config?

PR #1975, opened 2025-08-15
Tests: no network calls!

dscho:
As promised, this is my next attempt at removing network calls during tests (so that they would pass even when stuck in a plane without internet connection).

### DIFF
--- a/tests/ci-helper.test.ts
+++ b/tests/ci-helper.test.ts
@@ -105,6 +105,13 @@ class TestCIHelper extends CIHelper {
         this.ghGlue.getGitHubUserInfo = jest.fn(async (): Promise<IGitHubUser> => o);
     }
 
+    public letGHGetGitHubUserInfoThrow(err: string): void {
+        // eslint-disable-next-line @typescript-eslint/require-await
+        this.ghGlue.getGitHubUserInfo = jest.fn(async (): Promise<IGitHubUser> => {
+            throw new Error(err);
+        });
+    }
+
     public addMaxCommitsException(pullRequestURL: string): void {
         this.maxCommitsExceptions = [pullRequestURL];
     }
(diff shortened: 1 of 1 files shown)


webstech (commented):
In PR #1786 you removed the remote access in `git-notes.test.ts` by adding `insteadof` to the config.  Is there a reason to do this as well?  The tests run fine no matter what the value of `gitURL` is.

dscho:
> In PR #1786 you removed the remote access in `git-notes.test.ts` by adding `insteadof` to the config. Is there a reason to do this as well? The tests run fine no matter what the value of `gitURL` is.

Oh wow, you're right! No, it is not necessary. I simply forgot (apparently I fixed the same problem several times over the years... 🤦).

tests/ci-helper.test.ts
```diff
@@ -304,6 +311,7 @@ testQ("handle comment allow fail invalid user", async () => {
     };
 
     ci.setGHGetPRComment(comment);
+    ci.letGHGetGitHubUserInfoThrow("is not a valid GitHub username");
```
webstech, line 314: This test was originally intended to verify this line was run in `CIHelper`:
```
                    throw new Error(`User ${accountName} is not a valid GitHub username: ${reason}`);
```
Running the tests locally, the error is `Error: User bad_@@@@ is not a valid GitHub username: Error: Need a GitHub token for gitgitgadget`, which has as the reason an error thrown by `GitHubGlue` (and no network access).  If you want to verify there is no network while still allowing the original test, how about:
```
   const noNetwork = "No net";
   ci.letGHGetGitHubUserInfoThrow(noNetwork);
   await ...
   expect ...
   expect(ci.addPRCommentCalls[0][1]).toMatch(noNetwork); // new test for no network access
```

dscho: Oh wow, never mind! I must have been in a situation when I developed the patch originally where I had lost network access (being happy about Git's distributed nature!). However, I just reran the test suite after disconnecting from the network, and it all passed:

```
Test Suites: 26 passed, 26 total
Tests:       152 passed, 152 total
Snapshots:   0 total
Time:        115.744 s, estimated 116 s
Ran all test suites.
```

I should have verified whether this patch is still needed before opening this PR, but I [keep forgetting that when rebasing branch thickets](https://github.com/gitgitgadget/gitgitgadget/pull/1473#discussion_r2009052927), sorry!!!

Issue #609, opened 2021-06-22
Migrate away from Azure Pipelines to GitHub Actions


dscho:
A large part of GitGitGadget's operation runs inside Azure Pipelines, either in a scheduled run, or triggered directly via a `push` event or indirectly via the GitHub App (which is backed by an Azure Function that performs rudimentary validation and then triggers the Pipeline via a PAT, passing along some parameters).

Azure Pipelines seems to be developed less actively than GitHub Actions, therefore it is desirable to migrate from the former to the latter.

One challenge are the secrets used by the Azure Pipeline: we need to ensure that they stay, well, secret. It is much easier to accidentally leak them by carelessly accepting a seemingly innocuous PR that updates some random CI run (but really is designed to leak an unprotected secret). We might want to use this shiny new tool in the toolbox called ["environment secrets"](https://docs.github.com/en/actions/reference/environments#environment-secrets), where we can restrict in the GitHub UI which workflows are allowed to access them.

One of the very neat things GitGitGadget does is that it is shown in the "checks" of a PR, i.e. if GitGitGadget is currently preparing a patch series to send or to preview, the PR status is set to "in progress".

The way this is done is by registering the corresponding Azure Pipeline with the repository targeted by the PR. Currently, there are three such Pipelines:

- The [`GitGitGadget PR Handler` Pipeline](https://dev.azure.com/gitgitgadget/git/_build?definitionId=3) is associated with [`gitgitgadget/git`](https://github.com/gitgitgadget/git/pulls)
- The [`GitGitGadget PR Handler (git)` Pipeline](https://dev.azure.com/gitgitgadget/git/_build?definitionId=13) is associated with [`git/git`](https://github.com/git/git/pulls)
- (For testing only) The [`GitGitGadget PR Handler (dscho)` Pipeline](https://dev.azure.com/gitgitgadget/git/_build?definitionId=12) is associated with `dscho/git`.

Whenever GitGitGadget's GitHub App installed in the respective repository receives an event about an added PR comment (in the form of triggering [the Azure Function](https://github.com/gitgitgadget/gitgitgadget/tree/HEAD/azure-function/)), it manually triggers the respective Azure Pipeline.

Since the introduction of the `workflow_dispatch` event, it should be possible to port this strategy to GitHub workflows.

Except.

[...]

PR #1980, opened 2025-08-20
Prepare config handling for the upcoming GitHub Action

dscho:
This is part 1 of going for [the idea to switch from Azure Pipelines to GitHub workflows](https://github.com/gitgitgadget/gitgitgadget/issues/609).

In this PR, I tried to break out the commits into their own change set which are concerned with setting up GitGitGadget to do its work, including the configuration which repositories to clone and where to.

### DIFF
--- a/lib/ci-helper.ts
+++ b/lib/ci-helper.ts
@@ -582,7 +582,12 @@ export class CIHelper {
         };
 
         try {
-            const gitGitGadget = await GitGitGadget.get(this.gggConfigDir, this.workDir, this.notesPushToken);
+            const gitGitGadget = await GitGitGadget.get(
+                this.gggConfigDir,
+                this.workDir,
+                this.urlRepo,
+                this.notesPushToken,
+            );
             if (!gitGitGadget.isUserAllowed(comment.author)) {
                 throw new Error(`User ${comment.author} is not yet permitted to use ${this.config.app.displayName}`);
             }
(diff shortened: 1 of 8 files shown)


mjcheetham (approved):

lib/gitgitgadget.ts
```diff
@@ -168,6 +173,10 @@ export class GitGitGadget {
 
     // Send emails only to the user
     public async preview(pr: IPullRequestInfo, userInfo: IGitHubUser): Promise<IPatchSeriesMetadata | undefined> {
+        const smtpOptions = this.smtpOptions;
+        if (!smtpOptions) {
```
webstech, line 177: Could this check be moved to `sendMail()`?  Get it down to a single location.

lib/ci-helper.ts
```diff
@@ -5,13 +5,14 @@ import { ILintError, LintCommit } from "./commit-lint.js";
 import { commitExists, git, emptyTreeName } from "./git.js";
 import { GitNotes } from "./git-notes.js";
 import { GitGitGadget, IGitGitGadgetOptions } from "./gitgitgadget.js";
+import { getConfig } from "./gitgitgadget-config.js";
```
webstech, line 8: The original implementation only used this function from the outer layer (misc-helper, action handlers) to set the config singleton to the gitgitgadget instance.  At the time, there was an issue (node, js or ts?) so the singleton was set automatically during loading (this is mentioned in PR #887), so it should never be uninitialized.  It may mean this `getConfig()` never needs to be called.  I think the issue related to internal js testing of class constructing during load time.  Maybe the removal of the call on `config: IConfig = getConfig()` would get rid of the use before defined loading error.

Just trying to clarify why there are two functions and expected use.  The function used in the constructor was expected to get the singleton and not need to have it passed.  This is used in other classes as well.

PR #1981, opened 2025-08-20
More prerequisites to move GitGitGadget towards GitHub Actions

dscho:
After part 1 (https://github.com/gitgitgadget/gitgitgadget/pull/1980), this is part 2 of addressing #609. It is a stacked PR, with two more to follow in this here repository.

The themes of the patches included in this PR are:

- addressing various bugs in existing code paths that aren't yet exercised, but will be
- undoing the previous attempts at adding GitHub Actions on top of GitGitGadget, and
- lay the ground-work to add minimal pure-JavaScript GitHub Actions on top `CIHelper`.

PR #1982, opened 2025-08-20
Add a set of GitHub Actions to perform GitGitGadget's job

dscho:
This PR is part 3 of addressing #609, and it is stacked on top of #1980 and #1981 (and therefore contains also the commits of those PRs), therefore I will leave this in draft mode until those PRs are merged.

The grand idea is to bundle the `CIHelper` class together with all its direct and transitive dependencies into one big, honking `dist/index.js`, and then add a set of really minimal GitHub Actions that call into `CIHelper`. The Actions are added in sub-directories so that they can be called in GitHub workflows via e.g. `- uses: gitgitgadget/gitgitgadget/update-prs@1`.

The component used for bundling `CIHelper` is [`@vercel/ncc` ](https://www.npmjs.com/package/@vercel/ncc). To support acting as a GitHub Action, [`@actions/core`](https://www.npmjs.com/package/@actions/core) is installed.

To allow for really minimal GitHub Actions, the `CIHelper` class is augmented accordingly to re-implement more logic that is currently either in `misc-helper.ts` or in the (non-public 😞) Azure Pipelines definitions.

The naming convention for specifying the necessary tokens as GitHub Actions inputs is:
- `upstream-repo-token`: This is to comment on PRs in `git/git`
- `pr-repo-token`: This is to comment on PRs in `gitgitgadget/git` (as well as to be able to push to that repository)
- `test-repo-token`: This is to comment on PRs in `dscho/git` (used exclusively for testing)

To clarify, here is a diagram:

```mermaid
graph TD
    user["user (contributor)"]
    upstream-repo["upstream-repo (authoritative project repo)"]
    pr-repo["pr-repo (GitGitGadget-enabled GitHub repo)"]
    GitGitGadget["GitGitGadget"]
    mailing-list["mailing-list"]

    user -->|"opens PR"| pr-repo
    user -->|"opens PR (if GitHub App installed)"| upstream-repo

    upstream-repo -->|"GitGitGadget syncs branches to"| pr-repo

    pr-repo -->|"slash commands"| GitGitGadget
    upstream-repo -->|"slash commands (if App installed)"| GitGitGadget

    GitGitGadget -->|"sends patch series"| mailing-list
```

### DIFF
--- a/handle-pr-push/action.yml
+++ b/handle-pr-push/action.yml
@@ -1,0 +1,22 @@
+name: 'Handle PR Pushes'
+description: 'Handles when a PR was pushed'
+author: 'Johannes Schindelin'
+inputs:
+  pr-repo-token:
+    description: 'The access token to work on the repository that holds PRs and state'
+    required: true
+  upstream-repo-token:
+    description: 'The access token to work on PRs in the upstream repository'
+    required: true
+  test-repo-token:
+    description: 'The access token to work on PRs in the test repository'
+    required: true
+  pr-url:
+    description: 'The URL of the PR that was pushed'
+    required: true
+runs:
+  using: 'node20'
+  main: './index.js'
+branding:
+  icon: 'git-commit'
+  color: 'orange'
--- a/lib/ci-helper.ts
+++ b/lib/ci-helper.ts
@@ -67,7 +67,10 @@ export class CIHelper {
         this.urlRepo = `${this.urlBase}${this.config.repo.name}/`;
     }
 
-    public async setupGitHubAction(setupOptions?: { needsMailingListMirror?: boolean }): Promise<void> {
+    public async setupGitHubAction(setupOptions?: {
+        needsMailingListMirror?: boolean;
+        needsUpstreamBranches?: boolean;
+    }): Promise<void> {
         // help dugite realize where `git` is...
         const gitExecutable = os.type() === "Windows_NT" ? "git.exe" : "git";
         const stripSuffix = `bin${path.sep}${gitExecutable}`;
@@ -123,6 +126,9 @@ export class CIHelper {
             ["remote.origin.url", `https://github.com/${this.config.repo.owner}/${this.config.repo.name}`],
             ["remote.origin.promisor", "true"],
             ["remote.origin.partialCloneFilter", "blob:none"],
+            ["remote.upstream.url", `https://github.com/${this.config.repo.baseOwner}/${this.config.repo.name}`],
+            ["remote.upstream.promisor", "true"],
+            ["remote.upstream.partialCloneFilter", "blob:none"],
         ]) {
             await git(["config", key, value], { workDir: this.workDir });
         }
@@ -142,6 +148,41 @@ export class CIHelper {
         );
         console.timeEnd("fetch Git notes");
         this.gggNotesUpdated = true;
+        if (setupOptions?.needsUpstreamBranches) {
+            console.time("fetch upstream branches");
+            await git(
+                [
+                    "fetch",
+                    "origin",
+                    "--no-tags",
+                    "--depth=500",
+                    "--filter=blob:limit=1g",
+                    ...this.config.repo.trackingBranches.map(
+                        (name) => `+refs/heads/${name}:refs/remotes/upstream/${name}`,
+                    ),
+                ],
+                {
+                    workDir: this.workDir,
+                },
+            );
+            console.timeEnd("fetch upstream branches");
+            console.time("get open PR head commits");
+            const openPRCommits = (
+                await Promise.all(
+                    this.config.repo.owners.map(async (repositoryOwner) => {
+                        return await this.github.getOpenPRs(repositoryOwner);
+                    }),
+                )
+            )
+                .flat()
+                .map((pr) => pr.headCommit);
+            console.timeEnd("get open PR head commits");
+            console.time("fetch open PR head commits");
+            await git(["fetch", "--no-tags", "origin", "--filter=blob:limit=1g", ...openPRCommits], {
+                workDir: this.workDir,
+            });
+            console.timeEnd("fetch open PR head commits");
+        }
         // "Unshallow" the refs by fetching the shallow commits with a tree-less filter.
         // This is needed because Git will otherwise fall over left and right when trying
         // to determine merge bases with really old branches.
--- a/lib/ci-helper.ts
+++ b/lib/ci-helper.ts
@@ -18,6 +18,7 @@ import { IPatchSeriesMetadata } from "./patch-series-metadata.js";
 import { IConfig, getExternalConfig, setConfig } from "./project-config.js";
 import { getPullRequestKeyFromURL, pullRequestKey } from "./pullRequestKey.js";
 import { ISMTPOptions } from "./send-mail.js";
+import { fileURLToPath } from "url";
 
 const readFile = util.promisify(fs.readFile);
 type CommentFunction = (comment: string) => Promise<void>;
@@ -874,6 +875,11 @@ export class CIHelper {
         }
     }
 
+    public static async getWelcomeMessage(username: string): Promise<string> {
+        const resPath = path.resolve(fileURLToPath(new URL(".", import.meta.url)), "..", "res", "WELCOME.md");
+        return (await readFile(resPath)).toString().replace(/\${username}/g, username);
+    }
+
     public async handlePush(repositoryOwner: string, prNumber: number): Promise<void> {
         const prKey = {
             owner: repositoryOwner,
@@ -897,7 +903,7 @@ export class CIHelper {
             this.smtpOptions,
         );
         if (!pr.hasComments && !gitGitGadget.isUserAllowed(pr.author)) {
-            const welcome = (await readFile("res/WELCOME.md")).toString().replace(/\${username}/g, pr.author);
+            const welcome = await CIHelper.getWelcomeMessage(pr.author);
             await this.github.addPRComment(prKey, welcome);
 
             await this.github.addPRLabels(prKey, ["new user"]);
--- a/lib/ci-helper.ts
+++ b/lib/ci-helper.ts
@@ -1,6 +1,9 @@
+import * as core from "@actions/core";
 import * as fs from "fs";
+import * as os from "os";
 import * as util from "util";
 import addressparser from "nodemailer/lib/addressparser/index.js";
+import path from "path";
 import { ILintError, LintCommit } from "./commit-lint.js";
 import { commitExists, git, emptyTreeName } from "./git.js";
 import { GitNotes } from "./git-notes.js";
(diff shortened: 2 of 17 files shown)


lib/ci-helper.ts
```diff
@@ -62,6 +64,54 @@ export class CIHelper {
         this.urlRepo = `${this.urlBase}${this.config.repo.name}/`;
     }
 
+    public async setupGitHubAction(): Promise<void> {
+        // help dugite realize where `git` is...
```
webstech, line 75: Would the `find-git-exec` package be more environment agnostic to get the git location?

lib/ci-helper.ts
```diff
@@ -62,6 +64,54 @@ export class CIHelper {
...
+        process.env.LOCAL_GIT_DIRECTORY = "/usr/";
+        process.env.GIT_EXEC_PATH = "/usr/lib/git-core";
+
+        // get the access tokens via the inputs of the GitHub Action
+        this.setAccessToken("gitgitgadget", core.getInput("gitgitgadget-git-access-token"));
```
webstech, line 92: Could the input names be less GitGitGadget specific to make it simpler for other projects?  Maybe more purpose oriented.

lib/ci-helper.ts
```diff
@@ -62,6 +64,54 @@ export class CIHelper {
...
+        this.setAccessToken("git", core.getInput("git-git-access-token"));
+        this.setAccessToken("dscho", core.getInput("dscho-git-access-token"));
+
+        // eslint-disable-next-line security/detect-non-literal-fs-filename
+        if (!fs.existsSync(this.workDir)) await git(["init", "--bare", "--initial-branch", "main", this.workDir]);
+        for (const [key, value] of [
+            ["gc.auto", "0"],
+            ["remote.origin.url", `https://github.com/${this.config.repo.owner}/${this.config.repo.name}`],
```
webstech, line 120: Using `this.config.repo.host` could help with off-github support. 

This applies in many places so it could be handled at another time.

lib/ci-helper.ts
```diff
@@ -62,6 +64,54 @@ export class CIHelper {
...
+        console.time("fetch Git notes");
+        await git(
+            [
+                "-c",
+                "remote.origin.promisor=false", // let's fetch the notes with all of their blobs
+                "fetch",
+                "--no-tags",
+                "origin",
+                "--depth=1",
+                `+${GitNotes.defaultNotesRef}:${GitNotes.defaultNotesRef}`,
```
webstech: Reviewed up to here.  Will delete this when I get further.  Noting this line is updated later in the series.

lib/ci-helper.ts
dscho: I would have loved that! Unfortunately, it seems that the same access token cannot be reused for multiple repositories _unless_ they are in the same org (see [here](https://github.com/actions/create-github-app-token?tab=readme-ov-file#create-a-token-for-multiple-repositories-in-the-current-owners-installation), the underlying limitation is that the access tokens [can only be created per installation](https://docs.github.com/en/rest/apps/apps?apiVersion=2022-11-28#create-an-installation-access-token-for-an-app), which is per-org).

So the main problem here is GitGitGadget's now-problematic historical development, where it first used its own Git fork for all the work (`gitgitgadget/git`), but was later extended to work on `git/git` itself (albeit with a _different_ GitHub App, one that lacks write permission to the repository, for security reasons).

There is probably a good way to talk about this kind of setup, something like "primary-repo" and "own-fork" would be descriptive but probably too unintuitive to be useful when extending GitGitGadget to support 3rd-party projects.

Hmm. @webstech @mjcheetham how about `pr-repo-token`? And `git-git-token` would become `upstream-repo-token`? And `dscho-git-token` would either go away (that would be the easiest) or would become `test-repo-token`?

Maybe I should throw in a diagram?

```mermaid
graph TD
    user["user (contributor)"]
    upstream-repo["upstream-repo (authoritative project repo)"]
    pr-repo["pr-repo (GitGitGadget-enabled GitHub repo)"]
    GitGitGadget["GitGitGadget"]
    mailing-list["mailing-list"]

    user -->|"opens PR"| pr-repo
    user -->|"opens PR (if GitHub App installed)"| upstream-repo

    upstream-repo -->|"GitGitGadget syncs branches to"| pr-repo

    pr-repo -->|"slash commands"| GitGitGadget
    upstream-repo -->|"slash commands (if App installed)"| GitGitGadget

    GitGitGadget -->|"sends patch series"| mailing-list
```

lib/ci-helper.ts
```diff
@@ -62,6 +64,54 @@ export class CIHelper {
...
+        process.env.LOCAL_GIT_DIRECTORY = "/usr/";
+        process.env.GIT_EXEC_PATH = "/usr/lib/git-core";
+
+        // get the access tokens via the inputs of the GitHub Action
+        this.setAccessToken("gitgitgadget", core.getInput("gitgitgadget-git-access-token"));
+        this.setAccessToken("git", core.getInput("git-git-access-token"));
+        this.setAccessToken("dscho", core.getInput("dscho-git-access-token"));
+
+        // eslint-disable-next-line security/detect-non-literal-fs-filename
+        if (!fs.existsSync(this.workDir)) await git(["init", "--bare", "--initial-branch", "main", this.workDir]);
```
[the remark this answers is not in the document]
dscho: Oh, I should have mentioned that the `--initial-branch main` parameter is there _just_ to suppress the ugly warning. It serves no other purpose because upstream Git does not even update its own `main` branch anymore ([it has been deleted in the meantime, even](https://github.com/git/git/branches), showing just how the Git maintainer cared about said LOT of discussion).

lib/ci-helper.ts
```diff
@@ -62,6 +64,54 @@ export class CIHelper {
         this.urlRepo = `${this.urlBase}${this.config.repo.name}/`;
     }
 
+    public async setupGitHubAction(): Promise<void> {
+        // help dugite realize where `git` is...
+        process.env.LOCAL_GIT_DIRECTORY = "/usr/";
+        process.env.GIT_EXEC_PATH = "/usr/lib/git-core";
```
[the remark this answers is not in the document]
dscho: Hmm. I added that mostly because in my VS Code-based testing, `GIT_EXEC_PATH` was set and referred to `/usr/local/lib/git-core/` (because I installed `microsoft/git` there)...

I guess I'll just remove this statement and see whether things still work _on the runners_.

lib/ci-helper.ts
```diff
@@ -62,6 +65,67 @@ export class CIHelper {
...
+            {
+                workDir: this.workDir,
+            },
+        );
+        console.timeEnd("fetch Git notes");
+        this.gggNotesUpdated = true;
+        // "Unshallow" the refs by fetching the shallow commits with a tree-less filter.
+        // This is needed because Git will otherwise fall over left and right when trying
+        // to determine merge bases with really old branches.
+        const unshallow = async (workDir: string) => {
```
webstech, line 199: Was there more happening in `unshallow` during development?  Trying to understand why it is not in-line.

lib/ci-helper.ts
```diff
@@ -150,6 +151,58 @@ export class CIHelper {
...
+            }
+            console.time("fetch mailing list mirror");
+            await git(
+                [
+                    "-c",
+                    "remote.mirror.promisor=false", // let's fetch the mails with all of their contents
+                    "fetch",
+                    "mirror",
+                    "--depth=50",
+                    `+refs/heads/lore-${epoch}:refs/heads/lore-${epoch}`,
```
webstech, line 227: Should `lore` be in the config?  ie could someone follow a similar pattern for a mailing list but not quite follow the `lore.kernel.org` pattern?

lib/ci-helper.ts
```diff
@@ -142,6 +148,41 @@ export class CIHelper {
...
+                ],
+                {
+                    workDir: this.workDir,
+                },
+            );
+            console.timeEnd("fetch upstream branches");
+            console.time("get open PR head commits");
+            const openPRCommits = (
+                await Promise.all(
+                    this.config.repo.owners.map(async (repositoryOwner) => {
```
webstech, line 178: Noting this will pull from `dscho` repo when it is not needed in normal runs.

handle-pr-push/action.yml
```diff
@@ -0,0 +1,22 @@
...
+  pr-repo-token:
+    description: 'The access token to work on gitgitgadget/git'
+    required: true
+  upstream-repo-token:
+    description: 'The access token to work on git/git'
+    required: true
+  test-repo-token:
+    description: 'The access token to work on dscho/git'
+    required: true
+  pr-url:
```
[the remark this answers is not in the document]
webstech: Could be the same name that both `parsePRCommentURLInput` and `parsePRURLInput` use for `getInput`.  The contents are contextual based on the workflow.  Just a different point of view.

webstech:
I have finished reviewing.

lib/ci-helper.ts
dscho: Actually, it would appear that I was playing some games here, seeing as [there are already two epochs of the Git mailing list repository](https://lore.kernel.org/git/_/text/mirror/), where the "0th" epoch had a different tree layout (a wide one, which contained _all_ emails in the tip tree, as opposed to the current layout, where the tip tree contains the most recent email as the `HEAD:m` file). To be able to fetch both into the same mirror repository, I fetched the "0th epoch" into `lore-0` and the "1st epoch" into `lore-1`.

So it's probably safe to assume that whatever `public-inbox`-based mailing list mirror is used will have to potentially deal with different epochs, too, and come up with a similar naming scheme as I did.

But yes, it's unlikely to be the naming scheme `refs/heads/lore-${epoch}` that _I_ chose ;-) Therefore, I should _totally_ move the actual ref name into the config, similar to [the mirror URL](https://github.com/gitgitgadget/gitgitgadget/blob/0f5158b4338c43f63806b23dd3156c6748119a40/lib/gitgitgadget-config.ts#L23) of the Git repository.

For the record, I _do_ believe that we _must_ work with a mailing list mirror repository from which one _can_ make a partial clone; Trying to clone with `--filter` from, say, `lore.kernel.org` will result in this warning: "warning: filtering not recognized by server, ignoring".

mjcheetham (approved):

lib/ci-helper.ts
```diff
@@ -150,6 +151,58 @@ export class CIHelper {
...
+            }
+            console.time("fetch mailing list mirror");
+            await git(
+                [
+                    "-c",
+                    "remote.mirror.promisor=false", // let's fetch the mails with all of their contents
+                    "fetch",
+                    "mirror",
+                    "--depth=50",
+                    "+REF:REF".replace("REF", this.config.mailrepo.mirrorRef || `refs/heads/lore-${epoch}`),
```
webstech, line 227: Does `mirrorRef` need to include `refs/heads/`?  I suppose you _could_ put things pretty much anywhere you want in a repo.

PR #1988, opened 2025-08-28
Cc: the current Git GUI maintainer on corresponding patches

dscho:
I just noticed that Pratyush was still Cc:ed by default. Then I looked and was puzzled because I thought that [the `project` settings would be used](https://github.com/gitgitgadget/gitgitgadget/blob/631cfc116134c93f4c8ffa95a7d39e83392b0337/lib/project-options.ts#L19-L26) instead of [auto-magically figuring out](https://github.com/gitgitgadget/gitgitgadget/blob/631cfc116134c93f4c8ffa95a7d39e83392b0337/lib/project-options.ts#L27-L31) whether the PR in question is a Git GUI one. But then I saw that [the `defaultConfig`](https://github.com/gitgitgadget/gitgitgadget/blob/631cfc116134c93f4c8ffa95a7d39e83392b0337/lib/gitgitgadget-config.ts#L3-L45) does not have a `project` attribute ;-)

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -73,6 +73,7 @@
         "refname",
         "repo",
         "Schindelin",
+        "Sixt",
         "smee",
         "smtp",
         "stolee",
--- a/lib/project-options.ts
+++ b/lib/project-options.ts
@@ -27,7 +27,7 @@ export class ProjectOptions {
         } else if ((await revParse(`${baseCommit}:git-gui.sh`, workDir)) !== undefined) {
             // Git GUI
             to = "--to=git@vger.kernel.org";
-            cc.push("Pratyush Yadav <me@yadavpratyush.com>");
+            cc.push("Johannes Sixt <j6t@kdbg.org>");
             upstreamBranch = "git-gui/master";
         } else if ((await revParse(`${baseCommit}:git.c`, workDir)) !== undefined) {
             // Git


webstech:
This was probably mentioned somewhere as an outstanding issue.  The `cygwin`, `busybox` and test suite can be handled by a config but I left them for compatibility until we got there.  The handling of `git-gui` and `upstream-branch` is more problematic.  Didn't really want to add functions to the config.  Any thoughts on this?  It may be a git specific problem.

webstech (approved):

dscho:
We'll have to figure this out on the go, as we truly support more projects.

webstech:
> @webstech how does that plan sound to you?

I mentioned in #1993 I thought you had introduced something for handling this but maybe it was in the sync remotes stuff.  Sorry for any confusion.  

Yes, making project required is a good thing.  

Handling special cases is difficult.  For `git-gui` the commit has a specific ancestor.   Alternatively, it could be the path of any changed files are in a particular directory tree.  These could be filters in the project.

Having said that, the special case will work for now and other projects may drive a generic solution.

Just to re-iterate, `ProjectOptions` is only used in `PatchSeries` so it may be the config can be used instead now (except for cc and upstreamBranch at the moment).

dscho:
> `ProjectOptions` is only used in `PatchSeries` so it may be the config can be used instead now (except for cc and upstreamBranch at the moment).

Right. And `upstreamBranch` is unlikely necessary in the Pull Request workflow (it was needed to emulate [`mail-patch-series.sh`](https://github.com/dscho/mail-patch-series), back when I wanted to be able to use GitGitGadget _also_ as a replacement for that script).

webstech:
> For the record, this is how I am intending to go forward with this: [dscho@fb72cd6#diff-f15268abff24be0d82e67bb03d2d639789d27a2b734fdf93cf5e6c7fca61e44fL25-R31](https://github.com/dscho/gitgitgadget/commit/fb72cd650dcb6f0bf2a3691217553a8dab68afbb#diff-f15268abff24be0d82e67bb03d2d639789d27a2b734fdf93cf5e6c7fca61e44fL25-R31)

Missing something?
`cc.push("Johannes Sixt <j6t@kdbg.org>");`

PR #1991, opened 2025-09-01
Actions: start accepting a `config` as input

dscho:
GitGitGadget's GitHub Actions all implicitly use the [`defaultConfig`](https://github.com/gitgitgadget/gitgitgadget/blob/153badfc78153bb25d10e0e0d40c2cb3924d7f1b/lib/gitgitgadget-config.ts#L3-L45) that hard-codes the configuration of the OG GitGitGadget that supports the Git project (and the Git project _only_).

However, for a long time there have been feature requests and musings towards using GitGitGadget also for other projects.

GitGitGadget already _does_ have some beginnings of a support for other projects, e.g. the `IConfig` interface. With these changes, that interface gets extended a little, a validator is added that can verify whether or not an arbitrary object conforms to that interface, and then new inputs are introduced, accepting this (JSON-encoded) configuration as a user-provided string. The idea is that a project-specific fork of https://github.com/gitgitgadget/gitgitgadget-workflows/ contains this configuration in the `gitgitgadget-config.json` file in a dedicated `config` branch, from where it is synchronized via a GitHub workflow to the repository variable called `CONFIG`.

This somewhat non-trivial setup allows the config to be conveniently tracked via Git, updated via Pull Requests, validated via GitHub workflows, and still to be used in a trivial manner in the main workflows via `${{ vars.CONFIG }}` (as opposed to having to play games with multi-branch checkouts or `curl`'ing a file from a different branch).

This PR is step number 2 in that direction. Step number 1 was to register the [`cygwingitgadget` GitHub org](https://github.com/cygwingitgadget/) for experimenting with GitGitGadget support for a different mailing-list-based project than Git: Cygwin.

That org is where I experimented with this change as well as with all the others leading up to https://github.com/gitgitgadget/gitgitgadget-github-app/pull/7.

This PR is stacked on top of #1993 and #1994

### DIFF
--- a/lib/ci-helper.ts
+++ b/lib/ci-helper.ts
@@ -909,7 +909,7 @@ export class CIHelper {
                     await addComment(
                         `Submitted as [${
                             metadata?.coverLetterMessageId
-                        }](https://${this.config.mailrepo.host}/${this.config.mailrepo.name}/${
+                        }](${this.config.mailrepo.url.replace(/\/+$/, "")}/${
                             metadata?.coverLetterMessageId
                         })\n\nTo fetch this version into \`FETCH_HEAD\`:${
                             code


lib/project-config.ts
```diff
@@ -8,49 +8,70 @@ export type projectInfo = {
     urlPrefix: string; // url to 'listserv' of mail (should it be in mailrepo?)
 };
 
+export interface IRepoConfig {
+    name: string; // name of the repo
+    owner: string; // owner of repo holding the notes (tracking data)
+    baseOwner: string; // owner of upstream ("base") repo
```
dscho, line 14: @webstech since I started referring to the corresponding repository as `upstream-repo`, what would you think about the idea to rename this from `baseOwner` to `upstreamOwner`?

lib/project-config.ts
```diff
@@ -8,49 +8,70 @@ export type projectInfo = {
...
+    testOwner?: string; // owner of the test repo (if any)
+    owners: string[]; // owners of clones being monitored (PR checking)
```
dscho, line 16: @webstech what do you think about the idea to move this attribute to `IAppConfig` and to rename it to `installedOn`? That way, the `repo` still can have an `upstreamOwner` on whose repository, however, GitGitGadget is _not_ installed...

lib/project-config.ts
webstech: I have no issue with renaming.  _My_ choice in names may not always be as git oriented as they could be.

lib/project-config.ts
webstech: This is used several times in `CIHelper` so the location could be justified.  On the other hand, move it if you like.

webstech:
An aside while I was looking at this, I noticed `CIHelper.parsePRURLInput()` could be using `getPullRequestKeyFromURL`.

dscho:
> An aside while I was looking at this, I noticed `CIHelper.parsePRURLInput()` could be using `getPullRequestKeyFromURL`.

True! I have implemented that (still need to clean up the branch thicket, it needed fixes): https://github.com/dscho/gitgitgadget/commit/8b3badbd4cd0aaed77646c179fc3aa82c8c95eb4

> Note to self: check whether `default: ${{ vars.CONFIG }}` works, and would avoid the need for the explicit `config: ${{ vars.CONFIG }}` lines added in [gitgitgadget/gitgitgadget-workflows@2952485](https://github.com/gitgitgadget/gitgitgadget-workflows/commit/2952485401af144a47eb52d5e9af6dec0a925c24).

Okay, this does [**_not_** work](https://github.com/dscho/gitgitgadget/actions/runs/17606375501/job/50017946981#step:3:1):

image

<details><summary>as text</summary>

```
Error: /home/runner/work/gitgitgadget/gitgitgadget/./check-run/action.yml (Line: 7, Col: 14): Unrecognized named-value: 'vars'. Located at position 1 within expression: vars.CONFIG
Error: Failed to load /home/runner/work/gitgitgadget/gitgitgadget/./check-run/action.yml
```

</details>

Sadness!

Even more sad is [the excuse of a documentation of `inputs.<input_id>.default`](https://docs.github.com/en/actions/reference/workflows-and-actions/metadata-syntax#inputsinput_iddefault) which does not even mention that _some_ contexts _can_ be used, as [some smarter people found out](https://github.com/orgs/community/discussions/25218). But of course nothing official, because why would anyone want to know what they can use, and what not?&lt;/s>

lib/project-config.ts
dscho: For the record, I am going forward with this: https://github.com/dscho/gitgitgadget/commit/2de5309eea4d9c1dfa387e9e69eec58c3aa24db0

dscho:
@webstech would you kindly have a look?

PR #2091, opened 2025-12-19
commit-lint: prefix error messages with bullet points

HaraldNordgren:
I used gitgitgadget today and it's a very nice tool, but I was a bit confused about the warning messages.

I figured bullet points could make it much clearer to other developers.


HaraldNordgren:
Ping @dscho

PR #2097, opened 2025-12-23
mail-archive: clarify and bold message author

HaraldNordgren:
I find it a bit confusing when scrolling up and down, it's no immediately clear who wrote what.

It seems to me that in the information hierarchy the message author is most import.

Screenshot 2025-12-23 at 16 09 49

### DIFF
--- a/lib/mail-archive-helper.ts
+++ b/lib/mail-archive-helper.ts
@@ -191,9 +191,10 @@ export class MailArchiveGitHelper {
             );
 
             const archiveURL = `${this.config.mailrepo.url}${parsed.messageID}`;
-            const header = `[On the Git mailing list](${archiveURL}), ${
-                parsedMbox.from ? parsedMbox.from.replace(/ *<.*>/, "") : "Somebody"
-            } wrote ([reply to this](${replyToThisURL})):\n\n`;
+            const authorName = parsedMbox.from ? parsedMbox.from.replace(/ *<.*>/, "") : "Somebody";
+            const header =
+                `**${authorName}** wrote [on the Git mailing list](${archiveURL}) ` +
+                `([how to reply to this email](${replyToThisURL})):\n\n`;
             const comment = MailArchiveGitHelper.mbox2markdown(parsedMbox);
             const fullComment = header + comment;
             const prKey = getPullRequestKey(pullRequestURL);
@@ -212,7 +213,7 @@ export class MailArchiveGitHelper {
                     issueCommentId = result.id;
                 } catch (_error) {
                     const commits = await this.githubGlue.getPRCommits(prKey.owner, prKey.pull_number);
-                    const regarding = `${header.slice(0, -3)}, regarding ${originalCommit}${
+                    const regarding = `${header.slice(0, -3)} regarding ${originalCommit}${
                         commits.find((e) => e.commit === originalCommit) ? "" : " (outdated)"
                     }:\n\n`;
                     await this.githubGlue.addPRComment(pullRequestURL, regarding + comment);
--- a/tests/mail-archive-helper.test.ts
+++ b/tests/mail-archive-helper.test.ts
@@ -617,3 +617,70 @@ This Pull Request contains some ipsum lorem.
     expect(data).toBeDefined();
     expect(data?.pullRequestURL).toEqual(mailMeta.pullRequestURL);
 });
+
+test("test comment format with author name and links", async () => {
+    const repo = await testCreateRepo(sourceFileName, "-format");
+    await repo.commit("1", "1", "");
+
+    const mailMeta: IMailMetadata = {
+        messageID: "pull.12345.v1.git.gitgitgadget@example.com",
+        originalCommit: undefined,
+        pullRequestURL: `https://github.com/${config.repo.owner}/${config.repo.name}/pull/1`,
+        firstPatchLine: 5,
+    };
+
+    const replyMessageId = `i${mailMeta.messageID}`;
+    const authorName = "Harald Nordgren";
+
+    const mbox0 = `From 566155e00ab72541ff0ac21eab84d087b0e882a5 Mon Sep 17 00:00:00 2001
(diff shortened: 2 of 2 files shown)


webstech:
I see your point.  The current message is 'where, who'.  I think it would read better as 'who where'.  

Would this format work for you?  It doesn't add any lines to the comment in the PR.
**joe shmoe** wrote [on the Git mailing list](https://lore.kernal.org/blah/blah/pull/ff.ff):   ([how to reply to this email](https://gitgitgadget.github.io/reply-to-this))

The proposed changes suggest you can reply through the mailing list which is misleading.
As an aside, for me, bold adds very little visibility to the text when using dark mode (windows/firefox combo - hope it is not my eyes 😒).

I understand it can be confusing the first time.  You should be copied on all replies unless someone only replies to the list or drops you from a `cc:`.  Usually, you can reply from your email client (in text mode).

Rereading '[how to reply](https://gitgitgadget.github.io/reply-to-this)', maybe it  should be trying to cover a couple of circumstances.  

1. I got an email - how to reply referencing info in git-scm.
2. I do not have an email (repeating or referencing info from git-scm)
   1. how to download mbox
   2. choose a delivery method
      3. how to load mbox into email provider
      4. how to configure, edit and send using `git send-email`

@HaraldNordgren What do you think now that you have been through this a first time?  You know most of what to do now but what would help new users without being too noisy over time?
@dscho Thoughts?

lib/mail-archive-helper.ts
```diff
@@ -212,7 +213,7 @@ export class MailArchiveGitHelper {
                     issueCommentId = result.id;
                 } catch (_error) {
                     const commits = await this.githubGlue.getPRCommits(prKey.owner, prKey.pull_number);
-                    const regarding = `${header.slice(0, -3)}, regarding ${originalCommit}${
+                    const regarding = `**${authorName}** wrote, regarding ${originalCommit}${
```
webstech: This is an error type case so why not keep using `header`?  It has a link to the email on the mailing list, which may be helpful.

tests/mail-archive-helper.test.ts
```diff
@@ -617,3 +617,108 @@ This Pull Request contains some ipsum lorem.
...
+
+    let commentBody = "";
+
+    // eslint-disable-next-line @typescript-eslint/require-await
+    github.octo.hook.wrap("request", async (_request, options) => {
+        if ("/repos/{owner}/{repo}/pulls/{pull_number}" === options.url) {
+            return prListResponse;
+        }
+        if ("/repos/{owner}/{repo}/issues/{issue_number}/comments" === options.url) {
+            const body =
```
webstech: Sorry but I am not clear on these checks.  If they are needed should they be added to the other tests?

tests/mail-archive-helper.test.ts
```diff
@@ -617,3 +617,108 @@ This Pull Request contains some ipsum lorem.
...
+                typeof options.body === "object" && options.body !== null && "body" in options.body
+                    ? (options.body as { body: string }).body
+                    : (options.body as string);
+            if (body.includes("**") && body.includes("wrote")) {
```
webstech: The helper should not be deciding the test.  Maybe `commentBody` should be set for later checking outside of the called function.

tests/mail-archive-helper.test.ts
```diff
@@ -617,3 +617,108 @@ This Pull Request contains some ipsum lorem.
...
+
+    const logSpy = jest.spyOn(console, "log").mockImplementation(() => {});
+    await mail.processMails();
+    logSpy.mockRestore();
+
+    expect(commentBody).toBeDefined();
+    const lines = commentBody.split("\n");
+    expect(lines.length).toBe(7);
+    expect(lines[0]).toBe(
+        `**${authorName}** wrote [on the Git mailing list](https://lore.kernel.org/git/${replyMessageId}) ` +
```
webstech: Someday the `config.mailrepo.url` may be needed for other project testing.  Maybe not but just taking note of it.

webstech: Alternatively, the test could verify just the start of the line and ignore the mailing list URL.

tests/mail-archive-helper.test.ts
```diff
@@ -617,3 +617,108 @@ This Pull Request contains some ipsum lorem.
...
+            `([how to reply to this email](https://gitgitgadget.github.io/reply-to-this))`,
+    );
+    expect(lines[1]).toBe("");
```
webstech: I didn't expect these test lines.  What are they proving?

tests/mail-archive-helper.test.ts
```diff
@@ -617,3 +617,95 @@ This Pull Request contains some ipsum lorem.
...
+        }
+        if ("/repos/{owner}/{repo}/issues/{issue_number}/comments" === options.url) {
+            const body =
+                typeof options.body === "object" && options.body !== null && "body" in options.body
+                    ? (options.body as { body: string }).body
+                    : (options.body as string);
+            commentBodies.push(body);
+            return issueCommentResponse;
+        }
+        if ("/repos/{owner}/{repo}/pulls/{pull_number}/comments/{comment_id}/replies" === options.url) {
```
webstech: This request is not seen in this test.

tests/mail-archive-helper.test.ts
webstech: Maybe I wasn't clear.  Did you encounter issues that needed the checks/conversion for setting `body`?  The previous test case just set it `commentBody = options.body as string;`.  The request is a POST that is supposed to have a `body` string.  Have you seen something different?

HaraldNordgren:
@webstech Updated again!

Regarding squashing, I would suggest changing to only allow squash merge, then users won't need to squash, GitHub will handle it for us:

Screenshot 2026-01-08 at 09 25 35

HaraldNordgren:
@webstech Pushed new changes!

PR #2212, opened 2026-04-16
range-diff: rebase old series onto new base to reduce upstream "noise"

HaraldNordgren:
This is based on a discussion between me and JCH (https://lore.kernel.org/git/xmqqfr4xcz7s.fsf@gitster.g/), thought I could work on possibly making the range-diff more useful.

When the contributor rebases their patch series onto a newer upstream between iterations, the range-diff previously showed noise from upstream changes alongside the actual patch modifications. Now rebases the old series onto the new base before running `git range-diff`, so only the real differences between iterations are shown. Falls back to the original range-diff if the rebase fails (e.g. due to conflicts).

### DIFF
--- a/lib/patch-series.ts
+++ b/lib/patch-series.ts
@@ -724,7 +724,7 @@ export class PatchSeries {
    )}
    git range-diff <options> ${getRange(this.rangeDiff.previousRange)} ${getRange(this.rangeDiff.currentRange)}`);
             } else {
-                const previousRange = await this.rebaseOldSeriesIfNeeded();
+                const previousRange = await this.rebaseOldSeriesIfNeeded(logger);
 
                 const rangeDiff = await git(
                     ["range-diff", "--no-color", "--creation-factor=95", previousRange, this.rangeDiff.currentRange],
@@ -866,7 +866,7 @@ export class PatchSeries {
         return this.metadata;
     }
 
-    protected async rebaseOldSeriesIfNeeded(): Promise<string> {
+    protected async rebaseOldSeriesIfNeeded(logger: ILogger): Promise<string> {
         if (!this.rangeDiff || this.rangeDiff.oldBaseCommit === this.rangeDiff.newBaseCommit) {
             return this.rangeDiff!.previousRange;
         }
@@ -885,6 +885,7 @@ export class PatchSeries {
                 { workDir },
             );
             const rebasedHead = (await git(["rev-parse", "HEAD"], { workDir })).trim();
+            logger.log("Rebased old series onto new base for range-diff");
             return `${this.rangeDiff.newBaseCommit}..${rebasedHead}`;
         } catch (_) {
             try {
@@ -892,6 +893,7 @@ export class PatchSeries {
             } catch (__) {
                 // Just ignore it
             }
+            logger.log("warning: could not rebase old series onto new base, falling back to cross-base range-diff");
             return this.rangeDiff.previousRange;
         } finally {
             await git(["checkout", origHead], { workDir });
--- a/lib/patch-series.ts
+++ b/lib/patch-series.ts
@@ -51,8 +51,10 @@ const singletonHeaders: ISingletonHeader[] = [
 interface IRangeDiff {
     previousRange: string;
     currentRange: string;
-    baseCommit: string;
-    headCommit: string;
+    oldBaseCommit: string;
+    oldHeadCommit: string;
+    newBaseCommit: string;
+    newHeadCommit: string;
 }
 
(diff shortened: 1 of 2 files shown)


dscho:
I'm not so sure. When I look at https://lore.kernel.org/git/pull.2234.v12.git.git.1776171585.gitgitgadget@gmail.com/, I essentially see a long range-diff, but it's correct, and I don't think that it would become easier to read when comparing a rebased version of the previous iteration instead of the non-rebased version. The range-diff basically tells you that the changes that previously had been somewhere else are now in a new place. And the range-diff shows exactly that. So I don't understand what Junio is complaining about.

Your solution very much reminds me of what I had in the predecessor of GitGitGadget before I ported `range-diff` (then named `tbdiff`) to C. Basically, the tool did rebase for an "inter-diff" all the time, which resulted in many a merge conflict and prevented the whole thing from working correctly and robustly. The _entire point_ of introducing `git range-diff` and then using it in GitGitGadget _was_ to step away from that fragile rebase-using process.

The true problem here, of course, is that GitGitGadget is forced to pre-generate information that _might_ be useful, all because of the Git project's choices. A mailing list is simply not a good place to do code review because you don't have any dynamic tools like interactive git commands to inspect the code changes. You cannot increase the hunk context, you cannot look at the diffs with options like `--color-moved` or `-w`. You cannot look at the commit history of the touched code via `-L <line-range>:<path>`. None of that is the fault of GitGitGadget. Rather, it's due to the, let's say, "interesting" choice of trying to do everything ranging from project management to bug tracking to code review on a mailing list. GitGitGadget cannot predict all the ways in which the information from the Git branch should be transformed into static information to be passed along to the mailing list.

If fault is to be sought, it's probably mine for thinking that I can bring a couple of modern practices into the Git project by providing the convenience of GitGitGadget. But instead of recognizing the benefit of modern practices and modern tools rather than doing everything on a mailing list, and embracing more of those by allowing code contributions that do _not_ "evacuate" diffs from Git, we try to stuff more and more features into GitGitGadget to compensate for the questionable choice of using a mailing list for code review. That's of course not going to ever work, because a mailing list is simply not the right tool for the job, it's like trying to drive screws into a wall using a hammer, and no matter what we're trying to do here is going to really help because the crux is, and will always be, the limitation of forcing code review to be performed on a mailing list.

So: Sure, we can _try_ what you propose here, but it'll never be good enough. And it can also be very misleading because "helpfully" rebasing the previous iteration _pretends_ something that never was a reality. If a rebase happened between iterations, then that rebase happened, with all the changes that were necessary. (And most notably, with all the ramifications on test results from the original contribution, read: the rebased version has likely received less thorough testing than the non-rebased version, something that has bitten me time and time again when my contributions worked in my branch, but did not work on the base commit onto which Junio chose to apply them.)

Essentially, I'm dubious that this will actually improve anything. Yeah, sure. It's trying to accommodate Junio, implement what they asked for, but I don't deem it sensible to have asked for it. The only solution, really, would be for the Git project to adopt more modern practices, to embrace more modern tools. At this stage I think that's unlikely. Of course we can try to do in GitGitGadget what we can, what is reasonable, but where to draw the line? Unfortunately, I think this here PR is crossing that line of what's reasonable and helpful.

lib/patch-series.ts
```diff
@@ -861,6 +866,38 @@ export class PatchSeries {
         return this.metadata;
     }
 
+    protected async rebaseOldSeriesIfNeeded(): Promise<string> {
+        if (!this.rangeDiff || this.rangeDiff.oldBaseCommit === this.rangeDiff.newBaseCommit) {
+            return this.rangeDiff!.previousRange;
+        }
+
+        const workDir = this.project.workDir;
```
dscho, line 874: The name `workDir` is misleading. Back in the days when GitGitGadget ran on a dedicated self-hosted Azure Pipelines runner, it did have a worktree to work with.

Nowadays, however, when this is running on GitHub Actions, and has to re-initialize the setup every single time a `/submit` is performed, it is a bare repository containing a partial clone.

lib/patch-series.ts
```diff
@@ -861,6 +866,38 @@ export class PatchSeries {
...
+        const origHead = (await git(["rev-parse", "HEAD"], { workDir })).trim();
+        try {
+            await git(
+                [
+                    "rebase",
```
dscho, line 879: Since this is running in a bare repository, we could try to use `replay` here, but `rebase` won't work.

HaraldNordgren:
Hmm, closing this now. It doesn't seem salvageable.

PR #2219, opened 2026-04-25
patch-series: publish a moving `-latest` branch on /submit

HaraldNordgren:
Alongside the immutable per-iteration tag, also publish a force-moved `-latest` branch pointing at the most recent submission, and add a fetch line for it in the `/submit` reply. Using a branch (rather than a tag) means consumers can refresh it with plain `git fetch`.

I often run something like `git diff pr-git-2281/HaraldNordgren/checkout-fetch-start-point-v1...` to write a good cover letter for the next version, and it's annoying to keep track of exactly which is the latest version.

### DIFF
--- a/lib/ci-helper.ts
+++ b/lib/ci-helper.ts
@@ -906,6 +906,11 @@ export class CIHelper {
 
                     const metadata = await gitGitGadget.submit(pr, userInfo);
                     const code = "\n```";
+                    const aliasSection = metadata?.latestBranch
+                        ? `\n\nTo always fetch the latest iteration into \`FETCH_HEAD\`:${
+                              code
+                          }\ngit fetch ${this.urlRepo} ${metadata.latestBranch}${code}`
+                        : "";
                     await addComment(
                         `Submitted as [${
                             metadata?.coverLetterMessageId
@@ -917,7 +922,9 @@ export class CIHelper {
                             code
                         }\n\nTo fetch this version to local tag \`${metadata?.latestTag}\`:${
                             code
-                        }\ngit fetch --no-tags ${this.urlRepo} tag ${metadata?.latestTag}${code}${extraComment}`,
+                        }\ngit fetch --no-tags ${this.urlRepo} tag ${metadata?.latestTag}${code}${
+                            aliasSection
+                        }${extraComment}`,
                     );
                 }
             } else if (command === "/preview") {
--- a/lib/patch-series-metadata.ts
+++ b/lib/patch-series-metadata.ts
@@ -7,6 +7,7 @@ export interface IPatchSeriesMetadata {
     iteration: number;
     coverLetterMessageId?: string;
     latestTag?: string;
+    latestBranch?: string;
     referencesMessageIds?: string[];
     tipCommitInGitGit?: string;
     branchNameInGitsterGit?: string;
--- a/lib/patch-series.ts
+++ b/lib/patch-series.ts
@@ -649,6 +649,7 @@ export class PatchSeries {
             this.metadata.referencesMessageIds,
         );
         let tagName: string | undefined;
+        let latestBranch: string | undefined;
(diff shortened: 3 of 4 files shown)


dscho:
I'm not a fan. The already disgustingly wasteful CI of the regular Git branches would then kick in for even more refs, blocking many a contributor by delaying _their_ (legitimate) PR builds from even starting.

HaraldNordgren:
@dscho There must be a way to block CI from running for these branches? I haven't looked into how any of that is set up, but GitHub Actions has rules for it.

webstech:
@HaraldNordgren You started this patch as adding a tag and changed it to adding a branch.  I did not get a chance to review the earlier iteration.  What is the motivation for the change?  Sorry if I am missing something obvious.   Adding a branch is more expensive.

`I often run something like git diff pr-git-2281/HaraldNordgren/checkout-fetch-start-point-v1... to write a good cover letter for the next version, and it's annoying to keep track of exactly which is the latest version.
`
When is the latest version different from the version in your local repo when you are updating the PR?  Just trying to understand the problem.

HaraldNordgren:
@webstech The reason for changing from tag to branch is I realized that the tag will likely require the users to run `git fetch --force` to overwrite `-latest`, whereas a branch doesn't require any voodoo like that, you can just do regular `git fetch` and it will overwrite.

> When is the latest version different from the version in your local repo when you are updating the PR? Just trying to understand the problem.

When I keep pushing out new versions, like here (https://github.com/git/git/pull/2281), I don't keep track of which is the latest tag (it's v4 here currently), so I keep redoing this command to grab latest:

```
git fetch --no-tags https://github.com/gitgitgadget/git/ tag pr-git-2281/HaraldNordgren/checkout-fetch-start-point-v4
```

Realistically, instead of opening up GitHub, or the Git mailing list, I will manually increment `checkout-fetch-start-point-v3`, `checkout-fetch-start-point-v4`, `checkout-fetch-start-point-v5` until it doesn't exist.

Then I diff that against my latest changes (that are usually pushed to GitHub because I want the CI to pass first before a submission), and then describe those changes in the PR description on GitHub, to be included as a cover letter for the new patch when running `/submit`.

dscho:
Since you _already_ have to memorize the exact PR number for that to work, it should be _really_ easy to script this. Something like `git ls-remote https://github.com/gitgitgadget/git refs/tags/pr-git-2281\*` should get you started. You could, for example, pipe it to `| sed -n 's/.*\t\(.*-v\([0-9]*\)\)$/\2\t\1/p' | sort -n -k 1 -r | head -n 1 | cut -f 2` and then take it from there. As a bonus of scripting it like that, you could then also use the GitHub CLI to automatically find the PR number corresponding to your local branch, and even cache that information in the repository-local config under a key of your choosing.

The reason why I am still opposed to the approach taken in this PR? It sounds very much like a process problem on the local side, and not something that needs to clutter the gitgitgadget/git repository even more than it already is.

Besides, the current implicit contract is that any branch that lives in this repository is from upstream Git and therefore trusted. This PR would very much change that and break that contract.

In short, I don't think this is the right layer to tackle the problem. It obviously adds more problems, and the issues that that have been pointed out by @webstech and myself are indicative that this is the wrong direction.

Sure, you can fix those problems by addressing them in ways that squelch that _particular_ feedback you received. In all likelihood, however, the approach will keep raising issues of similar nature and impact.

HaraldNordgren:
Sure, it can be scripted like this 👍

```bash
pr=$(gh pr list --head "$(git branch --show-current)" --json number --jq '.[0].number')
latest=$(git ls-remote --sort=-version:refname --refs https://github.com/gitgitgadget/git 'refs/tags/pr-git-'$pr'/*' | head -1 | cut -f2 | sed 's#^refs/tags/##')

git fetch --no-tags https://github.com/gitgitgadget/git/ tag $latest && git diff $latest..
```

PR #2245, opened 2026-05-28
Switch GitHub Actions runtime from node20 to node24

dscho:
GitHub Actions [is phasing out the `node20` runtime](https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/) on hosted runners, with `node24` as the successor. Since GitGitGadget is only ever executed on GitHub-hosted runners (never on self-hosted ones), there is no consumer whose runner might still only ship
`node20`, and therefore no need to bump the major version of the published `v1` Actions: the existing `v1` tag can keep tracking the latest code.

This branch walks the project from `node20` to `node24` end to end, in small focused commits:

1. `package.json: require Node.js 24` raises `engines.node` (and the matching entry in `package-lock.json`) from `>= 18.16.1` to `>= 24`.
2. `actions: switch the runtime from node20 to node24` updates all five published `action.yml` files in lockstep.
3. `ci(push-github-actions): regenerate dist/ on Node.js 24` makes the workflow that produces the `v1` bundle use the same runtime that will load it.
4. `ci(build-and-test): exercise CI on Node.js 24` moves the GitHub Actions test matrix to `24.x`.
5. `ci(azure-pipelines): install Node.js 24` mirrors that bump in the optional Azure Pipelines definition.
6. `build(deps-dev): bump ts-patch from 3.3.0 to 4.0.1` folds in #2226: ts-patch 4.x raises its own Node floor to `>=22.12.0`, which now matches what we declare.

`npm ci`, `npm run lint`, `npm run build`, and `npm test` were all run locally on Node.js 24; the tracked test suite (14 suites, 223 tests) passes cleanly.


webstech (approved):
Thanks for looking after this.

PR #2253, opened 2026-06-03
ci-helper: only comment on first integration into a branch

### LINKED ISSUE #1646: Avoid adding new "This was integrated into" comments
It is nice to have the current information about the state of any given patch series. However, having a new comment whenever it is integrated into `seen` (which can happen multiple times a day) is not _so_ helpful, seeing as it increases "noise".

Let's find a previous comment to update instead.

While at it, add a comment for patch series that had been in `seen` (or `next`) but dropped out of it.

HaraldNordgren:
Fixes https://github.com/gitgitgadget/gitgitgadget/issues/1646.

Only post the "integrated into" comment the first time a series enters a branch; when the merge commit merely changes (e.g. a rebase) the label is already present, so the notes are updated silently.

### DIFF
--- a/lib/ci-helper.ts
+++ b/lib/ci-helper.ts
@@ -662,19 +662,25 @@ export class CIHelper {
             if (!prMeta.mergedIntoUpstream) {
                 prMeta.mergedIntoUpstream = {};
             }
-            if (prMeta.mergedIntoUpstream[branch] !== mergeCommit) {
+            const previousMergeCommit = prMeta.mergedIntoUpstream[branch];
+            if (previousMergeCommit !== mergeCommit) {
                 prMeta.mergedIntoUpstream[branch] = mergeCommit;
                 notesUpdated = true;
 
                 // Add label on GitHub
                 prLabelsToAdd.push(branch);
 
-                // Add comment on GitHub
-                const comment = `This patch series was integrated into ${branch} via https://github.com/${
-                    this.config.repo.upstreamOwner
-                }/${this.config.repo.name}/commit/${mergeCommit}.`;
-                const url = await this.github.addPRComment(prKey, comment);
-                console.log(`Added comment ${url.id} about ${branch}: ${url.url}`);
+                // Only comment when the branch is newly integrated, i.e. when
+                // the label is not there yet. When the merge commit merely
+                // changes (e.g. due to a rebase), the label is already present,
+                // so we update the notes silently.
+                if (previousMergeCommit === undefined) {
+                    const comment = `This patch series was integrated into ${branch} via https://github.com/${
+                        this.config.repo.upstreamOwner
+                    }/${this.config.repo.name}/commit/${mergeCommit}.`;
+                    const url = await this.github.addPRComment(prKey, comment);
+                    console.log(`Added comment ${url.id} about ${branch}: ${url.url}`);
+                }
             }
         }
 
--- a/lib/ci-helper.ts
+++ b/lib/ci-helper.ts
@@ -649,9 +649,23 @@ export class CIHelper {
 
         let upstreamMergeCommit: string | undefined;
         const prLabelsToAdd: string[] = [];
+        const prLabelsToRemove: string[] = [];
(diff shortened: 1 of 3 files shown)


lib/ci-helper.ts
```diff
@@ -668,13 +668,6 @@ export class CIHelper {
 
                 // Add label on GitHub
                 prLabelsToAdd.push(branch);
-
-                // Add comment on GitHub
-                const comment = `This patch series was integrated into ${branch} via https://github.com/${
-                    this.config.repo.upstreamOwner
-                }/${this.config.repo.name}/commit/${mergeCommit}.`;
-                const url = await this.github.addPRComment(prKey, comment);
-                console.log(`Added comment ${url.id} about ${branch}: ${url.url}`);
```
dscho, line 677: You are affected a lot more than other contributors because of the amount of parallel contributions you have open at any given time ;-)

Could we maybe _replace_ rather than _remove_ this?

I see that line 670 adds `branch` to the `prLabelsToAdd`. Maybe we can use that as a guard: If the label had already been there, suppress the comment. If it has not been there yet, add the comment.

Of course, such a change cries out loud for a corresponding "if a branch drops out of a branch, say so in a new PR comment" logic...

HaraldNordgren: Just to be clear: It's not the amount of emails I'm worried about, it's that I can't properly see the review history of a PR without pressing "Load diff" many times. So it would be the same even with only one topic open 😃

I think you idea is interesting and I will give it a shot as soon I'm at my computer again.

HaraldNordgren: > Of course, such a change cries out loud for a corresponding "if a branch drops out of a branch, say so in a new PR comment" logic...

Maybe yes, but since that doesn't exist now, it should be able to be handled as a follow-up instead 😬

dscho (approved):

lib/ci-helper.ts
dscho: Since the proposed change would remove the signal (absence of a comment documenting the (re-)integration used to be the indicator that a topic was dropped), I think we absolutely need it, not in a follow-up, but in this PR. So I've added it.

dscho:
@webstech what do you think?

webstech:
> @webstech what do you think?

I will try to finish this tomorrow.  It _feels_ like the 'commit removed' comment is added but the following 'integrated into' comment may still be added.  I will pull and adjust the test for multiple rounds to verify.

email filters could get rid of inbox noise while still letting the PR track status (the commit removal being a good thing to me).

In the end, the only tracked branch status of interest may be the final merge.  Other stale comments could be removed.

webstech:
> I will try to finish this tomorrow. It _feels_ like the 'commit removed' comment is added but the following 'integrated into' comment may still be added.

Yes, the comment is added when the merge happens. If a change is removed, the removal and re-add will be documented.  This is an improvement (to me).

However, doesn't this run on a time interval?  Should it run when the branch is pushed?  Running on a schedule means a removal and remerge could be missed.

dscho:
> However, doesn't this run on a time interval? Should it run when the branch is pushed? Running on a schedule means a removal and remerge could be missed.

Right.

But is it really such a terrible thing if a contributor is _not_ bothered with the information that the Git maintainer played games with the integration branch? I know that I'd hate to see my PR drop out of `seen`, wondering why, only to see a "never mind" PR comment mere minutes later.

webstech:
> But is it really such a terrible thing if a contributor is _not_ bothered with the information that the Git maintainer played games with the integration branch? I know that I'd hate to see my PR drop out of `seen`, wondering why, only to see a "never mind" PR comment mere minutes later.

No, it is not a terrible thing but they are only seeing part of the picture if they did care.  I suspect there won't be many delete and re-add comments.  Most of that activity will occur on the maintainer's fork.

webstech (approved):

HaraldNordgren:
Thanks a lot, @dscho and @webstech!

PR #2295, opened 2026-07-21
ci-helper: decouple mail-note updates from open PRs

dscho:
The `update-mail-to-commit-notes` action needs the upstream branch `seen`, but does not use open pull request heads. Requesting upstream branches nevertheless queried every owner in `config.app.installedOn`.

Since https://github.com/gitgitgadget/gitgitgadget/pull/2218 made `getOpenPRs()` authenticate, the workflow's token for `gitgitgadget/git` is insufficient. Setup aborts with `Need a GitHub token for dscho` before updating notes, see e.g.
https://github.com/gitgitgadget-workflows/gitgitgadget-workflows/actions/runs/29785245976/job/88495272227.

Treat these requirements independently. Mail-note updates can still fetch `upstream/seen` without querying open pull requests, while `update-prs` retains the head prefetch it needs.

### DIFF
--- a/tests/ci-helper.test.ts
+++ b/tests/ci-helper.test.ts
@@ -168,6 +168,46 @@ async function setupRepos(instance: string): Promise<{ worktree: TestRepo; gggLo
     return { worktree, gggLocal, gggRemote };
 }
 
+testQ("setup GitHub Action only fetches open PR heads when requested", async () => {
+    const gitConfigParameters = process.env.GIT_CONFIG_PARAMETERS;
+    const withoutOpenPRHeads = await setupRepos("setup-no-pr-heads");
+    const withOpenPRHeads = await setupRepos("setup-pr-heads");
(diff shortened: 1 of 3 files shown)


tests/ci-helper.test.ts
```diff
@@ -168,6 +168,46 @@ async function setupRepos(instance: string): Promise<{ worktree: TestRepo; gggLo
     return { worktree, gggLocal, gggRemote };
 }
 
+testQ("setup GitHub Action only fetches open PR heads when requested", async () => {
+    const gitConfigParameters = process.env.GIT_CONFIG_PARAMETERS;
```
webstech, line 172: This has me wondering if we really should be modifying `GIT_CONFIG_PARAMETERS` in the env in `setupGitHubAction`.  If (when?) more tests are created for `setupGitHubAction`, could there be a concurrency issue?

webstech (approved):
My concern doesn't affect moving forward with this.

PR #2296, opened 2026-07-24
Updates for typescript and typia versions AND migrate from jest to vitest

webstech:
First:
maint: remove tests-config version of ci-helper.test.ts The test was mor…
…e of a POC.  It is not being maintained and can not run at the same time as the other tests.

Simpler to remove than update to run in parallel with the regular tests with vitest conversion.

Then:
maint: tool version updates and vitest conversion
Updates to move to go based typescript and typia.
Migrate testing from jest to vittest.  This includes building tests as part of regular build.  Rerun of tests and watch mode seemlessly rebuild tests.

Two versions of the compiler are installed so eslint can continue to use the old version.
The git.test.ts includes a filter update on extracting config settings to reduce runtime.

The build may be slower due to cache rebuild.  Need to monitor this.

`build-dist` (ncc) is using the old version of the compiler.  Need to research how to resolve.


webstech:
Wasted a couple of days due to CI build failing.  Tried to figure out why checkout of test repo was getting auth errors.  Eventually reset the PAT value and it worked.  No reason the PAT secret would have been changed.

webstech:
ncc has an issue [#1336](https://github.com/vercel/ncc/issues/1336) regarding ts 7 support.  Bottom line is lack of an api. That is supposed to show up with ts 7.1.  

The dual ts install needed for linting is also needed for ncc.

Edit: building ./build/lib/ci-helper.js with ncc works. Will try to test.
Edit 2: Added a test to import from dist and it successfully ran `setupGitHubAction`.  Not sure why but I needed to set `LOCAL_GIT_DIRECTORY` in the environment.

webstech:
If we want ncc to use the ts 7 generated code:

- `push-github-actions.yml` will need to run `build`
- `build-dist` will need to point to the build output
- a test can be added to `ci-helper.test.ts` for running the dist code - the test could check for dist being present
- vitest would not rebuild dist if a file is changed so that would be incumbent on the dev (the build command could build dist)

@dscho Any thoughts on this?

tests/git.test.ts
```diff
@@ -53,12 +51,12 @@ test("sequencing", async () => {
 
     const lineHandler = async (line: string): Promise<void> => {
         waitTime--;
-        await sleep(waitTime * 50 + (waitTime % 2) * 60); // odd/even have different waits
+        await sleep(Math.abs(waitTime * 50 + (waitTime % 2) * 60)); // odd/even have different waits
```
dscho, line 54: This looks quite strange. It's almost as if `waitTime` could become negative (but isn't expected to)?

tests/git.test.ts
```diff
@@ -53,12 +51,12 @@ test("sequencing", async () => {
...
         buffer += `${line}\n`;
     };
 
-    expect(await git(["config", "--get-regexp", "TEST"], { lineHandler })).toMatch("");
-    expect(await git(["config", "--get-regexp", "TEST"], { trimTrailingNewline: false })).toEqual(buffer);
+    expect(await git(["config", "--get-regexp", "^TEST"], { lineHandler })).toMatch("");
+    expect(await git(["config", "--get-regexp", "^TEST"], { trimTrailingNewline: false })).toEqual(buffer);
```
dscho, line 59: What makes this change necessary?

package.json
```diff
@@ -98,7 +64,7 @@
     "json-stable-stringify": "^1.3.0",
     "jsonwebtoken": "^9.0.3",
     "mailparser": "^3.9.14",
-    "marked": "^18.0.6",
+    "marked": "^18.0.7",
```
dscho, line 67: Sneaky 😉

dscho:
> If we want ncc to use the ts 7 generated code:
> 
> * `push-github-actions.yml` will need to run `build`
> * `build-dist` will need to point to the build output
> * a test can be added to `ci-helper.test.ts` for running the dist code - the test could check for dist being present
> * vitest would not rebuild dist if a file is changed so that would be incumbent on the dev (the build command could build dist)
> 
> Any thoughts on this?

@webstech I have no problems with that, it's not like we're developing GitGitGadget super-actively, most of its functionality is basically finished, so it does not matter if one has to run `npm run build` before `npm run test`.

My main aim here is to make things as convenient as possible. If you think we should go for TS-7-generated code to make everything easier in the future, then I'm all for it.

tests/git.test.ts
webstech: > This looks quite strange. It's almost as if `waitTime` could become negative (but isn't expected to)?

The loop is over the number of `TEST.TS.*` entries added to the config but the regex is looking for anything with `TEST` in it.  The number of lines being returned could exceed the number of entries, causing `waitTime` to go negative.

tests/git.test.ts
webstech: Filtering on entries starting with `TEST` reduces the false hits on entries containing `TEST` that caused the `waitTime` to go negative.

dscho: Ah, so with this change, the `Math.abs()` above becomes unnecessary?

package.json
webstech: I do not remember doing that.  Sorry I missed it during my review.  Do you want it reverted?  There is an open PR for it now.

dscho: No worries, let's keep it and get this PR finished and merged.

dscho (approved):
Thanks for the explanations!

webstech:
> If you think we should go for TS-7-generated code to make everything easier in the future, then I'm all for it.

It gives me more confidence in not breaking things.  No idea of how long it will be until NCC supports TS-7.  I would like to do it in a separate PR if this is ready to go.

tests/git.test.ts
webstech: Well, we can't control what someone might have in their git config so I decided to leave the change in.

dscho:
> > If you think we should go for TS-7-generated code to make everything easier in the future, then I'm all for it.
> 
> It gives me more confidence in not breaking things. No idea of how long it will be until NCC supports TS-7. I would like to do it in a separate PR if this is ready to go.

Sounds good to me! I'm fine with this PR being merged.

dscho:
Oy, I fear that something still went wrong: https://github.com/gitgitgadget-workflows/gitgitgadget-workflows/actions/runs/30480565193/job/90673326452#step:6:11 reports:

```
file:///home/runner/work/_actions/gitgitgadget/gitgitgadget/v1/node_modules/typia/lib/transformers/NoTransformConfigurationError.mjs:4
	throw new Error([
^
Error: Error on typia.createValidate(): no transform has been configured.

Read and follow https://typia.io/docs/setup please.

If you've already completed the setup, it means there's a bug in your code. Run `tsc` command so that check what is wrong with your code.
    at NoTransformConfigurationError (file:///home/runner/work/_actions/gitgitgadget/gitgitgadget/v1/node_modules/typia/lib/transformers/NoTransformConfigurationError.mjs:4:1)
    at Module.createValidate (file:///home/runner/work/_actions/gitgitgadget/gitgitgadget/v1/node_modules/typia/lib/module.mjs:106:1)
    at CIHelper.external_util_ (file:///home/runner/work/_actions/gitgitgadget/gitgitgadget/v1/lib/ci-helper.ts:57:1)
    at file:///home/runner/work/_actions/gitgitgadget/gitgitgadget/v1/lib/ci-helper.ts:25:1
    at ModuleJob.run (node:internal/modules/esm/module_job:439:25)
    at node:internal/modules/esm/loader:643:26
    at run (/home/runner/work/_actions/gitgitgadget/gitgitgadget/v1/handle-pr-push/index.js:2:24)
```

dscho:
Same happened in https://github.com/git/git/pull/2285.

webstech:
> Oy, I fear that something still went wrong: https://github.com/gitgitgadget-workflows/gitgitgadget-workflows/actions/runs/30480565193/job/90673326452#step:6:11 reports:
> 

I am really sorry about this.  I have a test in an upcoming `ci-helper.test.ts` that runs the ncc dist build.  I thought I had run that with the ncc build before running ncc build with the `build` dir `js` files but I may have missed that.

Can the V1 branch be reverted easily to the previous commit for now?

I will work on finishing the changes to run NCC against the `build` dir.

dscho:
> Can the V1 branch be reverted easily to the previous commit for now?

Good call!

dscho:
Rolled back: https://github.com/gitgitgadget/gitgitgadget/compare/7a610052...1eb1036a.

dscho:
Fixed via #2305. Thank you so much @webstech!

PR #2312, opened 2026-08-05
perf: use caching to improve build performance

webstech:
ttsc (needed for typia) builds a cache to improve go performance.  It takes some time. Reusing the cache will make the builds run faster.

The use of the versions in the key limits the number of cache blobs that will be created.  This is especially true with the current dependabot setup.

The key suffix can be changed to `-ttsc-${{ hashFiles('./gitgitgadget/package-lock.json') }}` if using the versions becomes an issue.

The ttsc doc describes this [here](https://ttsc.dev/docs/ttsc/cache/).  The use of `versions` in the key came from this [issue](https://github.com/samchon/ttsc/issues/1064).

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -105,6 +105,28 @@ jobs:
       run: npm ci
       working-directory: gitgitgadget
 
+    - name: Gen version string
+      uses: actions/github-script@v9
+      id: cache-id
+      with:
+        script: |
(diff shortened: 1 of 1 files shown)


webstech:
A second [run](https://github.com/gitgitgadget/gitgitgadget/actions/runs/31054571018/job/92471487490?pr=2312)   of the check shows the cache being used.

dscho (approved):
Very nice!!!

PR #2333, opened 2026-08-19
deps: group build tools and slow down lint updates

webstech:
Add a group for the build tools to keep updates in synch.  Updates will be in a single PR.  Dependabot does not seem to fully support peer dependencies.

Separate lint related updates into a monthly scan to help reduce the PR frequency.

The ability to have two schedules was identified [in this issue](https://github.com/dependabot/dependabot-core/issues/1778#issuecomment-1988140219) using `target-branch`.

If the monthly interval is too long, a `chron` style interval could be used.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,6 +6,29 @@ updates:
     schedule:
       interval: "weekly"
     open-pull-requests-limit: 10
+    groups:
+      build-tools:
+        patterns:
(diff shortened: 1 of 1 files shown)


dscho (approved):
Awesome! This follows the advice from https://github.com/samchon/ttsc/issues/1063#issuecomment-5176182989 and extends the idea.

